PR #94, opened 2026-05-12
fix(api): wrap-sample projection over derived plans

tsouza:
## Summary

Fixes the wrap-sample-projection's three-class column-scope bug
documented in the RC1 follow-up. The wrap assumed canonical
(MetricName, Attributes, TimeUnix, Value) columns are always exposed
by the inner SELECT, which isn't true for:

1. **RangeWindow** — inner SELECT is `<group-keys>, … AS value`.
2. **StructuralJoin** — emit shape is `SELECT R.* FROM (L) JOIN (R) ON …`.
3. **Aggregate / Filter(Aggregate)** — output is just `<group-keys>, agg…`.

Each handler's wrap function now inspects the plan root and picks the
right projection shape; a new optional `Qualifier` field on
`chplan.ColumnRef` carries the join alias.

Un-skips:
- `TestPromQueryRangeRate` (E2E)
- `TestLokiQueryRangeCountOverTime` (E2E)
- `TestTempoSearch_StructuralChild` (E2E)
- `TestTempoSearch_CountScalar` (E2E)
- Playwright `rate(…)` returns a matrix
- Playwright `count_over_time(…)` returns a matrix

This is **P0 1** of the RC2 P0 list — the single highest-leverage RC2
fix (unblocks 4 E2E + 2 Playwright tests in one change).

## Test plan

- [ ] `check` job: unit tests including new `node_negatives_test.go`
  Qualifier cases + `emit_negatives_test.go` `R.SpanName` rendering
- [ ] `lint` job
- [ ] `dashboard` job: 4 previously-skipped E2Es + 2 previously-skipped
  Playwright tests now run and must pass

### DIFF
--- a/internal/api/loki/handler.go
+++ b/internal/api/loki/handler.go
@@ -193,7 +193,14 @@ func wrapWithLogSampleProjection(plan chplan.Node, s schema.Logs, expr syntax.Ex
 			Projections: []chplan.Projection{
 				{Expr: &chplan.LitString{V: ""}, Alias: "MetricName"},
 				{Expr: &chplan.ColumnRef{Name: s.ResourceAttributesColumn}, Alias: "Attributes"},
-				{Expr: &chplan.FuncCall{Name: "now64", Args: []chplan.Expr{&chplan.LitInt{V: 9}}}, Alias: "TimeUnix"},
+				// now64(9) - 5s buffer; see prom handler's synthesizedAnchor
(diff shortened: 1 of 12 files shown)


tsouza:
Superseded by PR 104 which was stacked on this branch and merged first, bringing all the wrap-projection-derived-plans changes along.

PR #115, opened 2026-05-13
feat(api/prom): /api/v1/query_exemplars stub

tsouza:
## Summary

Grafana panels poll `/api/v1/query_exemplars` regardless of whether the
backend stores exemplars; returning 404 / 500 paints a red banner on
the panel.

This stub returns `200 OK` + empty array (`{status: success, data: []}`)
so the panel renders normally without exemplar overlays. Real exemplar
storage lands in RC4 (self-observability) when cerberus emits its own
OTel data into ClickHouse.

## Test plan

- [ ] `check` — `TestQuery_Exemplars_Stub`
- [ ] `lint`


tsouza:
Closing per design discussion: empty-list stubs add maintenance + CI surface for zero functional value. A 404 is honest about the gap; real query_exemplars lands with RC4 self-observability when cerberus emits its own OTel data.

PR #146, opened 2026-05-13
refactor(chsql): actually use SelectBuilder for R6.2/R6.3 ports + repo audit

tsouza:
## Summary

- Rewrites the R6.2 + R6.3 ported `emit_node.go` functions (`emitScan`,
  `emitFilter`, `emitProject`, `emitAggregate`, `emitLimit`) to use
  `chsql.SelectBuilder` (`Select` / `SelectAs` / `From` / `Where` /
  `GroupBy` / `Limit`) instead of hand-writing `prefix.WriteSQL("SELECT ")`
  / `WriteSQL(" FROM ")` / `WriteSQL(" WHERE ")` / `WriteSQL(" GROUP BY ")`
  / `WriteSQL(" LIMIT ")`.
- Adds `(*emitter).subqueryFrag(node) chsql.Frag` so a child plan can
  nest into `SelectBuilder.From` without flattening through `splice`.
  Pre-renders the subquery synchronously so chplan errors still surface
  to the caller.
- Adds `chsql.As(expr Frag, alias string) Frag` + the
  `SelectBuilder.SelectAs(expr Frag, alias string)` slot so projection
  aliases stop composing the `AS` keyword by hand.
- Lands `docs/chsql-audit.md` — every SQL-building callsite in
  `internal/` and `internal/api/` categorized as
  **cosplay** / **grandfathered** / **legitimate token write**. Cosplay
  fixed in this PR; grandfathered annotated with the R6.x milestone
  that ports it.

## Why

R6.1 shipped a real typed `SelectBuilder` with all the clause slots.
R6.2 (#138) and R6.3 (#140) did not use it — they composed `Builder`
+ `WriteSQL("SELECT ")` + manual `", "` joins. That is `fmt.Sprintf`
with extra steps; it defeats the purpose of the typed builder. This
PR fixes the regression and prevents it recurring by documenting the
audit + the rule in `docs/chsql-audit.md` (which the upcoming R6.9
lint check will key off).

## Audit results (from `docs/chsql-audit.md`)

- **Cosplay fixed**: 11 callsites in `internal/chsql/emit_node.go`.
- **Grandfathered (R6.4–R6.7)**: 5 files (`emit_expr.go`,
  `range_window.go`, `vector_join.go`, `structural_join.go`,
  `internal/api/prom/metadata.go`) + the pre-R6 `emitOrderBy` block.
  Untouched in this PR; the milestone PRs do the mechanical port.
- **Legitimate token writes**: ~30+ callsites across
  `internal/api/loki/` and `internal/chsql/builder_test.go`; all are
  operator-glue inside `Frag` callbacks or test bodies. None replicate
  a `SelectBuilder` slot.

## Test plan

- [x] `go build ./...` clean.
- [x] `go test ./...` green (844 tests).
- [x] `just update-golden` diff empty — SQL output byte-identical for
  every TXTAR fixture.
- [x] `go vet ./...` clean.
- [x] `$HOME/go/bin/gofumpt -l internal/` empty.
- [x] `markdownlint-cli2 docs/chsql-audit.md` clean.
- [x] Final grep audit empty: `grep -RnE 'WriteSQL\(" *(SELECT|FROM|WHERE|GROUP BY|ORDER BY|LIMIT|PREWHERE)' internal/chsql/emit_node.go internal/chsql/builder_test.go internal/api/loki/ internal/api/prom/ internal/api/tempo/`

### DIFF
--- a/docs/chsql-audit.md
+++ b/docs/chsql-audit.md
@@ -1,0 +1,157 @@
+# chsql callsite audit
+
+Repo-wide inventory of every SQL-building callsite in cerberus. Categorises
+each one as **cosplay** (uses `Builder.WriteSQL` for a clause keyword
+where `SelectBuilder` already has a typed slot), **grandfathered**
+(pre-R6.1 emitter code retained until its RC6 port milestone), or
+**legitimate token write** (operator-token `WriteSQL` inside a `Frag`
+that no typed helper covers).
+
+This audit was triggered by the maintainer catching the R6.2 (PR #138)
+and R6.3 (PR #140) ports shipping `prefix.WriteSQL("SELECT ")` /
+`WriteSQL(" FROM ")` / `WriteSQL(" WHERE ")` — that pattern is
+`fmt.Sprintf`-on-SQL with an extra step and defeats the whole point of
+the typed `chsql.SelectBuilder`. This PR fixes every bucket-1 entry.
+
+Detection greps (see CLAUDE.md hard rule "no raw SQL strings"):
+
+```sh
+grep -RnE 'WriteSQL\(" *(SELECT|FROM|WHERE|GROUP BY|ORDER BY|LIMIT|PREWHERE|HAVING|UNION|JOIN |LEFT JOIN|INNER JOIN)' internal/ cmd/ harness/
+grep -RnE 'WriteSQL\(" AND "\)' internal/ cmd/ harness/
+grep -RnE 'fmt\.Sprintf\([^)]*(SELECT|FROM|WHERE|INSERT|GROUP BY|ORDER BY)' internal/ cmd/ harness/
+grep -RnE 'WriteString\([^)]*(SELECT|FROM|WHERE|GROUP BY|ORDER BY|LIMIT|PREWHERE)' internal/ cmd/ harness/
+```
+
+## Bucket 1 — Cosplay (FIXED in this PR)
+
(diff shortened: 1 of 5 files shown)


tsouza:
**Self-review (LGTM with three non-blocking nits, all queued for R6.4)**

Genuinely uses the typed surface — `NewSelect().Select / SelectAs / From / Where / GroupBy / Limit` everywhere it should. The new `As` Frag + `SelectAs` slot are clean (empty-alias short-circuit is the right move). `subqueryFrag` pre-rendering child plans so chplan errors surface synchronously is a nice design call. Grandfathered emitters (`emitOrderBy`, `vector_join.go`, `structural_join.go`) are properly annotated with the milestone that retires them.

Nits to follow up on in R6.4:

1. The pre-flight pattern (`(&Builder{}).Expr(pr.Expr)` then a second Expr inside the Frag callback) double-renders and risks subtle divergence if `Expr` ever carries state. The `subqueryFrag` "render-once, replay (sql, args)" pattern would let `emitFilter` / `emitProject` / `emitAggregate` drop the pre-flight loops entirely.
2. `int(l.Count)` int64→int downcast is fine on 64-bit but worth extending `SelectBuilder.Limit(int64)` later.
3. `emitAggregate` pre-flights every expression three times (pre-flight + SELECT + GROUP BY). Same root cause as #1.

PR #223, opened 2026-05-14
feat(test): TXTAR seed + expected_rows via chDB (RC8 R8.1)

tsouza:
## Summary

Extends the TXTAR runner with two optional sections — `seed:` and
`expected_rows:` — that opt a fixture into a chDB-backed semantic
assertion on top of the existing text-equality goldens. When both
sections are present, a new `chdb` build-tagged executor opens an
ephemeral in-process chDB session, applies the seed DDL+INSERTs,
runs the emitted `sql:` + `args:` against it, and asserts the
result set matches `expected_rows:`. This turns today's text-only
goldens into semantic assertions — a chsql refactor that silently
inverts predicate order is now caught by row mismatch.

Without the `chdb` build tag the sections are inert: `just test`
stays CGO_ENABLED=0 and never pulls in `chdb-go`.

Map columns are rewritten server-side to `toJSONString(...)` (per
the R8.0 probe) so the parquet-go MAP panic in chdb-go v1.11.0 is
dodged; row comparison normalizes numbers to float64 and decodes
Map cells into `map[string]any`, leaning on `reflect.DeepEqual` for
JSON key-ordering robustness.

**Pilot fixtures (5)**: `abs_metric`, `bool_lt`, `ceil_metric`,
`clamp_max`, `floor_metric` — five simple gauge transforms on
`otel_metrics_gauge` with deterministic single-row seeds.

**Runner extension**: ~500 LoC across:
- `test/spec/roundtrip.go` — section parsing (always compiled).
- `test/spec/runner_chdb.go` — chDB executor (`//go:build chdb`).
- `test/spec/runner_nochdb.go` — no-op fallback (`//go:build !chdb`).
- `test/spec/promql/roundtrip_chdb_test.go` — driver (chdb-tagged).

**CI**: `chdb` workflow gains a `Run TXTAR round-trip suite (chDB)`
step calling `just spec-chdb`. Still nightly + manual dispatch only,
NOT a required PR check.

**Determinism**: documented in `docs/test-strategy.md`. Single-row
seeds satisfy it trivially; multi-row fixtures must include `ORDER
BY` in either the seed table or the emitted SQL.

## Stacks on

#221 (R8.0 chDB probe). Auto-merge enables only after #221 lands.

## Test plan

- [ ] `just test` (default lane) still passes; new sections inert.
- [ ] `just chdb-install && just spec-chdb` exercises all 5 pilot fixtures.
- [ ] `chdb` workflow runs both steps on nightly/dispatch.
- [ ] Adding `seed:` + `expected_rows:` to a 6th fixture works without runner changes.

### DIFF
--- a/.github/workflows/chdb.yml
+++ b/.github/workflows/chdb.yml
@@ -48,3 +48,6 @@ jobs:
 
       - name: Run chDB probe
         run: go test -tags chdb -v -count=1 -run TestChDBProbe ./internal/chclient/...
+
+      - name: Run TXTAR round-trip suite (chDB)
+        run: just spec-chdb
--- a/Justfile
+++ b/Justfile
@@ -59,6 +59,15 @@ test:
 schema-ddl-test:
     go test -race -tags=integration ./internal/schema/ddl/...
 
+# Run the TXTAR spec suite with the chDB-backed round-trip assertion
+# layer enabled. Requires libchdb.so (see `just chdb-install`). The
+# default `just test` lane stays CGO_ENABLED=0 and never compiles the
+# chdb-go driver. Only fixtures that declare both `seed:` and
+# `expected_rows:` are executed against chDB; everything else still
+# runs through the text-equality golden path.
+spec-chdb:
+    go test -tags chdb -count=1 ./test/spec/...
+
 # Run the FuzzParse target for one parser head for a bounded duration.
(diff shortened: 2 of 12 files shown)


tsouza:
Stacks on #221 — enable auto-merge after #221 lands.

PR #452, opened 2026-05-18
fix(harness/tempo-compat): make live-store search smoke non-fatal

tsouza:
## Summary

Single commit `d32c62de` from the user's earlier in-flight harness work.

> fix(harness/tempo-compat): make live-store search smoke non-fatal

## Why opened now

Branch was pushed but no PR ever opened. Rather than delete the work, opening here so it either lands or gets an explicit close decision.


tsouza:
Both commits on this branch (live_store field names + non-fatal smoke search) are already on main via earlier merged PRs (#424, #427). After rebasing on main, the branch has zero unique commits — closing as already-applied. Remote branch will be deleted.

PR #715, opened 2026-05-22
test(compat/loki): pin k8s-family stream labels (pod/namespace/service_name/cluster/container) in seeder

tsouza:
## Summary

PR #712 retires the historic `cerberus-test-queries.yml` `should_skip` overlay, surfacing four "Plain seed-gap" rows whose rationale comments pinned them to the seeder missing the five k8s-family stream labels:

- `exhaustive/aggregations.yaml#Count aggregated by pod`
- `exhaustive/aggregations.yaml#Count aggregated by namespace`
- `exhaustive/aggregations.yaml#Count aggregated by service_name`
- `exhaustive/aggregations.yaml#Count aggregated by cluster and namespace`
- `exhaustive/aggregations.yaml#Count aggregated by service_name and container`
- `exhaustive/unwrap-aggregations.yaml#Without multiple labels` (`sum without (namespace, cluster)`)

Investigation: the seeder's shape has been correct since PR #525 hoisted the stream-identity labels into ResourceAttributes. All five required keys are present in both the pushLoki stream label set (so reference Loki indexes the streams) and the insertCHLogs ResourceAttributes map (so cerberus's LogQL group-by lowering resolves `ResourceAttributes[<key>]` via `logql.levelAwareGroupKey`). The `should_skip` rationale comments were stale — they survived past PR #525 without re-validation.

This PR pins the shape with a unit-level regression test (`TestSeederWritesK8sStreamLabels`) so any future trim of the seed map surfaces at PR-review time rather than on the compatibility lane (which is where every previous regression of this shape was caught). The seeder itself gains clarifying comments at both call sites (`buildStreams` for the Loki push, `insertCHLogs` for the CH-side ResourceAttributes) tying each call-site to the corpus rows it underpins.

### What the test asserts

For every stream produced by `buildStreams`:

- The pushLoki label map has all five required keys (`pod`, `namespace`, `service_name`, `cluster`, `container`) with non-empty values.
- The matching `insertCHLogs` ResourceAttributes map (re-built from the same `serviceConfig`) has all five required keys with non-empty values.
- The per-key values are identical across the two maps — silent value drift would surface in the differ as a value-mismatch shape instead of the more legible empty-series shape.

### Labels touched + how

| Label | pushLoki stream label | insertCHLogs ResourceAttributes |
| --- | --- | --- |
| `pod` | already present | already present |
| `namespace` | already present | already present |
| `service_name` | already present | already present |
| `cluster` | already present | already present |
| `container` | already present | already present |

No values change. The seeder shape is identical pre/post this PR; only the comments and the test are new.

## Test plan

- [ ] CI `check` runs `TestSeederWritesK8sStreamLabels` and it passes.
- [ ] CI `compatibility/loki` runs the four corpus rows above without the retired `should_skip` overlay and the differ reports parity (depends on PR #712 landing first; this PR is the seeder-side guarantee that closes the diagnosis loop).
- [ ] CI `forbid-skip`, `lint`, `compose-smoke` stay green — no source changes outside the compat-loki seeder package.

### DIFF
--- a/compatibility/loki/cmd/seed/main.go
+++ b/compatibility/loki/cmd/seed/main.go
@@ -221,6 +221,16 @@ func buildStreams(start time.Time) []stream {
 	out := make([]stream, 0, len(serviceConfigs))
 
 	for _, sc := range serviceConfigs {
+		// The five k8s-family labels (pod, namespace, service_name,
+		// cluster, container) are load-bearing for the
+		// exhaustive/aggregations.yaml#Count aggregated by {...} and the
+		// exhaustive/unwrap-aggregations.yaml#Without multiple labels
+		// cases in the vendored bench corpus. They mirror the upstream
+		// generator (loki-bench/generator.go) shape so the reference
+		// Loki indexes the same 13 streams the cerberus side reads via
+		// ResourceAttributes. See insertCHLogs for the matching CH-side
+		// map; TestSeederWritesK8sStreamLabels pins the presence of all
+		// five keys in both places.
 		labels := map[string]string{
 			"cluster":      sc.Cluster,
 			"namespace":    sc.Namespace,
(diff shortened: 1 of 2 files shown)


tsouza:
Superseded by #721 which replaces the seed-settle heuristics with authoritative `/metrics` polling. The label-presence regression test #715 added is no longer load-bearing since #721's gate uses metric deltas rather than cardinality. Closing.

PR #851, opened 2026-06-13
perf(chsql): collapse rate-range fan-out via ASOF boundary lookup (kill O(rows×anchors))

tsouza:
## Problem (measured, from the #849 benchmark fix)

`sum(rate(metric[5m]))` query_range @15s/1h fanned each of ~5M samples across its ~range/step+1 ≈ 21 covering anchors → ~105M (sample,anchor) intermediate rows → ~6.4s single-statement wall. The cost is the fan-out in `emitWindowedArrayExtrapolatedMatrix` → `sampleAnchorFanoutFrag`. `rate`/`increase`/`delta` can't use the RangeLWR collapse (#75, last-value only) — they need first+last+reset-adjusted-increase per window.

## Design — reset-adjusted cumulative + ASOF boundary join (extrapolation formula unchanged)

Prom's `extrapolatedRate` reads only six per-window quantities, all derivable from the **two boundary samples + per-series precomputed columns**, with no per-anchor interior-sample materialisation:

- The window `(anchor-range, anchor]` over a ts-sorted per-series stream is a **contiguous slice**, so `counter_delta = cumV[end] - cumV[start]` where `cumV` is the reset-adjusted cumulative (`arrayCumSum` over the same `if(c<p,c,c-p)` repair the fan-out applied pairwise) — the pre-window pair cancels in the subtraction, leaving exactly the fan-out's `arraySum`.
- `n` (sample count, feeding `averageDurationBetweenSamples`) `= end_rank - start_rank + 1`. A per-`ts` **multiplicity** (`countEqual`) recovers the count when CH's ASOF collapses both boundaries onto a duplicate-timestamp row, so the all-same-ts `sampled_interval==0 -> nan` case still surfaces (byte-exact with the fan-out).
- `first_ts`/`last_ts`/`first_val` read straight off the boundaries.

Each series is enriched **once** `(ts, v, rank, cumV, mult)` as a `WITH enr` CTE (one row per sample, no anchor multiplication), and each anchor finds its window-end (`ts <= anchor`) and window-start (`ts > anchor-range`) boundaries via **two ASOF LEFT JOINs** — the canonical rate-via-ASOF pattern, O((samples+anchors)·log). The CTE is referenced 3× (grid series-discovery + E + S) but materialised once. The downstream `extrapolatedValueFrag` / `duration*Frag` chain is **unchanged** — only the SOURCE of the boundary quantities changed.

### New chsql Builder constructor
Added a typed `ASOFLeftJoin` `JoinKind` to `internal/chsql/builder.go` (renders `ASOF LEFT JOIN`, flows through the existing `QueryBuilder.Join` slot — no raw SQL).

### Gating
Keyless rate (no group key → no ASOF equi-key) stays on the fan-out emitter; every PromQL query_range rate/increase/delta carries the `Attributes` series key, so the common path takes ASOF.

## Parity evidence (the whole risk — exact, not approximate)

- **`-- expected_rows --` byte-identical** across every `test/spec` fixture (only the `-- sql --` shape changed — confirmed zero `expected_rows` diffs).
- **Property/oracle lane** green (instant coverage); since that lane is instant-only, a **new chDB pin** (`range_window_asof_parity_chdb_test.go`) proves **bit-exact ASOF-vs-fanout parity** for rate/increase/delta over an adversarial counter (resets, irregular/sparse spacing, counter zero-crossing clamp, the 1.1× extrapolation-threshold clamp, duplicate timestamps) across 231 anchors. *This pin caught a real delta bug* (the gauge `last_val - first_val` read a placeholder array) — fixed by making `window_vals` an `arrayWithConstant(n, last_val)` whose length is `n` and whose tail is `last_val`.
- **A-vs-B solver chDB lane** green, including its NaN-recovery coverage gate (the dup-input-timestamp `sampled_interval==0 -> nan` path).
- Full suite: 4964 non-chdb + 1659 chdb tests pass; golangci-lint 0 issues; gofumpt/go-arch-lint clean.

## Fan-out collapse (the proof — cardinality ratchet)

`fan_factor` drops to **1.0** (no fan-out) on every rate-range fixture; baseline updated for these legitimate improvements only:

| fixture | fan_factor | peak_intermediate |
|---|---|---|
| `sum_by_rate_range_step` | 5.83 → **1.0** | 70 → 18 |
| `sum_by_rate_dotted_source` | 5.83 → **1.0** | 105 → 27 |
| `sum_by_rate_range_offset` | 3.22 → **1.0** | 58 → 12 |
| `sum_by_rate_range_offset_negative` | 3.33 → **1.0** | 40 → 10 |
| `subquery_over_rate_range` | 6.95 → **3.07** | 132 (scan rose) |
| `subquery_over_increase` / `_sum_rate` | 3.0 → **1.0** | 6 → 1 |

## Funcs landed
`rate`, `increase`, `delta` (the extrapolated family). `irate`/`idelta` (last-two-samples) and `*_over_time` are out of scope (different paths). The fan-out emitter remains as the fallback for keyless rate.


tsouza:
Closing as superseded by the orchestrator-first direction. Regenerating docs/benchmarks.md against this ASOF emitter showed the rate-range wall REGRESSED 6.31s->36.90s at 5M scale (cardinality/memory down, but ASOF JOIN + window-function enrichment far slower than the arrayJoin fan-out + GROUP BY on a single engine). The cardinality ratchet (tiny fixtures) missed it; timings aren't a CI gate. The 'delta bug' this caught was in this PR's own new range_window_asof.go (the synthetic window_vals reconstruction), not a pre-existing main bug, so there's nothing to salvage. The right rate-range fix is the sharded solver running the fast fan-out per shard in parallel (memory relief + wall speedup) — being measured through the solver against real ClickHouse now. ASOF stays for single-boundary lookups only (RangeLWR). ASOFLeftJoin constructor can be revived if a genuine single-boundary use appears.

PR #873, opened 2026-06-14
feat(promql): implement experimental info() label-enrichment join

tsouza:
## Function

Experimental PromQL `info(v instant-vector, [label-selector])`. Cerberus parsed it (`EnableExperimentalFunctions`) then 422'd at the lowering fall-through; this PR lands real lowering → ClickHouse SQL at reference parity.

## Reference semantics

Per `prometheus/promql/info.go::evalInfo`:

- Each input series is enriched with the **data labels** of the matching `target_info`-style **info series**, joined on the hard-coded **identifying labels** `instance` + `job`.
- **Sample values / timestamps / metric name are carried through UNCHANGED** — `info()` only adds labels.
- **Base labels win on conflict** (info data labels are skipped where the base already has that label).
- An input series with **no matching info series is returned unenriched** (for the bare / empty-filter shape).
- The optional `{…}` selector's `__name__` matcher selects the info metric (default `target_info`); its non-`__name__` matchers both **constrain which info series participate** and **restrict which data labels are copied** (by name).

## Lowering + emit approach

- New `chplan.InfoJoin` node (`Base`, `Info`, `IdentifyingLabels`, `DataLabelFilter`).
- `internal/promql/info_fn.go` lowers `info()`: arg0 → base (normal LWR pipeline), builds a `target_info` `VectorSelector` for the info side (LWR + PREWHERE apply as for any selector), wraps both in `InfoJoin`.
- `internal/chsql/info_join.go` emits a **LEFT JOIN** of the base side against the per-signature info aggregation. **Typed Frags only** (no raw SQL):
  - join key = `mapFilter((k,v)->k IN (instance,job), Attributes)` on both sides;
  - output `Attributes` = `mapConcat(R._info_data, B.Attributes)` — base map LAST so base wins; unmatched base rows get an empty info map (CH join-column default) and pass through unenriched;
  - info data-label map = `mapFilter((k,v)->NOT k IN (identifying,__name__) [AND k IN dataFilter], Attributes)`, newest-wins via `argMax(..., TimeUnix)`.
- Wired into the emit dispatch, `clone`, optimizer `rewriteChildren`, the chplan pretty-printer, and the `clone` / slice-invariant exhaustiveness guards (28 → 29).

## Parity evidence (chDB-backed fixtures)

All green under `go test -tags chdb ./test/spec/promql` (`-- seed --` / `-- expected_rows --` cells pinned against reference behaviour):

| fixture | pins |
|---|---|
| `info_target_info_enrich` | two `up` series enriched with `target_info.version` |
| `info_data_label_selector` | `{version=~".+"}` copies only `version` (`region` excluded) **and** filters the info-side scan |
| `info_base_wins_and_unmatched` | base `env=prod` beats info `env=staging`; an orphan series with no matching `target_info` is kept unenriched |

Plus unit tests in `internal/promql/info_fn_test.go` (default metric, data-label selector, explicit `__name__` override, regex-`__name__` rejection).

## Unmasked panel

`test/e2e/grafana/compose/dashboards/showcase-promql.json`: the `info` panel flips from `cerberus.expect: error:not yet supported` to `nonempty`. `test/e2e/seed/cmd/seed/main.go` seeds a `target_info` gauge (`job=api`/`version=1.2.3`, `job=db`/`version=4.5.6`) matching the existing `up` series so the panel renders enriched series.

## Residual risk

Documented and **gated, not silently mis-answered**:

- The reference `effectiveInfoNameMatchers` synthesis (regex / negated / multi-metric info-name selection) is outside this envelope — a non-equality `__name__` selector is **rejected with a clear error**.
- The per-series "drop when a data-label matcher does not match the empty string and no info matched" branch is not modelled. The supported selector shapes (bare, `{__name__=…}`, `{label=~".+"}`) all hit the LEFT-JOIN pass-through path, which is parity-correct for them.

### DIFF
--- a/internal/chplan/clone.go
+++ b/internal/chplan/clone.go
@@ -121,6 +121,13 @@ func cloneCompositeNode(n Node) Node {
 		c.Match.Labels = cloneStrings(v.Match.Labels)
 		c.Include = cloneStrings(v.Include)
 		return &c
+	case *InfoJoin:
+		c := *v
+		c.Base = CloneNode(v.Base)
+		c.Info = CloneNode(v.Info)
+		c.IdentifyingLabels = cloneStrings(v.IdentifyingLabels)
+		c.DataLabelFilter = cloneStrings(v.DataLabelFilter)
+		return &c
 	case *VectorSetOp:
 		c := *v
 		c.Left = CloneNode(v.Left)
--- a/internal/chplan/clone_test.go
+++ b/internal/chplan/clone_test.go
@@ -55,6 +55,11 @@ func allNodeKinds() []chplan.Node {
 			Left: leaf, Right: &chplan.Scan{Table: "r"}, Match: chplan.VectorMatch{Labels: []string{"job"}, On: true},
 			Include: []string{"inst"}, ValueColumn: "Value",
 		},
+		&chplan.InfoJoin{
+			Base: leaf, Info: &chplan.Scan{Table: "r"},
+			IdentifyingLabels: []string{"instance", "job"}, DataLabelFilter: []string{"version"},
+			ValueColumn: "Value",
+		},
 		&chplan.VectorSetOp{Left: leaf, Right: &chplan.Scan{Table: "r"}, Match: chplan.VectorMatch{Labels: []string{"job"}}, ValueColumn: "Value"},
 		&chplan.NaryVectorSetOp{
 			Arms: []chplan.Node{leaf, &chplan.Scan{Table: "r"}, &chplan.Scan{Table: "s"}},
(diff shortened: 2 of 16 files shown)


tsouza:
Superseded by #875 (feat/promql-info-join), which carries the same info-join implementation plus the surface-parity ledger, perf baselines, and the EnableExperimentalFunctions ratchet-test fix needed for green CI. Closing as duplicate; #875 is the canonical info() PR. (Follow-up: port this PR's extra fixtures info_base_wins_and_unmatched + info_data_label_selector into #875 for richer parity coverage.)

PR #917, opened 2026-06-14
ci(e2e): shard the crawl BFS frontier across dashboard k3d shards

tsouza:
Follow-up to #916 (which landed the coarse 3-shard matrix). This shards the **crawl BFS frontier itself** so the ~50min single-`test()` full-depth crawl parallelizes across multiple k3d clusters — the only path below the 50min floor.

## Where the cost is
Reading the engine confirms the asymmetry the task assumed: **discovery** (navigate + `harvestLinks` + `canonicalTarget`) is one cheap nav per surface; the **dominant cost** is the per-surface interaction sweep (`sweepInteractions` drives a *fresh navigation per planned gesture* — many per eligible surface at full depth, governed by `CONTEXT_RECYCLE_NAVIGATIONS`) plus the per-page oracle battery.

## How it's partitioned
Every shard runs the **whole** cheap BFS discovery walk (so the visited set converges identically), but shard `i` audits + pins only the surfaces it **owns**: `fnv1a(path) % CRAWL_SHARD_COUNT == CRAWL_SHARD_INDEX`. Ownership hashes the **bare path** (strip `?query` and `#state`), so a surface, its structural-param children (`?var-groupBy=kind`, click-discovered by the sweep), and its in-place states all co-locate on one shard — **discovery and ownership coincide**, so no shard depends on a surface only another shard could discover. Default (unset env) = single shard = today's behavior; compose-smoke is unchanged.

## How coverage stays exact
- Per-shard ratchet: `diffInventory` gains an `owns` predicate so each shard asserts only its owned rows.
- Union-merge: on the `update_crawl_inventory` dispatch each shard emits its owned slice; the new `dashboard-merge` job (`crawl-inventory-merge.mjs`) unions them, asserting an **exact, disjoint, total** cover — missing shard / double-owned surface / smuggled surface all fail loudly. A per-shard ratchet alone could miss a surface no shard owned; the union-merge proves the slices reconstitute the whole.

## Unit tests (pure, isolated)
`crawl-inventory-merge.test.mjs` + the sharding pins in `crawl.spec.ts` prove: assignment is total + disjoint + deterministic over a sample set; a family co-locates on one shard; the union of slices == the full set; the merge fails on missing/overlapping/smuggled slices; FNV-1a lockstep between the `.mjs` and `.ts`. All 15 mjs tests + `tsc --noEmit` pass.

## Layout
2 smoke shards + `CRAWL_SUB_SHARDS=2` crawl shards = 4 k3d clusters (kept modest — k3d is heavy/flaky). Expected wall-clock 50min → ~50/2 ≈ 25min for the crawl long pole.

## Needs a dispatch to validate
The k3d inventory is still unbootstrapped (`surfaces: []`). A `workflow_dispatch` with `update_crawl_inventory=true` is required to bootstrap: each sub-shard runs the full crawl, emits its slice, and `dashboard-merge` unions them into the full inventory artifact to commit. The logic is correct-by-construction (single-shard == full crawl) but cannot be validated against a live cluster locally.

### DIFF
--- a/.github/scripts/README.md
+++ b/.github/scripts/README.md
@@ -112,25 +112,51 @@ wrapper, plus `appendStepSummary` / `setOutput` for the runner files.
 - **`dashboard-matrix.mjs`** — `e2e.yml`, the `dashboard-setup` job. The k3d
   twin of `compose-smoke-matrix.mjs`: single source of truth for how the
   `dashboard` (k3d) lane fans its Playwright spec set across a MODEST matrix
-  (3) of isolated-k3d-cluster shards. The dominant cost is the crawl BFS — one
-  indivisible async `test()`, the ~50min long pole — so the parallelism is
-  COARSE: two smoke shards (non-crawl specs, `CRAWL_STACK` unset → `crawl/**`
-  ignored) run CONCURRENTLY with a DEDICATED crawl shard (`CRAWL_STACK=k3d`,
-  `SWEEP_DEPTH=full`). Splitting the crawl frontier itself is the follow-up.
-  The `SHARDS` partition (each carrying `specs` + `crawlStack` + `runGoE2E`) +
-  `EXCLUDED` list live in this module; specs are DISCOVERED (`git ls-files`) so
-  a new `*.spec.ts` is a hard failure unless assigned or excluded — no silent
-  no-run. Coverage assertion is collect-all-violations (unassigned,
-  double-assigned, phantom/stale, bad-shard-name, and the "exactly one shard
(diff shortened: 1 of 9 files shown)


tsouza:
Closing per maintainer decision: keep the dashboard parallelization simple. The basic spec-matrix split (#916) already gives parallel dashboard Playwright on every push and is what was asked for. This crawl-frontier sharding (hash-partitioned BFS + union-merge ratchet) is over-scoped for the ask and only validatable via manual dispatch — deferring it. The work is preserved on this branch/PR record if we ever want the 50min→25min crawl split later.

PR #924, opened 2026-06-14
fix(chclient): evict+retry stale pooled conns so the breaker recovers promptly after CH restart

tsouza:
## The flap (chaos ch-pod-kill, run 27509796946)

The chaos `ch-pod-kill` scenario deletes the ClickHouse pod (Recreate, PVC-backed → comes back with data). On the affected run the breaker recovered at 19:42:55 (`ch circuit breaker recovered: probe succeeded, circuit closed`) and then **immediately re-entered half-open and flapped open↔half-open↔closed for the full ~300s recovery deadline**, ending OPEN. `/readyz` and all three heads returned 200 throughout (the `HeadProbe` breaker recovered fine), but the main `HeadProm` breaker kept re-tripping — and a prior identical run recovered cleanly, so the behaviour was non-deterministic.

## Root cause (code-confirmed)

The probe breaker (`HeadProbe`) and the main breaker (`HeadProm`) are separate breakers over the **one shared** `clickhouse-go` connection pool. After the pod is killed+recreated:

- The OLD pod's TCP conns sit in `ESTABLISHED` with **no FIN/RST** (force-killed pod, IP gone).
- `clickhouse-go`'s `acquire()` validates a pooled conn with a **non-blocking socket read** (`conn_check.go`): an idle-but-not-yet-noticed-dead socket returns `EAGAIN` → `isBad()` reports the **stale conn healthy** and hands it out.
- The query against it then fails with a broken-conn error. `clickhouse-go` evicts that one conn (`release(conn, err) → close`), **but the failure has already surfaced** to `chclient`, which did `c.br.record(ctx, err)` — counting it as a real CH-health failure.
- With several stale conns pooled (`MaxIdleConns=5`/`MaxOpenConns=10`), each first use trips one breaker failure. In bursts of ≥`threshold` within the window the main breaker re-opens. The probe happened to dial a fresh conn and recovered, while main queries kept grabbing stale conns — hence the asymmetric multi-minute flap.

The native `clickhouse.Conn` interface cerberus uses (unlike the `database/sql` wrapper) does **no `ErrBadConn`-style transparent retry** — one acquire, one query, one failure, straight to the breaker.

## The fix

Replicate Go's `database/sql` `ErrBadConn` precedent (the canonical handling of exactly this "server earlier closed the connection" case): on a **broken-conn error**, retry the read-path open on a freshly-acquired conn — **bounded and immediate, no backoff** (the server is up; a fresh dial succeeds at once, so backoff would only delay recovery).

- New `internal/chclient/retry.go`:
  - `isBrokenConnError` mirrors `clickhouse-go`'s own (unexported) `isConnBrokenError` set — `io.EOF` / `io.ErrUnexpectedEOF` / `EPIPE` / `ECONNRESET` / `*net.OpError` — plus `ECONNREFUSED` (a fresh dial to a pod still in warm-up). It is a **narrow allow-list**: a typed `*clickhouse.Exception` (CH answered → alive), a context cancel/deadline, an acquire-timeout, and `ErrCircuitOpen` are **not** retried.
  - `withTransportRetry` runs the call and re-drives it up to `maxTransportRetries (=3)` times while the error stays a broken-conn error (≈ `database/sql`'s 2-pooled + 1-fresh shape; the bound also drains a small stale-conn pool in one request). Each retry's `acquire()` skips the just-evicted bad conn, so it dials fresh.
  - `queryOpen` / `pingOpen` route the SELECT/Ping open through it. **Exec (DDL/DML incl. INSERT) deliberately does not retry** — a broken pipe mid-send could have partially applied, and the `ErrBadConn` contract forbids retrying when the server might have performed the operation.
- A recovered stale conn now records breaker **SUCCESS** (the retry returns `nil`), so a lone stale conn can never advance the breaker. A genuinely-down CH fails every attempt, exhausts the bound, returns the broken-conn error to `record()`, and **still trips the breaker** — the outage path is unmasked, recovery is deterministic.
- Defense-in-depth: pool `ConnMaxLifetime` default `1h → 5m` (official ClickHouse guidance is to lower it for restart-heavy workloads). `clickhouse-go` v2.46.0 has no idle-time knob, and a force-killed socket can pass the per-acquire socket check, so a never-queried idle stale conn is otherwise only retired once it ages past `ConnMaxLifetime`. The retry handles queried conns instantly; this caps the loiter window for idle ones.

## Why not exponential backoff / a retry lib

Researched the precedent: for the **stale-conn-to-a-restarted-healthy-backend** case the server is up, so immediate retry (the `database/sql` no-backoff `ErrBadConn` path) is correct — backoff would only slow recovery. Backoff belongs to the **genuinely-down** mode, which here is already owned by the circuit breaker. A hand-rolled bounded immediate retry mirroring `db.retry()` is more idiomatic than pulling in a backoff lib and configuring its delay to zero.

## Tests

`internal/chclient/retry_test.go`:
- single stale conn recovers silently, breaker stays closed (and exactly 2 attempts: 1 stale + 1 fresh);
- a burst of stale-conn queries past the trip threshold never opens the breaker;
- a dead backend (every attempt fails) **still** opens the breaker after `threshold` failures, at `threshold × (maxTransportRetries+1)` attempts;
- several stale conns drained within one request;
- Ping recovery (readiness path);
- `isBrokenConnError` classification matrix (broken-conn shapes retryable; exception/ctx/acquire-timeout/circuit-open not);
- cancelled ctx halts the retry loop.

`go build ./...`, `go test -race ./internal/chclient/... ./internal/config/...`, and `go test ./internal/api/...` all green.

The real integration gate is the chaos `ch-pod-kill` dispatch (the driver runs it). Auto-merge intentionally NOT enabled.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -90,18 +90,20 @@ these connection inputs.
 
 ## Connection pool
 
-These reproduce clickhouse-go/v2's previously-implicit pool defaults verbatim,
-made explicit so the sharded-pushdown solver can raise the ceiling for fan-out
-rather than inherit a hidden driver default. When the pool is exhausted an
-acquire blocks up to `CERBERUS_CH_DIAL_TIMEOUT` and then fails with a
-breaker-neutral acquire-timeout (a local pool-sizing signal, not a
+The connection-count defaults reproduce clickhouse-go/v2's previously-implicit
+pool sizing verbatim, made explicit so the sharded-pushdown solver can raise the
+ceiling for fan-out rather than inherit a hidden driver default. The
+connection-lifetime default departs deliberately (see the row below) so a stale
+conn to a restarted ClickHouse backend is recycled fast. When the pool is
+exhausted an acquire blocks up to `CERBERUS_CH_DIAL_TIMEOUT` and then fails with
+a breaker-neutral acquire-timeout (a local pool-sizing signal, not a
 ClickHouse-health failure).
 
-| Variable                        | Type     | Default | Description                                                        |
-| ------------------------------- | -------- | ------- | ------------------------------------------------------------------ |
-| `CERBERUS_CH_MAX_OPEN_CONNS`    | int      | `10`    | Total pooled ClickHouse connections (busy + idle). Must be > 0.    |
-| `CERBERUS_CH_MAX_IDLE_CONNS`    | int      | `5`     | Idle ClickHouse connections kept warm for reuse. Must be > 0.      |
-| `CERBERUS_CH_CONN_MAX_LIFETIME` | duration | `1h`    | Max age of a pooled connection before it is recycled. Must be > 0. |
+| Variable                        | Type     | Default | Description                                                                                                                                                                           |
+| ------------------------------- | -------- | ------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `CERBERUS_CH_MAX_OPEN_CONNS`    | int      | `10`    | Total pooled ClickHouse connections (busy + idle). Must be > 0.                                                                                                                       |
+| `CERBERUS_CH_MAX_IDLE_CONNS`    | int      | `5`     | Idle ClickHouse connections kept warm for reuse. Must be > 0.                                                                                                                         |
+| `CERBERUS_CH_CONN_MAX_LIFETIME` | duration | `5m`    | Max age of a pooled connection before it is recycled. Kept short (vs the driver's 1h) so a stale conn to a restarted ClickHouse pod is recycled in minutes, not an hour. Must be > 0. |
 
 ## Query limits and memory
 
--- a/internal/chclient/breaker.go
+++ b/internal/chclient/breaker.go
@@ -503,8 +503,11 @@ func (b *breaker) record(ctx context.Context, err error) {
 }
 
 // observeLevel returns the breaker's CURRENT lifecycle phase as the gauge
-// level (breakerGaugeClosed / _Open / _HalfOpen) plus its stable string label,
-// read under b.mu so the observable-gauge callback gets a tear-free snapshot.
+// level (breakerGaugeClosed / _Open / _HalfOpen), read under b.mu so the
+// observable-gauge callback gets a tear-free snapshot. The level IS the phase
+// — there is no companion string label, because a phase-keyed label would
+// orphan the prior-phase series on every transition and leave it stale (see
+// registerStateCallback).
 //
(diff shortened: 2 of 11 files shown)


tsouza:
## Update: found and fixed the ACTUAL root cause (verified live)

The first commit (stale-conn evict+retry) is a genuine resilience improvement, but it was **not the cause of the chaos `ch-pod-kill` assertion failure**. I stood up the full k3d chaos stack locally and reproduced the failure, then traced it to a second, distinct bug.

### Real root cause: `cerberus_ch_breaker_state` is an orphan-series gauge

Reproduced live: after CH recovers, the breaker genuinely transitions back to CLOSED on every replica (confirmed in the pod logs: `ch circuit breaker recovered: probe succeeded, circuit closed`). Yet the Prom-head read of `cerberus_ch_breaker_state` still returned a `{state="open"}=1` series, so the chaos lane's `max()`-across-series read saw the breaker as permanently OPEN.

The gauge encoded the lifecycle phase as a metric **label** (`state="open"/"closed"/"half-open"`). An OTel **observable** gauge only overwrites series it re-observes each interval. Keying the phase as a label means every transition mints a NEW series and **orphans** the previous one — the old `{state="open"}=1` is never re-observed, so it lingers at its last value over OTLP forever, sitting next to the live `{state="closed"}=0`. `max()` reads 1.

The pre-existing metrics unit test passed only because the in-memory `ManualReader` re-collects a fresh snapshot each call, hiding the push-time orphan behaviour.

### Fix

Drop the `state` label entirely — the numeric level (`0=closed/1=open/2=half-open`) already IS the phase, so the label was pure redundancy. The gauge is now keyed on `head` alone: **one series per head, overwritten in place every interval — it can never go stale.** `observeLevel()` drops its now-unused string return. The metrics test is re-keyed on `head` and hardened to assert NO `state` label and at most one series per head.

### Live verification

Rebuilt the k3d chaos image with this fix and ran the live `ch-pod-kill` scenario against a real CH-pod-delete:
```
main breaker driven to CLOSED (cerberus_ch_breaker_state == 0)
cerberus_ch_breaker_trips_total == 1
cerberus_ch_breaker_state == 0 (CLOSED)      <- the assertion that previously failed
scenario ch-pod-kill: all resilience contracts held
chaos lane: all selected scenarios passed     <- full phase-1 green
```

### What stays from commit 1

The stale-conn evict+retry + `ConnMaxLifetime 1h→5m` is retained — it's correct and reduces real breaker flapping during CH warm-up (fewer spurious trips), it's just not what the chaos assertion was tripping on. Both fixes ship together.

PR #927, opened 2026-06-14
fix(chclient): bound CH-restart recovery with a short ConnMaxLifetime, not a 5m ceiling

tsouza:
## Problem

The informational `chaos` CI lane (`ch-pod-kill` scenario) flakes: after the
ClickHouse pod is killed and rescheduled, one cerberus replica's per-head CH
circuit breaker stays **OPEN for almost exactly 5 minutes** before recovering,
tripping the `cerberus_ch_breaker_state should be 0 (CLOSED)` assertion.

Real CI timeline: CH killed ~21:03:55; one replica's breaker closed at 21:04:41
(CH already healthy), but a **second replica stayed OPEN until 21:09:47** —
process-start + 5m. That 5m == the prior `ConnMaxLifetime` default.

## Root cause

cerberus uses clickhouse-go/v2's **native** `clickhouse.Conn` pool, which has
**no idle-health knob** (v2.46.0). A force-killed CH pod leaves its old socket
half-open (ESTABLISHED, no FIN/RST), so the driver's per-acquire non-blocking
socket check (`conn_check.go`) passes the dead conn through as **healthy** and
hands it out. The next query then blocks on a read against the dead peer until
the driver's `ReadTimeout` (300s) or the request's ctx budget, surfacing as
`context.DeadlineExceeded` — which is *not* a broken-conn error, so it is
neither fast-retried (`withTransportRetry`) nor fast-evicted, and it counts as a
real breaker failure. Recovery is therefore bounded by whatever **ages the stale
conn out** first.

## Fix — TCP keepalive (root cause), ConnMaxLifetime as backstop

The **preferred root-cause fix** is to detect the dead peer at the *socket*
layer. `New()` now dials every CH connection through a `net.Dialer` with
`KeepAliveConfig` (Go 1.26 `net.KeepAliveConfig{Enable,Idle,Interval,Count}`),
so the kernel probes idle sockets and, after `COUNT` unanswered probes, declares
the peer dead. The blocked read then returns **ECONNRESET fast**, which
`isBrokenConnError` classifies → `withTransportRetry` evicts + redials. With the
defaults (Idle 10s, Interval 5s, Count 3) dead-peer detection is **~25s**, well
under the chaos `BREAKER_CLOSE_DEADLINE_MS` (300s) and under 60s, on **every**
replica. Keepalive probes fire **only on idle sockets**, so a long streaming
query is never interrupted. When keepalive is disabled the dialer still dials
(`Enable:false`) — `DialContext` behaviour is uniform.

Four new env knobs (every-knob-is-env), wired through the full config pipeline
with fail-fast validation (idle/interval/count must be > 0 when enabled):

| Variable | Default | Meaning |
| --- | --- | --- |
| `CERBERUS_CH_KEEPALIVE_ENABLED` | `true` | Arm TCP keepalive on CH sockets |
| `CERBERUS_CH_KEEPALIVE_IDLE` | `10s` | Idle before first probe |
| `CERBERUS_CH_KEEPALIVE_INTERVAL` | `5s` | Gap between probes |
| `CERBERUS_CH_KEEPALIVE_COUNT` | `3` | Unanswered probes → socket dead (~25s) |

`ConnMaxLifetime` stays at **30s** as the **age-eviction backstop** if keepalive
is ever disabled; its verbose 5-min-recovery-saga docs are trimmed to a light
field description (the rationale now lives with the keepalive consts/comment).

## Why keepalive, not just the lifetime backstop

A short `ConnMaxLifetime` heals via *unconditional age eviction*, recycling the
whole idle pool on a timer even in steady state. Keepalive heals via *targeted
dead-peer detection* — only the actually-dead socket is torn down, the moment
the kernel proves the peer is gone, independent of how long the conn has lived.
It is the mechanism the maintainer asked for as the real fix.

## Tests

- `internal/config`: `ch_keepalive_test.go` (defaults / overrides /
  disabled-skips-validation / fail-fast invalid) + `viper_test.go` default pins.
- `internal/chclient`: `keepalive_test.go` asserts `New()` builds a working
  dialer with keepalive enabled and disabled (dials a local listener), plus the
  bare zero-Config sanity. Existing `retry_test.go` (incl.
  `TestWithTransportRetry_BlockingStaleConn_NotRetried`) kept.
- `go test -race ./internal/config/... ./internal/chclient/...` green.

## k3d verification

See the verification comment for the recorded recovery timings, including the
**high-`ConnMaxLifetime` variant** (1h lifetime + keepalive on) that isolates
keepalive as the mechanism doing the recovery.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -99,11 +99,11 @@ exhausted an acquire blocks up to `CERBERUS_CH_DIAL_TIMEOUT` and then fails with
 a breaker-neutral acquire-timeout (a local pool-sizing signal, not a
 ClickHouse-health failure).
 
-| Variable                        | Type     | Default | Description                                                                                                                                                                           |
-| ------------------------------- | -------- | ------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `CERBERUS_CH_MAX_OPEN_CONNS`    | int      | `10`    | Total pooled ClickHouse connections (busy + idle). Must be > 0.                                                                                                                       |
-| `CERBERUS_CH_MAX_IDLE_CONNS`    | int      | `5`     | Idle ClickHouse connections kept warm for reuse. Must be > 0.                                                                                                                         |
-| `CERBERUS_CH_CONN_MAX_LIFETIME` | duration | `30s`   | Max age of a pooled connection before it is recycled. Kept short (vs the driver's 1h) because it is the effective recovery bound after a ClickHouse restart: the driver has no idle-health knob and a stale conn to a force-killed pod blocks reads for minutes, so age eviction is what heals the pool — 30s recovers in seconds. Must be > 0. |
+| Variable                        | Type     | Default | Description                                                                                                                                                                                                                                                     |
+| ------------------------------- | -------- | ------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `CERBERUS_CH_MAX_OPEN_CONNS`    | int      | `10`    | Total pooled ClickHouse connections (busy + idle). Must be > 0.                                                                                                                                                                                                 |
+| `CERBERUS_CH_MAX_IDLE_CONNS`    | int      | `5`     | Idle ClickHouse connections kept warm for reuse. Must be > 0.                                                                                                                                                                                                   |
+| `CERBERUS_CH_CONN_MAX_LIFETIME` | duration | `30s`   | Max age of a pooled connection before it is recycled. Kept short (vs the driver's 1h) because it bounds recovery after a ClickHouse restart: a stale conn to a force-killed pod is only aged out by this window, so 30s heals the pool in seconds. Must be > 0. |
 
 ## Query limits and memory
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -99,11 +99,24 @@ exhausted an acquire blocks up to `CERBERUS_CH_DIAL_TIMEOUT` and then fails with
 a breaker-neutral acquire-timeout (a local pool-sizing signal, not a
 ClickHouse-health failure).
 
-| Variable                        | Type     | Default | Description                                                                                                                                                                                                                                                     |
-| ------------------------------- | -------- | ------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `CERBERUS_CH_MAX_OPEN_CONNS`    | int      | `10`    | Total pooled ClickHouse connections (busy + idle). Must be > 0.                                                                                                                                                                                                 |
-| `CERBERUS_CH_MAX_IDLE_CONNS`    | int      | `5`     | Idle ClickHouse connections kept warm for reuse. Must be > 0.                                                                                                                                                                                                   |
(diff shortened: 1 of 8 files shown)


tsouza:
## k3d verification — recovery timings + honest scope of what the local substrate can prove

Verified on local k3d (`cerberus-e2e`, docker 29.5.3, 2 cerberus replicas via HPA `minReplicas=2`, CH on a PVC, chaos overlay `BREAKER_THRESHOLD=3`). Recovery is measured **per replica** by probing each pod's `/readyz` **directly by pod IP** (not through the Service, so a still-OPEN replica can't hide behind round-robin), timed from **t0 = "CH rollout Available"**.

### Shipped defaults (keepalive ON, `ConnMaxLifetime=30s`, `QueryTimeout=2m`) — `kubectl delete pod -l app=clickhouse`, 3 runs

| run | replica A | replica B |
| --- | --- | --- |
| 1 | +2s | +5s |
| 2 | +1s | +1s |
| 3 | +2s | +2s |

Breaker tripped OPEN (503 + `Retry-After:5`, `/readyz` 503 with `circuit`) on **both** replicas every run, then both healed **within 1–5s of CH coming back** — deterministically **well under 60s** and under the chaos `BREAKER_CLOSE_DEADLINE_MS` (300s). No flake across 3 runs.

### Keepalive-isolation variant (keepalive ON, `ConnMaxLifetime=1h`) — pod-kill, 3 runs

With age-eviction effectively disabled (1h lifetime), recovery was **still 1–6s** on both replicas across 3 runs — so the heal is **not** coming from `ConnMaxLifetime`. This isolates the socket-layer mechanism (keepalive + fresh-dial on the half-open probe) as what closes the breaker.

| run | replica A | replica B |
| --- | --- | --- |
| 1 | +1s | +2s |
| 2 | +2s | +2s |
| 3 | +3s | +6s |

### Honest scope: the original 5-minute *blocking-read* failure mode does NOT reproduce deterministically on local k3d

I attempted to reproduce the exact original flake (a half-open `ESTABLISHED`-no-FIN/RST socket whose read blocks for minutes) using three fault models — `kubectl delete pod`, a deny-egress `NetworkPolicy` blackhole (confirmed **enforced** on this k3d: egress to CH:9000 hung for the full client timeout, not refused), and the blackhole **with the production 2m `QueryTimeout`** to un-mask the chaos overlay's neutralising 5s timeout. In **every** keepalive-**OFF** control run, recovery was *also* fast (1–9s) and the breaker barely tripped:

- A `kubectl delete pod` triggers CH's graceful teardown, so old conns get RST/FIN and fail *fast* (broken-conn → `withTransportRetry` evicts + redials) **without** needing keepalive.
- Under the NetworkPolicy blackhole, cerberus's existing `withTransportRetry` + readiness-ping path + the per-acquire socket check already absorb most blocked reads, and with the chaos overlay's 5s `QueryTimeout` a blocked read is cancelled neutrally before it can stack breaker failures.

So the original 5-minute hang was a **rare, specific timing alignment** (a stale conn passing the cheap socket check *and* handed to a 2m-budget blocking read), not something the local substrate reproduces deterministically. The keepalive-OFF control therefore can't be driven into the slow state here to show a clean before/after on k3d.

### Conclusion

What the k3d runs **do** prove, repeatably: with the shipped config the breaker trips OPEN on both replicas under a CH outage and **recovers deterministically in 1–5s (<60s) across all runs**, and the 1h-lifetime variant shows that recovery is **not** dependent on `ConnMaxLifetime` age-eviction. The keepalive dialer is exercised on every CH connection (unit-tested: `New()` dials a live socket with keepalive enabled and disabled). The socket-layer mechanism is the kernel-level analogue of the fast-RST path the controls already heal through; `ConnMaxLifetime=30s` remains the unconditional age-eviction backstop. The 30s backstop keeps the informational chaos lane green regardless.

PR #949, opened 2026-06-16
perf(prom): bound resource-attr query memory + e2e OOM diagnostics

tsouza:
Nightly e2e on the rc.5 commit failed at "Assert zero cerberus restarts": cerberus was **OOMKilled** (2 restarts). Deep-investigated — a **real, bounded memory regression** from #947, not a flake.

## Root cause (profiled)
The rc.5 merge `mapUpdate(sanitize(ResourceAttributes), sanitize(Attributes))` lowers as a Project **directly above the Scan leaf**, so ClickHouse materializes a sanitized-merged `Map` per scanned row and cerberus decodes the larger map per result row. Measured (chDB full handler stack, 7 resource keys): **2.36MB vs 1.43MB/op (~+65% heap)**. Dominant allocators: per-row `NormalizeLabelMap` + `CanonicalKey` + `WithMetricName`. **Bounded per-query (no leak)** — but the amplification under concurrent dashboard load tipped the 1Gi pod into OOM, and implies a production memory cost.

## Fix
- **Optimised the avoidable transient:** the cursor stamps each `Sample` with a reflect-free `SeriesID`; the prom matrix/vector pivots **memoise wire-label normalization by `(SeriesID, MetricName)`** → pivot bench **7.98MB → 3.59MB/op (−55% heap, −44% allocs, −60% wall)**.
- **Right-sized the genuine cost:** e2e pod limit **1Gi → 1536Mi**, `GOMEMLIMIT` **800→1228MiB**; documented the production memory expectation in `docs/operations.md` (size memory / trim via `CERBERUS_PROM_RESOURCE_LABELS`).
- No SQL emission change → no goldens; feature + locked precedence/compat behavior untouched.

## Diagnostics (the "missing diag dumps" gap)
- Gated `net/http/pprof` via `CERBERUS_DEBUG_PPROF` (off by default; on in the e2e ConfigMap).
- Restart gate rewritten as `.github/scripts/e2e-cerberus-restart-gate.mjs`: surfaces `lastState.terminated` (**OOMKilled** loud — `--previous` is empty for OOM), `resources.limits` + `GOMEMLIMIT`, `kubectl top`, and pulls `/debug/pprof/heap` → CI artifact.

Verified: build (default + `-tags chdb`), golangci-lint v2.12.2 + go-arch-lint clean, prom/chclient/config/cmd tests green (incl `-race`).


tsouza:
## Follow-up fix: cross-shard SeriesID aliasing (was: 136 prometheus-compat diffs)

The label memo added here trusts `SeriesID` as a per-cursor series identity. The chclient `rowsCursor` upholds that, but the solver's composed `shardCursor` (route B) did **not**: it concatenates K child cursors, each numbering its series independently from 1, and re-interned the label *map* across shards but never re-stamped `SeriesID`. Under `Mode=auto` (the production/env default — unset `CERBERUS_EVAL_ROUTE` flips to `auto`, not the library `single` default), eligible plans route to the composed cursor, so the memo aliased two distinct series that shared a per-shard ordinal → one series' samples bucketed under another's labels.

This is why unit tests stayed green (in-process `DefaultConfig` = `Mode=single` = one cursor) while the env-driven compat binary went **574/574 → 438/574 (136 value diffs)**.

**Fix (commit cc756c01):**
- `shardCursor.reintern` now also assigns a stable cross-shard ordinal (off the same canonical key it already computes for label dedup) and `Next` stamps `s.SeriesID` — the composed cursor becomes one coherent SeriesID namespace.
- `fetchSeries` (`/series`) drops the memo: it folds several independent per-chunk/per-variant queries (each its own SeriesID namespace) and the memo bought nothing there anyway (one label set per series, already deduped).
- Corrected the `labelMemo` doc + added `TestExecute_CrossShardSeriesIDBijective` and `TestMatrixFromCursor_MemoPreservesBucketing`.

**Validated locally:** prometheus-compat **574/574, 0 diffs** with the memo ON, in both `Mode=auto` and forced `Mode=sharded`.

PR #1057, opened 2026-06-24
feat(solver): per-deployment self-tuning route calibration (generic loop, local constants)

tsouza:
## What

Per-deployment **self-tuning** of the route A/B (single vs time-slice-sharded) thresholds. The router's hand-tuned thresholds (`MinFanout`, `MinAnchorPairs`) are now calibrated **at runtime from each deployment's own** `cerberus_router_corpus` — the Stage-0 corpus (#1053) of observed cost.

## Architecture — generic ships, local is learned

- **GENERIC (shipped to every deployment):** the calibration logic (`internal/solver/calibrate.go`), the background loop (`internal/solver/selftune.go`), and the conservative `DefaultConfig`. Deployment-independent — **no** deployment's learned constants are baked into the shipped code. A grep of the diff confirms the only deployment-specific data path is the corpus reader; there are no hardcoded deployment-specific values.
- **LOCAL (never shipped):** the calibrated constants, derived in-memory from THIS deployment's corpus. this deployment's data tunes itself only; another install tunes itself. A single corpus deliberately never leaks into shipped defaults (avoids cross-deployment over-fit).

This is **optimizer self-tuning, not result caching** — it changes which plan shape the router picks, never memoizes results.

## Safety rails (each tested)

- **Conservative-floor / tighten-only:** calibrated thresholds may only move in the SAFER direction (lower → shard more readily near the OOM frontier; PARQO asymmetric penalty). Calibrate never *raises* a threshold and never lowers past the shipped floor.
- **Fail-open / no-op without signal:** thin corpus, no `below-threshold` decisions, or no OOM/cost-danger exemplars → `defaults` returned **verbatim**. a no-signal corpus (all route-A, below-threshold=0, no failures) is a true no-op — pinned by `TestCalibrate_NoOpOnNoSignalCorpus`.
- **Off by default:** gated behind `CERBERUS_ROUTE_SELFTUNE` (default off). Unset → static shipped Config, loop never starts, **zero behavior change** for existing deployments (`TestStartRouteSelfTune_OffByDefault`).

## Mechanics

- `Planner.cfg` is now an `atomic.Pointer[Config]`; the `Plan` path reads it lock-free. The loop atomically swaps a freshly calibrated Config in.
- Reader (`internal/optcorpus/frontier.go`) runs a few aggregate SELECTs per `(N,F[,D])` bucket — never all rows — under the same rate-limited / deprioritised / read-only / failure-open discipline as the query_log source.
- Loop derives its ctx from a cancelable parent and cancels in-flight reads on shutdown (mirrors + fixes the `breaker_recovery.go` ctx shape; goleak-clean).
- Every recalibration logs the active thresholds + a `CalibrationReport` (sample count, what moved vs default and why, detected frontier) so operators can SEE what their deployment self-tuned to.

## Tests (assert-based)

Calibrate purity + determinism (order/mutation-independent); no-op-without-signal; clear-frontier tightens but never past the floor; tighten-only invariant across a fuzz of corpora; atomic swap under `-race`; fail-open on reader error; clean shutdown (goleak, ctx cancels in-flight). `go build ./...`, scoped golangci-lint, go-arch-lint, gofumpt -extra all clean.

### DIFF
--- a/cmd/bench-report/matrix_chdb.go
+++ b/cmd/bench-report/matrix_chdb.go
@@ -181,7 +181,7 @@ func measureMatrixCell(ctx context.Context, s *session, iters int, st matrixStra
 	if err := cfg.Validate(); err != nil {
 		return matrixCell{}, fmt.Errorf("solver config: %w", err)
 	}
-	pl := &solver.Planner{Cfg: cfg}
+	pl := solver.NewPlanner(cfg)
 	gs, ge, gstep := solver.GridOf(plan)
 	dec, routed := pl.Plan(plan, solver.RequestMeta{
 		Lang: solver.LangPromQL, Start: gs, End: ge, Step: gstep,
--- a/cmd/bench-report/sharded_chdb.go
+++ b/cmd/bench-report/sharded_chdb.go
@@ -164,7 +164,7 @@ func measureSharded(s *session, iters int) (shardedResult, error) {
 	if err := cfg.Validate(); err != nil {
 		return shardedResult{}, fmt.Errorf("sharded config invalid: %w", err)
 	}
-	pl := &solver.Planner{Cfg: cfg}
+	pl := solver.NewPlanner(cfg)
 	gs, ge, gstep := solver.GridOf(plan)
 	dec, routed := pl.Plan(plan, solver.RequestMeta{
 		Lang:  solver.LangPromQL,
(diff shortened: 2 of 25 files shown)


tsouza:
Closing as superseded by the router-rules catalog (#1060/#1062).

Investigation confirmed this PR is purely the self-tuning subsystem — its own parallel corpus reader plus the calibrate/selftune loop and the atomic.Pointer planner refactor. The corpus write-sink and the catalog's own reader already live on main independently, so nothing here is needed by the catalog.

The catalog is the better primitive: declarative, growable, broad, and human-in-the-loop. The route A/B auto-tuning this PR automates is low-value for the observed workload (dominated by non-shardable / instant queries; below-threshold count ~0). If silent auto-tuning is ever wanted, the clean path is a thin executor that consumes catalog rules tagged auto-apply — not this parallel subsystem.

Branch kept; reopenable if we revisit.

PR #1113, opened 2026-06-27
perf(traceql): fold the search window into every compound search shape (GAP-3)

tsouza:
## Problem (resource-bound audit — GAP-3)

TraceQL **compound** searches (`&&`/`||`, structural `>>`/`<<`/`&>>`, `select(nestedSet*)`) emitted **no `Timestamp` predicate at all**. The request window was folded into SQL only by `stampSearchTraceLimit → andWindow`, which matches only the *plain*-search shape; every other shape **scanned full retention** on `/api/search` (and could return out-of-window spans).

## Fix

`stampSearchWindow`: a lowering pass that ANDs the request window (`Timestamp BETWEEN start AND end`) into every **leaf** `Filter/Scan` of a search row source, generalizing #1110's `pushBoundedTraceGate` into a shared `pushLeafPredicate` (now also descends `NestedSetAnnotate.Input`).

- Runs **last**, after `stampSearchTraceLimit`, and does **not** descend the already-windowed plain `SearchTraceLimit` node → plain search stays byte-identical (**no double-fold**).
- Gated on `limit > 0` — the exact "`/api/search`, not metrics" discriminator (only the search handlers set `WithSearchTraceLimit`).

### Why it's safe where it counts

The window reaches only chplan **leaves**. The structural closures' recursive `t`-scan and the nested-set numbering CTE (`_cerberus_ns_paths`) are **emitter-synthetic** (not chplan children), so they're physically unreachable by the pass and correctly stay **unwindowed**:
- the closure inherits the window via the #77 seed re-render of its now-windowed leaf (so it's bounded for free);
- the numbering must number **whole traces** regardless of the window (Tempo numbers at ingest) — windowing the anchor would collapse numbering to `0/0/0`.

## Validation

- **`just compat-traceql` = 48/48 PASS, 0 diffs, 0 hard errors** (100%) — cerberus matches reference Tempo on every case *with* the fold. The fold is a no-op on the in-window corpus while bounding the scan.
- Proven: the numbering CTE stays unwindowed; plain search not double-folded; metrics path guarded (`limit==0` → no fold); **zero** existing-golden churn (the ctx-gated pass is a no-op unless a window is set).

## Tests

- 3 new windowed fixtures (compound `&&`, structural `>>`, `select(nestedSet*)`) pinning the leaf window + the unwindowed numbering CTE;
- unit assertions: compound-leaf windowing, the `limit>0` metrics guard, plain-search no-double-fold;
- a `-- search_window --` fixture seam.

## Out of scope (noted)

For the #1110 structure-tab shape, the **top-N trace ranking** stays newest-ever (a separate #1110 follow-up to rank-in-window). GAP-3 only adds the row-source window (`top-N ∩ window`), which is *more* correct for the common recent-window case. The per-span vs per-trace-overlap boundary difference is pre-existing (the shipped plain path already uses the same per-span window) and corpus-invisible.


tsouza:
Updated after the adversarial review:

- **Finding B (blocking, fixed in 2f22ef8):** the review caught that folding the window onto the #1110 structure-tab leaves, while the top-N ranking stayed newest-*ever*, would return an **empty** structure tab on a historical window (`topN_ever ∩ window = ∅`). Fixed by windowing the top-N ranking itself (`boundedRootScopeFrag` + `BoundedTraceScope.Window*` / `NestedSetAnnotate.Window*`, threaded identically to both the numbering scope and the leaf gate). New chDB roundtrip fixture pins it (historical window excluding the newest 2 of 4 traces → returns the 2 in-window, not empty).
- Fixed 3 pre-existing handler chDB tests (probe job) that seeded fixed historical data + searched windowless — their `select()` shapes were silently ignoring the window and now correctly respect it.
- Re-validated: `just compat-traceql` **48/48 PASS, 0 diffs**; full `test-chdb` lane green.

The 'top-N ranking stays newest-ever' caveat in the PR description is now resolved (it's windowed).

PR #1143, opened 2026-06-29
test(parser): mutation coverage for the clean-room LogQL/TraceQL parsers

tsouza:
## Problem

The de-AGPL clean-room TraceQL parser (`internal/traceql/ast`) landed without mutation coverage. With ~250 new mutable lines and no tests pinning them, the `gremlins phase4-traceql` lane (`./internal/traceql @ 95%`) dropped to **90.02%**, failing the efficacy gate on the v1.8.3 release PR (#1137). The `phase4-logql-*` lanes were already green and needed no work.

## Fix

Targeted `*_mutation_test.go` suites (pure unit tests, no chDB) that pin the behaviour each surviving mutant would break. Every kill was verified by hand-applying the mutation, confirming the test fails, then reverting.

| scope | before | after |
| --- | --- | --- |
| `gremlins phase4-traceql` (`./internal/traceql`) | **90.02%** (424 killed / 47 lived) | **96.51%** (443 killed / 16 lived) |

### What the new tests pin

- **enum** — `String()` spelling of every `StaticType` / `Operator` / `AttributeScope` / `Intrinsic` / `Status` / `Kind` / `AggregateOp` / `MetricsAggregateOp` / `SecondStageOp`, plus the bare-identifier → intrinsic resolution (the `continue`-vs-`break` loop, the `parent`/`none` refusal).
- **static** — the typed `(value, ok)` accessors, numeric coercion, value equality (incl. numeric-vs-non-numeric coercion traps), array iteration, and the surface encoder for every type incl. the float `.0` suffix.
- **attribute** — the `NewScopedAttribute` intrinsic-resolution guard, `impliedType` per intrinsic, and the `String()` prefix rules (dotted user attr vs bare intrinsic vs scoped/parent).
- **parser** — operator precedence + associativity (field arithmetic, `^` right-assoc, spanset left-assoc, scalar expr), trace-id operand normalization on either side, and topk/bottomk dispatch.
- **lexer** — duration-suffix scanning, Prometheus-first units (`1w`/`1d`), and exact suffix consumption (no over-read).
- **rewrite** — the four array-fold rules, attribute-on-either-side extraction, and the no-fold cases (mismatched attr/operator, non-string regex operands).
- **pipeline / metrics** — `RootExpr` round-trip rendering of the metrics first/second stages, the `with(...)` hint clause, quantile precision, the chained-second-stage separators, and the metrics accessors.
- **search_limit** (margin) — `WithSearchWindow` storing a half-open window, and `stampSearchTraceLimit` wrapping a plain-search source (the summaries-drain OOM guard).

### Remaining survivors (genuinely equivalent — documented, not skipped)

The 16 survivors at 96.51% cannot be killed by any reachable input:

- **Defensive bounds that never fire on valid input**: the `cursor.advance()` `pos < len-1` guard (`parser.go:43`), the scope-prefix length cap (`lexer.go:517`), the always-true chained-separator index guard (`metrics.go:225`).
- **Redundant control flow**: the EOF/space `break` (`lexer.go:543`) is subsumed by the following non-duration-rune `break`; the metrics-pipeline `break`→`continue` (`parser.go:260`) is a no-op because the loop's `tokPipe` condition exits anyway.
- **Unreachable distinguishing inputs**: the `ParseIdentifier` `||` (`parse.go:57`) and the `Hints.String()` nil/empty guard (`pipeline.go:40`) — the parser never produces the inputs that would separate the branches.
- **Caught downstream**: the array-fold loop `continue`→`break` and regex `restrict` checks (`rewrite.go:112/115/116`) — no later fold rule can succeed after the skip, and `staticMerge` rejects the mismatch anyway; the single-nil equality guard (`static.go:216`) falls through to a type-mismatch `false` regardless.
- **Pre-existing**: 3 `lower.go` `CONDITIONALS_NEGATION` survivors predate this PR (the non-AST lowering set was already green at ~95.6% before the clean-room parser existed). Left for a separate follow-up; the combined lane is well over the bar without them.

Validated locally with the `tsouza/gremlins` fork (`gremlins unleash ./internal/traceql`); CGO + libchdb. The `mutation` lane is skipped on non-`release/*` PRs, so this branch's required `mutation` check passes through green; the real gate is exercised when these land on the release PR.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/internal/traceql/ast/attribute_mutation_test.go
+++ b/internal/traceql/ast/attribute_mutation_test.go
@@ -1,0 +1,133 @@
+package ast
+
+import "testing"
+
+// Mutation-coverage tests for attribute.go: intrinsic resolution rules and the
+// per-intrinsic implied type.
+
+// TestNewScopedAttributeIntrinsicResolution pins the guard that only the bare,
+// unscoped, non-parent form resolves an intrinsic. The guard is
+// `scope == AttributeScopeNone && !parent`:
+//   - negating the first clause makes the scoped form resolve an intrinsic;
+//   - turning `&&` into `||` makes the scoped OR parent form resolve one.
+//
+// Each row below distinguishes one of those mutations.
+func TestNewScopedAttributeIntrinsicResolution(t *testing.T) {
+	t.Parallel()
+
+	// Bare unscoped non-parent: "duration" IS the intrinsic.
+	if got := NewScopedAttribute(AttributeScopeNone, false, "duration").Intrinsic; got != IntrinsicDuration {
+		t.Errorf("unscoped duration intrinsic = %v; want IntrinsicDuration", got)
+	}
+	// Scoped: must NOT resolve (kills `scope == None` negation and `&&`→`||`).
+	if got := NewScopedAttribute(AttributeScopeSpan, false, "duration").Intrinsic; got != IntrinsicNone {
+		t.Errorf("span.duration intrinsic = %v; want IntrinsicNone", got)
+	}
+	if got := NewScopedAttribute(AttributeScopeResource, false, "name").Intrinsic; got != IntrinsicNone {
+		t.Errorf("resource.name intrinsic = %v; want IntrinsicNone", got)
+	}
+	// Parent-qualified unscoped: must NOT resolve (kills the `!parent` clause
+	// and `&&`→`||`).
+	if got := NewScopedAttribute(AttributeScopeNone, true, "duration").Intrinsic; got != IntrinsicNone {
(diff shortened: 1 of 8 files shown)


tsouza:
Traceability: the deferral in this PR body ("3 `lower.go` `CONDITIONALS_NEGATION` survivors … Left for a separate follow-up") is now tracked as #1524. The count has since grown from 3 to 7 — CI mutation run 30695961440 lists them at `internal/traceql/lower.go:651, 655, 675, 682, 1338, 1409, 1484`.

Issue #1168, opened 2026-07-01
Histogram metrics do not work


Jannik2099:
### Cerberus version

v1.8.4

### ClickHouse version

26.5

### Which query language?

PromQL

### Query (or curl reproducer)

```shell
`any_prometheus_histogram_metric`
```

### What did you expect?

I have a bunch of histogram metrics, both in `otel_metrics_histogram` and `otel_metrics_exponential_histogram`.

Neither show up at all in Grafana.

### What did cerberus return?

```shell
nothing
```

### Schema overrides (if any)

```sql

```

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/tsouza/cerberus/blob/main/CODE_OF_CONDUCT.md).

tsouza:
Please, try https://github.com/tsouza/cerberus/releases/tag/v1.9.0 and see if it fixes.

Basically, we do not have such type of metric so we did not catch it.

Jannik2099:
No, histograms are still missing in 1.9.0

tsouza:
> No, histograms are still missing in 1.9.0

I will need more context to debug as I was not able to reproduce this issue in 1.9.

Can you please attach here the cerberus logs and also the full otel schema that you have? Additionally, the helm chart values deployed and/or cerberus settings deployed.

tsouza:
Hi, due to inactivity I am closing out this.

Please, feel free to reopen it if the problems persists and you have more context to provide so I can properly debug.

PR #1171, opened 2026-07-01
fix(prom): histograms fail on stock OTel-CH exporter deployments

### LINKED ISSUE #1167: Exponential Histogram table name is incorrect
### Cerberus version

v1.8.4

### ClickHouse version

26.5

### Which query language?

Not applicable (HTTP / deploy / build issue)

### Query (or curl reproducer)

```shell
-
```

### What did you expect?

The clickhouse exporter defaults to `otel_metrics_exponential_histogram`, NOT `otel_metrics_exp_histogram` - this is a documentation bug on their end. See

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/fd0f261a4b3ddd82c46bb5b8915faaefb9bc5828/exporter/clickhouseexporter/config.go#L89

### What did cerberus return?

```shell
-
```

### Schema overrides (if any)

```sql

```

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/tsouza/cerberus/blob/main/CODE_OF_CONDUCT.md).

tsouza:
## What this delivers

- **Exp-histogram table.** cerberus's exponential-histogram table default and its schema-provision **readiness probe** use `otel_metrics_exponential_histogram` — the table the upstream OTel-CH `clickhouseexporter` creates (`config.go`: `otel_metrics` + `_exponential_histogram`). On a stock exporter deployment cerberus becomes ready and a native `histogram_quantile` resolves; verified end-to-end against the real exporter schema (`/readyz` 200, classic `histogram_quantile` returns the correct value).
- **Columnar decode.** With `columnar_result_decode` enabled, decoding the exporter's real `Map(LowCardinality(String), String)` label maps — which ch-go's `Auto` inference cannot construct — falls back to the row decoder and returns 200 rather than a 502. The columnar path is an optimisation, never a correctness requirement, so the fall-back does not trip the breaker.
- **Schema-drift guard** (`e2e-bwc-verify-placement.mjs`). The E2E fails if the collector creates any `otel_metrics_*` base table cerberus does not read — i.e. if cerberus's read-side table-name defaults ever drift from the exporter's.

## Tests

- `internal/chclient/columnar_integration_test.go` — CI-executing (`chdb.yml` → `just chclient-integration`). A `Map(LowCardinality(String), String)` fixture drained through the columnar path, asserting full-count row↔columnar parity. Revert-to-red proven: remove the fall-back → the test fails with the exact inference error; restore → green.
- `internal/api/prom/handler_histogram_integration_test.go` — local (`-tags=integration`). The full Prometheus HTTP path against the real exporter schema (LowCardinality label maps, `otel_metrics_exponential_histogram`, columnar ON): native exp quantile, classic `_bucket`/`_count`/`_sum` discovery + reconstruction, and the columnar → row fall-back at 200.

## Scope

Closes #1167. Addresses #1168 in part. One classic-histogram gap remains as a **separate follow-up**: regex / empty `__name__` matchers (`{__name__=~".+"}`, Grafana Metrics-Drilldown's `match[]=…`) don't reach the histogram table — its `Count`/`Sum`/`BucketCounts` row shape is disjoint from the `merge()` Sample contract the unknown-name fan-out uses, so surfacing it needs the per-arm `UnionAll` lowering (an architectural change). #1168 stays open pending the reporter's `CERBERUS_CH_OPTIMIZATIONS` (to confirm whether columnar decode is enabled) and their Grafana surface.


Jannik2099:
> https://github.com/tsouza/cerberus/issues/1167's readiness fix is the most likely explanation for the reporter's "neither show up at all in Grafana" (an unready gateway blanks the whole datasource)

Note that I am not seeing any histograms even after overriding the table name, and after cerberus confirms readiness. Other metric types do show up fine.

Jannik2099:
> With columnar_result_decode enabled, decoding the exporter's real Map(LowCardinality(String), String) label maps — which ch-go's Auto inference cannot construct — falls back to the row decoder and returns 200 rather than a 502. The columnar path is an optimisation, never a correctness requirement, so the fall-back does not trip the breaker.

Afraid this ain't the root cause either. My cerberus runs with

```
msg="clickhouse optimizations resolved" selection=auto mode=enforcing server_version=26.5 server_ts_grid_capability=available enabled=aggregation_in_order,condition_cache,ts_grid_changes,ts_grid_range,ts_grid_resample,ts_grid_resets
```

, so with `columnar_result_decode` disabled - unless I misunderstood the analysis

PR #1174, opened 2026-07-01
test(e2e): add classic-histogram telemetrygen collector-write-path check

tsouza:
## What

Adds a `telemetrygen-classichist` sidecar to the compose E2E, the explicit-bucket sibling of the existing `telemetrygen-exphist` (exponential) source. This gives the compose-smoke lane symmetric coverage of the OTel collector's **classic (explicit-bucket) histogram write path**, end to end.

## Why

The exponential source validates that the collector's clickhouseexporter write path for `otel_metrics_exponential_histogram` is not just *shaped* right (readiness proves the table exists) but *readable* through cerberus's native quantile. The classic histogram shape — real OTLP `Histogram` → exporter writes `otel_metrics_histogram` → cerberus reconstructs `_bucket`/`le` series → `histogram_quantile` — had no equivalent live write-path check. This closes that gap so both histogram shapes the collector emits are exercised through the real exporter, not only the deterministic direct-INSERT seed.

## What it adds

- **`docker-compose.yml`** — `telemetrygen-classichist` service: same image/version, `depends_on`/`restart: on-failure` wait pattern as `telemetrygen-exphist`, but `--metric-type Histogram` emitting a distinct `e2e_latency_classic_hist` so it lands in `otel_metrics_histogram` and never collides with the deterministic showcase seed or the exponential source.
- **`.github/scripts/e2e-verify-classic-histogram.mjs`** — mirrors the exp verifier. Polls cerberus's Prom API for `histogram_quantile(0.9, rate(e2e_latency_classic_hist_bucket[5m]))` (the classic `_bucket`-rate idiom cerberus uses, matching the showcase classic panel) until it returns a **finite** value. Assertion-tolerant — telemetrygen data is nondeterministic, so it asserts *finite*, not an exact value. `node:` builtins only; `::error::`/`::notice::` + `process.exit`; reuses `lib/gh.mjs`. Quantile, poll interval, rate window are named consts (no magic constants).
- **`.github/workflows/e2e.yml`** — runs the new verifier in the same compose-smoke lanes as the exp one.
- **`.github/scripts/README.md`** — inventory entry with env contract.

## Constraints honoured

- Does **not** touch `test/e2e/seed` or the showcase/dashboard panels — those need exact seeded values; the nondeterministic telemetrygen feed lives only in this tolerant verifier lane.
- No `t.Skip`/soft-assert/tolerate; no magic constants; no `unsafe.Pointer`.

## Local validation

- `node --check` on the new `.mjs`; ran it against no server to confirm it enters the tolerant retry loop.
- `docker compose config` resolves; `docker compose config --services` shows both `telemetrygen-classichist` and `telemetrygen-exphist`.
- `actionlint .github/workflows/e2e.yml` clean.
- Pre-push gate (golangci-lint / markdownlint / discipline greps) passed on push.

### DIFF
--- a/.github/scripts/README.md
+++ b/.github/scripts/README.md
@@ -372,6 +372,21 @@ wrapper, plus `appendStepSummary` / `setOutput` for the runner files.
   - Env: `CERBERUS_URL` (default `http://localhost:8080`), `EXP_HIST_METRIC`
     (default `e2e_latency_exp_hist`), `POLL_SECONDS` (default `120`).
   - Exit: `0` the quantile returned a finite value, `1` if not within budget.
+- **`e2e-verify-classic-histogram.mjs`** — `e2e.yml`, the `compose-smoke-shard`
+  (required) + `compose-smoke-shard-info` (informational) jobs, run after
+  `docker compose up --wait`. Explicit-bucket sibling of
+  `e2e-verify-exp-histogram.mjs`: asserts the OTel collector's classic
+  (explicit-bucket) histogram write path is queryable end to end. A second
+  telemetrygen sidecar feeds OTLP classic `Histogram` points named
+  `e2e_latency_classic_hist` through the collector into the exporter-created
+  `otel_metrics_histogram` table, and this module polls cerberus's Prom API for
+  `histogram_quantile(0.9, rate(e2e_latency_classic_hist_bucket[5m]))` until it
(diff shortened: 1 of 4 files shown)


tsouza:
Superseded by #1175 — histogram verification moved into the Playwright E2E suite instead of bespoke .mjs.

PR #1201, opened 2026-07-04
feat(tempo): extend the structural two-phase split to the gRPC Search path

tsouza:
Closes the last two-phase asymmetry: the split was **HTTP-only**, so a recursive structural query HTTP bounds to top-N traces drained in *full* over gRPC (the byte budget fail-closed it with `ResourceExhausted` rather than OOMing — but gRPC still rejected queries HTTP serves).

### The fix: one shared entry point
Collapses HTTP's inline parse→gate→execute into **`Handler.SearchResult(ctx, query, limit)`** — `lang.Parse` → `structuralTwoPhaseTarget` → `runStructuralTwoPhase`-or-`QueryPlan`. **Both transports call it**, so the two-phase eligibility decision lives in exactly one place and the transports **physically cannot diverge again** — which is the whole bug.

### Why eager is a drop-in (no memory regression)
gRPC already fully buffered into a `[]Sample` before shaping (`toTraceSummaries` groups spans by TraceID), so it never streamed on the Go heap. Dropping `QueryCursor`/the drain loop/`cursor.Close` for `res.Samples` keeps the same memory profile — and two-phase *shrinks* it to the response traces. The byte budget still charges the eager drain.

### A≡B across transports, by construction
The ctx folds (`WithSearchTraceLimit`/`Window`/`DrainByteBudget`) are already identical on both heads, so the same plan flows through the same gate. Parity is structural, not empirical: there's no second copy of the gate to drift.

### Also
- Cancellation now surfaces at the query boundary (not per-row); `mapEngineError` maps `context.Canceled`/`DeadlineExceeded` to the proper gRPC status instead of `Internal`.
- Tests: cross-transport routing non-vacuity (a gRPC structural query fires phase A), ctx-error mapping, and the cancellation test rewritten to clean stream termination (the gRPC-owned cursor is gone).

Holding for adversarial review.

### DIFF
--- a/internal/api/tempo/grpc/map_engine_error_test.go
+++ b/internal/api/tempo/grpc/map_engine_error_test.go
@@ -44,8 +44,12 @@ func TestMapEngineError_ContextErrors(t *testing.T) {
 		want codes.Code
 	}{
 		{context.Canceled, codes.Canceled},
-		{context.DeadlineExceeded, codes.DeadlineExceeded},
+		// Deadline + CH wall-clock timeout both map to Unavailable (the gRPC 503),
+		// symmetric with HTTP classifySearchErr which maps both to 503.
+		{context.DeadlineExceeded, codes.Unavailable},
+		{chclient.ErrQueryTimeout, codes.Unavailable},
 		{fmt.Errorf("engine: execute: %w", context.Canceled), codes.Canceled},
+		{fmt.Errorf("engine: execute: %w", chclient.ErrQueryTimeout), codes.Unavailable},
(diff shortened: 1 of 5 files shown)


tsouza:
Traceability: the two "Out of scope (noted, separate follow-ups)" items in this PR are now tracked as #1531 (InspectedTraces means trace count on gRPC, sample count on HTTP) and #1532 (HTTP `classifySearchErr` lacks a `context.Canceled` case). Both verified still live on `origin/main`.

PR #1215, opened 2026-07-16
feat(solver): register VectorJoin slice-invariant; fail-close instant joins on route A

tsouza:
## What

Registers the PromQL `VectorJoin` node as **slice-invariant** so step-aligned vector-vector joins — ratio shapes like `sum(rate(a[5m])) / sum by(job)(rate(b[5m]))` — route through the solver's sharded **route B** instead of falling to route A with `decision_reason="not-sliceable"`.

## Why (live-prod evidence)

Querying the prod router-calibration corpus (`otel.cerberus_router_corpus`) surfaced that **route B has never fired in 3+ weeks of `ModeAuto`** — every one of ~137K observed queries fell into a rejection bucket. Investigating the buckets:

- `high-D` (the closest-to-eligible bucket) is a **correct, permanent** geometric rejection: for `rate(x[range])` where the selector window spans the whole panel, `D ≈ OuterRange`, so `floor(OuterRange/max(D,Step)) < 2` and K-way time-slicing buys zero I/O reduction. Not fixable, not a bug.
- `below-threshold` is correctly gated (genuinely cheap queries below `MinFanout`/`MinAnchorPairs`).
- **`not-sliceable` join shapes are the one real, addressable miss.** `cerb:project;agg=1;rw;join` (215 rows, avg fanout 31, N×F≈50K) and `cerb:project;agg=2;rw;join` (59 rows) clear every gate that matters and have healthy K-clamp geometry — they fall to route A **purely** because `VectorJoin` was never registered in `chplan.sliceInvariantKinds`. The registry doc comment already scoped this as deliberately-deferred, one-node-family-per-PR work.

## Slice-invariance proof

Each output row of a **step-aligned** join reduces exactly one anchor's window on each arm: the emitter ANDs `L.TimestampColumn = R.TimestampColumn` into the ON clause and adds the anchor timestamp to each side's `GROUP BY`. So evaluating over a disjoint sub-grid of anchors (with correspondingly bounded arm scans) yields exactly the rows route A produces. Holds across **all** matching (`on`/`ignoring`), **all** cardinalities (`group_left`/`group_right` — the many-to-one `throwIf(uniqExact>1)` dedup and `Include` `mapConcat` are per-`(match-key, anchor)` because the anchor timestamp is in the join key), and all ops incl `bool`.

## Fail-closed boundary

The registry is keyed by node kind, so it **also** admits the instant-mode (`StepAligned==false`) join — whose emitter synthesizes the join-side timestamp with `now64(9)`, a wall-clock that diverges across shards and never reaches the plan-level now64 scanner. That shape is held on route A by an explicit planner guard (`sawInstantVectorJoin` → `ReasonInstantJoin`), wired into **both** `walkNode` and `walkScalarInterior` — the latter closes a hole an adversarial review caught, where an instant join buried in a `scalar(sum(a)/sum(b))` interior would otherwise route B and replicate a per-shard wall-clock scalar. `VectorSetOp` / `NaryVectorSetOp` (and/or/unless) remain deliberately absent — each is its own future PR.

## Changes

| File | Change |
|---|---|
| `internal/chplan/reanchor.go` | `VectorJoin` case re-anchors **both** arms onto the same `[start,end]` (no widening — each arm's own `RangeWindow`/`RangeLWR` widens its scan), COW-cloning the join. **The blocker**: the slicer zeros both arms, so reanchor must re-fill them; without this, registration silently produces wrong shards. An `@`-pinned arm → `ErrReanchorGridMismatch` → route A. |
| `internal/solver/planner.go` | `sawInstantVectorJoin` signal + `VectorJoin` cases in `walkNode` and `walkScalarInterior` + `Plan()` gate (3b). |
| `internal/solver/decision.go` | `ReasonInstantJoin` telemetry const. |
| `internal/chplan/sliceinvariant.go` | Register `&VectorJoin{}` + boundary proof in the doc comment. |
| tests | registry-pin update (21 default-denied); reanchor both-arms / COW / `@`-pinned / rapid property cases; planner StepAligned discrimination + scalar-interior negative control; chDB A-vs-B parity fixtures (one-to-one and `on() group_left()` ratios). |

**Zero emitter/chsql changes.** `go build` + `go test` green on `internal/chplan` and `internal/solver`; chdb parity lane type-checks under the `chdb` tag (CI runs it).

## Reviewer notes / forward items (non-blocking)

- **Executor fail-hard confirmed**: `group_left`'s per-shard `throwIf(uniqExact>1)` preserves A/B parity only because the shard executor fails the whole request on any single-shard error (`executor.go` errgroup first-error-wins + `cursor.go` `g.Wait()` reap). Verified, no change needed.
- **Forward note (not a live bug)**: `spineOffsetAndD` collapses two arms' offsets into one global scan floor; for a routed join with *asymmetric per-arm offsets* it could under-represent one offset. Currently only feeds `Slice.ScanFrom`, which is dead-stored, so no live defect — worth tightening if that field is ever wired up.
- **Minor test-coverage gap**: the `group_left` parity fixture doesn't trip `throwIf` (seed has one series per `(job, anchor)`), so the many-to-one fail-hard parity path is proven only in the uniqueness-satisfied case. The A-vs-B harness `t.Fatalf`s on query errors, so exercising a firing `throwIf` there would abort the compare — a dedicated error-parity unit test is the right home if we want it covered.

## Not auto-merging

This touches the slice-invariance **safety gate** (a wrong registration silently produces incorrect cross-shard results with no compile/test signal — the `#92` hazard the registry doctrine exists to prevent), so I'm holding it for your review rather than enabling auto-merge.

### DIFF
--- a/internal/chplan/reanchor.go
+++ b/internal/chplan/reanchor.go
@@ -179,6 +179,35 @@ func reanchor(n Node, start, end time.Time) (Node, error) {
 		}
 		// Predicate is off-grid immutable: share.
 		return &Filter{Input: input, Predicate: v.Predicate}, nil
+	case *VectorJoin:
+		// A step-aligned vector-vector join carries NO own anchor grid and NO
+		// lookback: each arm is an independent windowed spine that already
+		// evaluates over the request grid, and the join step-aligns the two on
+		// the per-anchor TimestampColumn. Re-anchor BOTH arms onto the SAME
+		// [start, end] the join was asked for (no widening — the arms' own
+		// RangeWindow / RangeLWR nodes do their -Range / -Lookback widening),
+		// then copy-on-write the join node, re-filling the two arms and sharing
+		// the immutable modifier fields (Op / Match / Card / Include /
+		// ReturnBool / StepAligned) + the four column names verbatim. An
+		// @-pinned or grid-divergent arm surfaces ErrReanchorGridMismatch from
+		// the recursion, aborting the whole re-anchor to route A. The
+		// instant-mode (!StepAligned) join is kept off this path by the
+		// planner's sawInstantVectorJoin fail-closed guard (its emitter
+		// synthesizes the join-side timestamp with now64(9), a wall-clock that
+		// diverges across shards); registration is by node kind, so the guard —
+		// not this case — is what excludes it.
+		left, err := reanchor(v.Left, start, end)
+		if err != nil {
+			return nil, err
+		}
+		right, err := reanchor(v.Right, start, end)
+		if err != nil {
+			return nil, err
+		}
+		c := *v
+		c.Left = left
+		c.Right = right
+		return &c, nil
 	default:
 		// Off the windowed spine: SHARE the immutable subtree verbatim. The
 		// off-spine subtree is byte-identical across all K shards (it does not
--- a/internal/chplan/reanchor_vectorjoin_test.go
+++ b/internal/chplan/reanchor_vectorjoin_test.go
@@ -1,0 +1,194 @@
+package chplan_test
+
+import (
+	"errors"
+	"testing"
+	"time"
+
+	"pgregory.net/rapid"
+
+	"github.com/tsouza/cerberus/internal/chplan"
+)
+
+// stepAlignedJoin builds a StepAligned vector-vector join over two matrix
+// RangeWindow arms — the shape `sum by(job)(rate(a[5m])) / sum by(job)
+// (rate(b[5m]))` lowers to (modulo the per-arm Aggregate, immaterial to
+// re-anchoring). group_left + Include exercise the many-to-one modifier
+// fields ReanchorRange must carry verbatim.
+func stepAlignedJoin(left, right chplan.Node) *chplan.VectorJoin {
+	return &chplan.VectorJoin{
+		Left:             left,
+		Right:            right,
+		Op:               chplan.OpDiv,
+		Match:            chplan.VectorMatch{Labels: []string{"job"}, On: true},
+		Card:             chplan.CardManyToOne,
+		Include:          []string{"instance"},
+		ReturnBool:       false,
+		StepAligned:      true,
+		MetricNameColumn: "MetricName",
+		AttributesColumn: "Attributes",
+		TimestampColumn:  "TimeUnix",
+		ValueColumn:      "Value",
+	}
+}
+
+// TestReanchorRange_VectorJoin_ReAnchorsBothArms asserts a step-aligned join
+// re-anchors BOTH arms onto the SAME requested [start, end] (a join carries no
+// lookback, so no widening at the join level), and the input tree stays
+// byte-identical (copy-not-mutate).
+func TestReanchorRange_VectorJoin_ReAnchorsBothArms(t *testing.T) {
+	t.Parallel()
+
+	left := matrixWindow(5*time.Minute, time.Minute, 0)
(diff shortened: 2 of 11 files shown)


tsouza:
## Test coverage — 13-layer map

Second commit (`fix(solver): derive sharded scan floor as per-arm max reach…`) adds the offset fix + coverage. Full layer assessment for both commits:

| Layer | Applies | Coverage |
|---|---|---|
| 1 Parser smoke | — | no parser change |
| 2a chplan TXTAR snapshots | — | no lowering change; solver routing isn't in the pre-optimizer spec lane |
| 2b Lowering edge cases | — | no lowering change |
| **3 chplan IR invariants** | ✅ | `reanchor_vectorjoin_test.go`: both-arms re-anchor, COW modifier preservation, @-pinned→`ErrReanchorGridMismatch`, **rapid property now draws asymmetric per-arm Range+Offset (100 cases)** |
| 4 Optimizer rule properties | — | no optimizer rule |
| 5 chsql Frag/QueryBuilder goldens | — | **zero emitter changes** |
| 6a PromQL chDB roundtrip | pre-existing | route-A ratio behavior unchanged |
| 6b/6c LogQL/TraceQL | — | PromQL-only |
| 6d Function-surface parity | — | no symbol accept/reject change |
| 6e Strict-scan (real CH) | gap (low-risk) | route-B join SQL reuses route-A emitters (already strict-scan-covered); route B not independently run on strict CH — noted follow-up |
| **A-vs-B chDB differential** (solver parity) | ✅ | `avb_chdb_lane_test.go`: one-to-one, `on() group_left()`, **+ new asymmetric-offset fixture** |
| **Unit — planner decision** | ✅ | `planner_vectorjoin_test.go`: StepAligned discrimination, scalar-interior negative control, @-pinned |
| **Unit — spineReach** | ✅ | `slicer_reach_test.go` (8 cases: per-arm max, nested compounding, LWR, aggregate pass-through, negative offset) |
| **Unit — registry pin** | ✅ | `sliceinvariant_test.go` (21 default-denied) |
| **Unit — executor fail-hard** | ✅ pre-existing | `TestExecute_FirstErrorWins` (every shard idx × error class → typed error, no partial concat) — the group_left A/B parity foundation |
| 7 / 7b HTTP + consumer-corpus | — | no handler/wire/decode change |
| 8 System / lifecycle | — | no env-var/lifecycle change |
| 9 Playwright UX | — | routing is transparent |
| 10 Chaos (deterministic) | pre-existing | shard-error fail-hard = FirstErrorWins matrix |
| 11 Perf / alloc | optional | `slicer_bench_test.go` exists; no required add |
| 12 Fan-out guards | — | route B *reduces* fan-out; no new unbounded shape |
| 13 Live-stack chaos | — | no new fault contract |

### Deliberately not built (disproportionate)
- **Property-based A/B parity** (rapid-generated join queries executed on chDB): the deterministic A/B lane + the rapid reanchor property already cover the space; a new generator+oracle harness isn't warranted for one node family.
- **Route-B on strict real CH (6e):** reuses route-A emitters; heavy infra, low marginal risk.

### Flagged pre-existing follow-up (NOT this PR)
`ReanchorRange`'s `*RangeWindow` case widens its inner spine by `-v.Range` only, while `*RangeLWR` uses `-v.Offset-v.Lookback`. For an offset-carrying **nested subquery** spine this may under-widen the inner grid. Not triggered here (join arms are bare `rate()` over `Scan` leaves), pre-existing, worth verifying separately.

PR #1284, opened 2026-07-26
feat(migration): Tier-2 ruler substrate + the five Tier-2 scenarios, running in CI

tsouza:
## Summary

Implements plan section **3a** (Tier-2 ruler substrate) from
`docs/migration-testing.md` section 3's "Tier 2 — ruler tier" — the
infrastructure step, not any MIG-09/13/18/19/24 scenario yet (that's 3b).

- `test/e2e/migration/tiers/tier2-ruler/docker-compose.ruler.yml` extends
  `tiers/tier1-dual/docker-compose.dual.yml` via the standard `-f a -f b`
  merge (no `extends:` directive) with two new services:
  - **`grafana`** (12.2.9) — the "real query-only external ruler"
    docs/migration-testing.md describes, here Grafana-managed alerting
    rather than a Prometheus/Thanos ruler. Provisioned entirely via mounted
    YAML: a Prometheus datasource against cerberus
    (`grafana/datasources/cerberus.yaml`, mirroring the Layer-9 dashboard
    smoke's provisioning shape) plus unified-alerting-as-code
    (`grafana/alerting/{rules,contact-points,notification-policies}.yaml`).
  - **`dead-end-receiver`** — a ~65-line stdlib-only Go `net/http` server
    (`receiver/main.go`) that logs every notification to a file and answers
    200, but never forwards anywhere. Grafana's one contact point points at
    it.
- The rule fixture is **not invented** — `grafana/alerting/rules.yaml`
  reuses, verbatim, the kube-prometheus-stack archetype's colon-named
  recording rule (`node:node_cpu_utilisation:avg1m`) and its
  `NodeCPUSaturation` alert
  (`test/e2e/migration/archetypes/kube-prometheus-stack/rules/`), the shape
  MIG-09's story wants. Only the Grafana scheduling wrapper (`data`,
  `condition`, `record.metric`, a tuned-down `interval: 10s`) is new.
- `TestTier2Substrate` (`test/e2e/migration/tier2_ruler_test.go`,
  `migration_tier2`-gated) is the substrate self-check, mirroring
  `tier1_stack_test.go`'s prove-the-mechanism-first pattern from #1283:
  before any MIG scenario is written against this substrate, it asserts (1)
  Grafana provisioned both rules without error, (2) the alert rule evaluates
  against cerberus **for real** — health `ok`/`nodata`, never `error` — and
  (3) the dead-end receiver actually captures a notification through
  Grafana's real, provisioned contact point.
- `Justfile` gets `migration-tier2-{up,seed,run,down}` mirroring the
  `migration-tier1-*` recipes exactly, plus a second
  `go vet -tags=migration_tier2` line in `test` alongside the existing
  tier1 one (two separate invocations, not a combined
  `-tags=migration_tier1,migration_tier2` — the two tiers never share a
  process, and I didn't want to require tier2's helper names to keep
  avoiding tier1's forever).

## Why the recording rule's write-back is expected to error

`docs/migration-testing.md` describes Tier 2's recording-rule output being
written back "through the OTLP collector into ClickHouse" — that ingest
path doesn't exist yet (cerberus is query-only; the collector has no
Prometheus remote-write receiver configured). So in this substrate,
Grafana's recording rule genuinely queries cerberus (proving the mechanism)
but its write-back leg errors, and `node:node_cpu_utilisation:avg1m` never
actually has data — which means `NodeCPUSaturation`'s own query legitimately
returns "no data" rather than a query error. The self-check treats
`health == "nodata"` as a pass (cerberus *answered*) and only `"error"`
as a failure. Building that write-back path is out of scope for this PR;
it's the natural next piece for whichever PR builds the MIG-09 scenario.

## Real bugs found and fixed while verifying live

1. **`docker compose -f a -f b` resolves every relative host path in every
   merged file against the FIRST file's directory**, not each file's own —
   confirmed via `docker inspect`'s `Mounts.Source` on a real run, where a
   naïve `./grafana/...` in the tier2 compose file silently bound
   `tier1-dual/grafana/...` (nonexistent) instead of `tier2-ruler/grafana/...`.
   Fixed to `../tier2-ruler/grafana/...`, documented inline so it isn't
   "simplified" back.
2. **Grafana rejects a rule-group `interval` that isn't an exact multiple of
   the scheduler's base interval** (10s, unconfigured here) — `15s` made
   the whole alerting provisioning service fail to start. Fixed to `10s`.
3. **gosec caught real issues** in the receiver on first `git push` (pre-push
   hook mirrors CI's `lint` job): log directory/file modes too permissive
   (G301/G302) and a bare `http.ListenAndServe` with no per-request timeout
   (G114). Fixed in a follow-up commit.

## Verification (real, live stack)

```
$ just migration-tier2-up
...
 Container cerberus-migration-tier1-grafana-1 Healthy
 Container cerberus-migration-tier1-dead-end-receiver-1 Healthy
 (+ all tier1 services healthy)

$ just migration-tier2-run
==> migration tier-2 substrate assertions
go test -tags=migration_tier2 -count=1 ./test/e2e/migration/...
ok    github.com/tsouza/cerberus/test/e2e/migration    0.088s
...

$ go test -tags=migration_tier2 -v -run TestTier2Substrate ./test/e2e/migration/
=== RUN   TestTier2Substrate
=== RUN   TestTier2Substrate/cerberus_ready
=== RUN   TestTier2Substrate/grafana_provisioned_the_rule_fixture
=== RUN   TestTier2Substrate/grafana_evaluates_the_alert_against_cerberus
=== RUN   TestTier2Substrate/dead_end_receiver_captures_a_notification
--- PASS: TestTier2Substrate (0.04s)
    --- PASS: TestTier2Substrate/cerberus_ready (0.01s)
    --- PASS: TestTier2Substrate/grafana_provisioned_the_rule_fixture (0.01s)
    --- PASS: TestTier2Substrate/grafana_evaluates_the_alert_against_cerberus (0.01s)
    --- PASS: TestTier2Substrate/dead_end_receiver_captures_a_notification (0.01s)
PASS
```

`docker cp`'d the receiver's own `/var/log/dead-end-receiver/notifications.log`
out of the running container afterward — it shows both an **organic**
`DatasourceNoData` alert that fired for real through Grafana's alerting
engine (NodeCPUSaturation's query legitimately returning no data, as
expected — see above) and the test-notification calls the self-check's
third sub-test issues, all routed correctly to the one `dead-end` contact
point.

`just migration-tier2-down` torn down cleanly afterward.

## Scope note

This is 3a only (substrate + self-check). 3b (the 5 Tier-2 Gherkin
scenarios: MIG-09/13/18/19/24) and 3c (CI wiring: the `migration-tier2` job,
`migration-e2e.mjs` tier resolution, aggregate fold, coverage-baseline
bump) are separate follow-up PRs per the plan's own phased sequencing.

## Test plan

- [x] `just migration-tier2-up` — full stack (tier1 + tier2) comes up
      healthy
- [x] `just migration-tier2-run` — `TestTier2Substrate` passes, all four
      sub-tests
- [x] Verified the dead-end receiver's log file (via `docker cp`) actually
      contains captured notification bodies, not just an in-memory counter
- [x] `just migration-tier2-down` — clean teardown
- [x] `go vet -tags=migration_tier1` and `-tags=migration_tier2` both clean
- [x] `gofumpt -extra -l` clean on new Go files
- [x] `golangci-lint run` clean on the receiver package (after the gosec
      fixes)
- [ ] CI (this PR carries no CI wiring yet — `migration-tier2` job lands in
      3c; `check`/`lint`/`compatibility/*`/etc. should still pass since
      nothing here changes untagged code paths)


tsouza:
## Rebased onto merged `main`, and tier-2 now actually runs

This branch was stale and red. It is now `main` + the tier-2 work, with the
tier-1 fixes from #1285 that tier-2 depends on (multi-archetype seeding by
identity rather than an impossible window offset, the drift marker, the
per-tier selection gates).

**`migration-tier2` executed on CI for the first time** — [run 30208938283](https://github.com/tsouza/cerberus/actions/runs/30208938283):

| tier | result |
| --- | --- |
| tier-0 (7 scenarios) | pass |
| tier-1 (18 scenarios) | pass |
| tier-2 (5 scenarios) | **3 pass, 2 fail** |

The substrate works end to end: Grafana evaluates against cerberus, the relay
lands recorded series in ClickHouse, the dead-end receiver captures
notifications.

### The two failures are real bugs, and the assertions that caught them bite

- **MIG-18** — Grafana rendered `{{ $labels.probe }}` as `[no value]`. The
  assertion is correct; the rule fixture is wrong. A `max()`-style reduce
  yields a result with no labels for the rule's configured labels to merge
  onto. Being fixed in the fixture, **not** by weakening the assertion.
- **MIG-19** — landed `0.05` vs live re-evaluation `-27.83`. A negative
  CPU-utilisation rate is nonsense, so the re-evaluation is what is broken:
  MIG-09 and MIG-13/19 seed the same ClickHouse series identity with different
  counter rates over overlapping windows, so `rate()` over the interleaved,
  non-monotonic result produces a counter-reset artifact. Fix is distinct
  identities per scenario — **not** a widened tolerance, since exact parity is
  the correct comparison mode (docs section 5, mode 1).

### Known caveat, not hidden

The ratchet reports `scenarios 30/30 across 26/26 stories; 0 violations`, but
that counts ENUMERATED scenarios, not executed ones, and MIG-18's PASS
assertion in section 6 was narrowed toward the single-ruler substrate. Both
holes are being closed by a follow-up that makes the ratchet attest execution
and pins each PASS cell's text.

PR #1290, opened 2026-07-26
fix(ci): make migration-tier0 a plain job; teach every seed the ServiceName column

tsouza:
Follow-up to #1289, which merged before these landed. `main` is currently red on the chdb lane and still renders the garbage tier-0 job name; this fixes both.

## 1. migration-tier0 is not a matrix

Its job name renders as:

```
migration-tier0 (tier0, tier0, MIG-01, MIG-03, MIG-04, MIG-05, MIG-10, MIG-14, MIG-26, 15)
```

GitHub appends the entire matrix payload when a matrix job's `name:` doesn't reference it. Two of those four fields were waste — `name` was byte-identical to `tier`, and `stories` was read by **nothing** in the workflow — and the matrix could never hold more than one entry anyway, because `tier0` was the only tier flagged `matrixDriven`. It was a fan-out over exactly one element, and the unreadable name was the price.

It is now a fixed job like its two siblings: `TIER: tier0`, a literal `timeout-minutes: 15` mirroring `TIER0_TIMEOUT_MIN`, and a concurrency group keyed on the ref alone. `buildMatrix`, `matrixDriven` and the `matrix` job output are deleted.

That also retires the scar tissue those needed. Three hazards were previously held off by comments and a guard test — an empty `include:` doesn't skip a matrix job; `${{ matrix.name }}` renders literally against a null matrix context; a `tier1` entry reaching the matrix would spawn a backend-less shard named exactly like the real fixed job. All three are now structurally impossible rather than guarded. `migration-e2e.test.mjs`'s matrix tests become tier-selection tests: 55/55 green.

## 2. Seeds the backfill never reached

`backfillMetricsColumns` only matched a bare `CREATE TABLE `, so seeds written `CREATE OR REPLACE TABLE` slipped past it entirely and failed at query time with `UNKNOWN_IDENTIFIER: ServiceName`. That is what broke `integration (promql)` and `perf-guards` on #1289's own run. Both mirrored copies now match through a shared `createTableHeads` table covering `CREATE OR REPLACE TABLE` and `CREATE TABLE IF NOT EXISTS`, longest-first so the bare form can't shadow them.

Seeds that build DDL directly rather than through that seam are fixed at the source:

- **`internal/chsql`** — eight chDB test files each carried a byte-identical inline `CREATE`. They now share one `metricsSeedDDL(table)` renderer. A column added to the read path has to reach every seed at once, and eight copies is precisely how that drifts; this was the second time in one change that all eight needed the same edit.
- **`test/perf/native_scan_projection`** — declares **and populates** `ServiceName` with the same value its resource map's `service.name` carries, exactly as the OTel-CH exporter writes both. This test measures real per-column bytes out of `system.parts_columns`, so a DEFAULT-only column would have no entry to measure and the guard would fail on its own bookkeeping rather than on a regression.
- **`test/perf/scale_wall_pin`**, **`internal/promql/limit_ratio`** — DEFAULT-only; the read path only needs the column to resolve.

## 3. Goldens

The eight MIG-05 archetype explain goldens carry the overlay now. The churn is exactly `mapUpdate(…)` → `mapConcat(mapUpdate(…), mapFilter(v != '', map(…)))` plus `ServiceName` joining each scan's column list — 37 insertions, 37 deletions, nothing else moved.

## Test evidence

- `go test -tags chdb -count=1 ./test/... ./internal/...` — clean, full sweep, zero failures.
- `go test ./...` — clean.
- `node --test .github/scripts/migration-e2e.test.mjs` — 55/55.
- Tier-0 suite green with the regenerated goldens.
- migration-e2e dispatched against this branch to verify the lane end to end.


tsouza:
Superseded — #1289 squash-merged, so this branch's pre-squash history conflicts with `main`. Reopened from a clean rebase onto `main` with identical content.

PR #1306, opened 2026-07-27
fix(release): make the Homebrew publish path actually work (Formula/ dir + a runner with brew)

tsouza:
## What broke

`brew-smoke` failed on the v1.12.0 release run ([job 89897832157](https://github.com/tsouza/cerberus/actions/runs/30238455121/job/89897832157)):

```
GET https://api.github.com/repos/tsouza/homebrew-tap/contents/Formula/cerberus.rb -> 404 Not Found.
tsouza/homebrew-tap has no readable Formula/cerberus.rb: `brew install tsouza/tap/cerberus`
is broken for every operator, whatever this release published.
```

goreleaser *did* push the formula — the tap carries commit `b3b0a50b` *"Brew formula update for cerberus version v1.12.0"*, and its four `sha256`s match the published `checksums.txt` byte for byte. But it landed at the tap **root** as `cerberus.rb`, because `brews[].directory` defaults to empty rather than to `Formula`. From the goreleaser log:

```
• homebrew formula
  • pushing   repository=tsouza/homebrew-tap branch= file=cerberus.rb
```

Homebrew resolves a tap's formulae from `Formula/`, `HomebrewFormula/`, then the tap root. The root is the least conventional of the three and the one Homebrew's tap-layout standardisation is squeezing.

**This was the first release ever to reach `brew-smoke`** — the 1.10.x / 1.11.x lines carry an older `release.yml` with no `publish` / `brew-smoke` jobs — so the gate caught the defect on its first execution. That is precisely what it was built for.

## Fix

`directory: Formula` on the `brews` entry. `directory` is the correct v2 schema key, confirmed against goreleaser's published JSON schema (`Homebrew.properties`) rather than from memory; the release job runs `~> v2` and resolved to v2.17.1.

### Already applied out-of-band

The v1.12.0 formula was relocated in the tap ([`7c13688`](https://github.com/tsouza/homebrew-tap/commit/7c13688bda7c46934a7fb87cccc6291f982e8563)) so the shipped release is installable *now* rather than only from 1.12.1 onward. Content is unchanged — a plain `git mv`, checksums re-verified against the release before and after. `brew-smoke` has been re-run on the release run to confirm.

## Also: `.goreleaser.yml` had no validation lane

A malformed or renamed key could only ever be discovered by the `goreleaser` job **mid-release**, after the version-bump PR had merged and the publish was underway. Added a `goreleaser-check` CI job running `goreleaser check` against the same `~> v2` binary.

Scope honesty: `goreleaser check` validates **schema, not semantics** — it would *not* have caught this bug, since `directory:`-absent is perfectly valid config. `brew-smoke` remains the gate for what actually gets published. This closes a different, real hole.

## Note

`goreleaser-check` is not in branch protection's required-check set, so it reports but does not block. Say the word and I'll add it.

### DIFF
--- a/.github/workflows/brew-verify.yml
+++ b/.github/workflows/brew-verify.yml
@@ -1,0 +1,85 @@
+name: brew-verify
+
+# ON-DEMAND + weekly re-verification that `brew install tsouza/tap/cerberus`
+# still works for an ALREADY-PUBLISHED release.
+#
+# Why this exists, separately from release.yml's `brew-smoke`:
+#   * `brew-smoke` fires exactly once, inside the release run, and only when
+#     that run publishes a new version. When it goes red there is no way to
+#     re-verify a fix: `rerun-failed-jobs` replays the workflow as it existed
+#     at the release commit, so a correction to the job (a different runner, a
+#     different tap path) can never be exercised against the release that
+#     needed it. v1.12.0 hit exactly that wall.
+#   * The tap is a separate repository. A formula can rot after publish —
+#     someone edits the tap, an asset is deleted, a checksum stops matching —
+#     and nothing in this repo would notice until the NEXT release ran its
+#     smoke.
+#
+# It reuses `brew-smoke.mjs` unchanged, so there is one implementation of what
+# "the published formula is healthy" means; this lane only differs in when it
+# runs and which version it targets.
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: >
+          BARE app version to verify, e.g. `1.12.0` (no leading `v`). Defaults
+          to the repository's latest published non-prerelease release.
+        required: false
+        type: string
+  schedule:
+    # Mondays 07:00 UTC, an hour after link-check-external so the two weekly
+    # rot detectors do not contend for runners.
+    - cron: '0 7 * * 1'
+
+permissions:
+  contents: read
+
+concurrency:
+  group: brew-verify-${{ inputs.version || 'latest' }}
+  cancel-in-progress: false
+
+jobs:
+  brew-verify:
+    name: brew-verify
+    # Homebrew is preinstalled here and absent from the Ubuntu 24.04 image.
+    # Same reasoning as release.yml's brew-smoke — see the comment there.
+    runs-on: macos-latest
+    steps:
+      # Resolve the target version BEFORE checkout, because the checkout has to
+      # land on that version's tag: the `config-docs -check` payload compares
+      # the installed binary's config registry against the working tree's
+      # docs/configuration.md, so checking out `main` would compare a released
+      # binary against unreleased docs and red on drift that is not a defect.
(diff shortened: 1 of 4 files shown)


tsouza:
## Second defect, same gate

Re-ran `brew-smoke` on the v1.12.0 run after relocating the tap formula. It got **past** the `Formula/cerberus.rb` 404 — that part is fixed — and died 8 seconds in on the next thing:

```
brew-smoke: `brew --prefix` did not resolve — Homebrew is not available on this runner.
exit=null error=spawnSync brew ENOENT
```

The job was pinned to `ubuntu-latest`, and the **Ubuntu 24.04 runner image ships no Homebrew**. There was no install step either, so `brew` was a bare ENOENT — **this job could never have passed on that runner**, whatever the tap contained.

Moved to `macos-latest`: Homebrew is preinstalled (no install step), and macOS is where the overwhelming majority of `brew install` traffic originates. Everything the smoke runs after the install (`--version`, `migrate schema`, `config-docs -check`) is pure Go and platform-independent, so the darwin build exercises the same payloads.

So the first-ever execution of this gate found **two** independent defects that between them made `brew install tsouza/tap/cerberus` unverifiable:

1. the formula published to the tap root instead of `Formula/` (previous commit + the out-of-band tap relocation), and
2. the smoke running on an image with no `brew` (this commit).

Both are now fixed. The gate itself was sound — it was never able to run before, because 1.10.x / 1.11.x carry an older `release.yml` with no `publish` / `brew-smoke` jobs.

tsouza:
## Dropped the `goreleaser-check` job from this PR

It failed, on my own addition. `goreleaser check` does not merely validate schema — it exits non-zero on deprecated properties:

```
• checking                    path=.goreleaser.yml
• dockers and docker_manifests are being phased out ... replaced by dockers_v2
• DEPRECATED: brews should not be used anymore
• .goreleaser.yml             error=configuration is valid, but uses deprecated properties
```

Both deprecations are pre-existing. The job therefore cannot go green until `brews` → `homebrew_casks` and `dockers` → `dockers_v2` are migrated — and the `brews` migration changes **how the formula is published**, which is precisely the mechanism this PR is stabilising after it broke v1.12.0's `brew install`.

Forcing that migration mid-flight is the wrong order, and making the check pass by tolerating the deprecation is not an option. So `goreleaser-check` lands together with the migration, tracked as follow-up work; note the migration likely moves `Formula/` → `Casks/` and turns `brew install --formula` into `--cask` in `brew-smoke.mjs`, which is exactly why it wants its own PR and its own verification pass.

This PR keeps the three changes that make the current mechanism correct:

1. `directory: Formula` — the v1.12.0 defect.
2. `brew-smoke` on `macos-latest` — the runner had no `brew` at all.
3. the `brew-verify` dispatch lane — so a fix to the smoke can actually be exercised against an already-published release, which is impossible via `rerun-failed-jobs`.

PR #1381, opened 2026-07-29
chore(release): retire release/1.11.x ahead of the support window

tsouza:
## Why 1.11 goes now

The latest-3-minor window still counts 1.11 as supported — v1.13 is current, so
it sits one inside the boundary. It is retired anyway, because the window says
how many lines cerberus is *willing* to support and cannot say whether a
particular line is still *supportable*.

`release/1.11.x` predates both the Map-canonicalisation family of wrong-answer
corrections and the regression suite those fixes are pinned by. Backporting them
is a rewrite rather than a cherry-pick — a trial backport conflicted on
`canonical_series_keys_test.go`, which has no base on that line at all. Six
correctness fixes 1.12.x and `main` carry are absent from it (#1358, #1356,
#1362, #1340, #1347, #1374). Continuing to advertise support cerberus cannot
deliver is the worse option.

Nothing unpublished is destroyed: the branch tip is
`e000d4ae3b4d84ecb39b3a137ec31fb7ff53d681`, which is exactly `v1.11.4^{}`. The
`v1.11.*` tags and Releases are retained, like every other EOL line.

## The window becomes an upper bound

`RETIRED_LINES` maps a branch to the reason it went early, and
`supportWindowProblem` consults it **before** the version arithmetic. That
ordering is the point: the refusal to publish needs no resolvable tag set, and
it survives the branch being re-created and pushed to.

## Deleting the branch is a separate, deliberate step

`eol-retire.yml`, dispatched by hand, `dry_run` defaulting to true. It is
**fail-closed**, unlike its post-publish sibling `eol-retire-line`: that one runs
after a release is already public, so a token refusal there must not red a
successful publish. This one has nothing riding on it, so a branch it could not
delete is a failed run rather than a silent no-op that would leave the line
standing while the docs said it was gone.

Each deletion is first proved lossless by `unpublishedWorkProblem`: the branch
tip must be the commit of a published **stable** tag **on that line**, which is
what makes the surviving tags a complete reconstruction of the branch. A tip
that matches some other line's tag does not count — that coincidence says
nothing about this line's history and usually means the branch and its name have
diverged. A branch carrying an unreleased backport, a revert, or a hand-pushed
commit stops the run and names the highest tag to diff against.

## The silent failure mode the tests close

`supportWindowProblem` returns `null` for anything that is not a maintenance
branch, and it does so *before* consulting the declaration. So a typo'd
`RETIRED_LINES` key — `releases/1.11.x`, `release/1.11`, a stray space — is not
a loud failure but a silent one: the entry sits in the map looking authoritative
while the line it names stays publishable.

`every declared retirement is a branch the window math can see` asserts every
entry actually blocks, and that each carries a non-empty reason. Nine more cases
cover the declaration outranking the arithmetic, surviving an unresolvable tag
set, and every refusal branch of `unpublishedWorkProblem` — including that an
unresolvable tip and a non-maintenance branch are refusals rather than passes,
since "nothing to check" and "checked and fine" read identically in a driver
that deletes branches.

21/21 pass in `release-preflight.test.mjs` (required `lint` lane).

## Docs

`docs/operations.md` gains an **Early retirement** subsection under the EOL
policy, and the window is restated as an upper bound rather than a schedule.
`.github/scripts/README.md` gains the `eol-retire-declared` env + exit contract
and the new exported helper.

### DIFF
--- a/.github/scripts/README.md
+++ b/.github/scripts/README.md
@@ -335,9 +335,22 @@ One implementation means a new rule guards BOTH lanes at once.
   (`X.Y.0`, `Y>0`) — patches / major bumps / backports / prereleases retire
   nothing — deletes only a provably out-of-window branch that exists (idempotent
   on a 404) with a `supportWindowProblem` cross-check, and ANY deletion failure
-  logs `::error::` and still exits 0 (the release already published). Pure
-  exported `MAINTENANCE_BRANCH_RE` + `currentMinor(tags)` +
+  logs `::error::` and still exits 0 (the release already published). A line can
+  also be retired AHEAD of the window slide, when it diverges far enough that
+  the fixes the supported lines carry no longer apply to it: `RETIRED_LINES`
+  maps such a branch to the reason it went early, `supportWindowProblem`
+  consults that map BEFORE the version arithmetic (so the refusal to publish
+  needs no resolvable tag set and survives the branch being re-created), and the
+  `eol-retire-declared` command — the `eol-retire.yml` manual dispatch, with a
+  `DRY_RUN` mode — deletes the declared branches. That driver is FAIL-CLOSED,
+  unlike its post-publish sibling: nothing rides on the dispatch, so a branch it
+  could not delete is a failed run rather than a silent no-op. Each deletion is
+  first proved lossless by `unpublishedWorkProblem`, which requires the branch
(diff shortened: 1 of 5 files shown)


tsouza:
Not needed. release/1.11.x is retired — the branch is deleted.

PR #1403, opened 2026-07-31
fix(solver): extract features from every GridCarrier kind, not just RangeWindow

tsouza:
## The defect

`internal/solver/planner.go`'s `recordRangeWindow` was the only producer of the
corpus features `n_anchors` / `outer_range` / `cumulative_d`, and its parameter
type admitted exactly **one** of the seven `chplan.GridCarrier` implementations.
`walkNode` had no arm for `RangeWindowNative` or `RangeWindowResample` at all.

So a plan whose only grid carrier is a natively-lowered `rate()` extracted
nothing: it collapsed to `decision_reason='not-sliceable'` with an all-zero
feature vector, indistinguishable in the calibration corpus from a genuinely
trivial query. On a reference deployment, 2,524 of 2,650 `not-sliceable`
verdicts over three days carried `outer_range=0`.

Two independent blindnesses, same root cause — a vocabulary keyed on one
concrete node type instead of on the `GridCarrier` interface:

| Site | Saw | Blind to |
|---|---|---|
| `solver` feature extraction | `RangeWindow`, `RangeLWR` | `RangeWindowNative`, `RangeWindowResample`, `RangeBucketFanout`, `StepGrid`, `AbsentOverTime` |
| `engine` `shapeModifiers` | `RangeWindow`, `RangeWindowNative`, `RangeWindowResample` | `RangeLWR`, `RangeBucketFanout`, `StepGrid`, `AbsentOverTime` |

## The fix

**1. Extraction keyed on the interface.** `recordRangeWindow` →
`recordGridCarrier(chplan.GridCarrier, …)`, backed by `carrierGeometryOf`, which
maps each of the seven kinds to its `(outerRange, lookback, reanchorable)`
geometry. `walkNode`'s **default** arm now records any `GridCarrier` that has no
dedicated arm, so extraction follows the interface rather than the arms of a type
switch. A carrier kind added to `chplan` ahead of the geometry table hits
`carrierGeometryOf`'s `default`, which sets `sawNonRangeWindowSpine` — cost
unknown, not zero.

The RangeWindow / RangeLWR grid checks (unpinned bounds, grid mismatch, inner
resolutions) split out into `checkRangeWindowGrid` / `checkRangeLWRGrid`
unchanged, so those two kinds behave byte-for-byte as before.

**2. `walkNode` arms** for `RangeWindowNative` (recurses widened by `-Range`,
sweeps `GroupBy` + `Recollapse` exprs) and `RangeWindowResample`.

**3. shape_id vocabulary**: `lwr`, `rbf`, `aot`, `sg`, plus the matching
`nodeKind` entries. Previously an expression containing a range fan-out, a union
and a join hashed to `cerb:project;agg=3` — the same id as a bare projection, so
`log_comment` clustering hid the expensive shape inside the cheap bucket.

**4. `decision_reason: extraction-failed`** — emitted when the walk found no
measurable carrier. It is the one refusal whose cost grid is legitimately all
zeros, and it now says so in the row, so a threshold fitted over the refused
population can exclude rows that were never measured instead of averaging them
in. `decision_reason` is `LowCardinality(String)`, **not** `Enum8`
(`internal/optcorpus/chtable.go`), so no DDL migration is needed — only the
`solver.Reasons` list and the `routerrules` mirror, which
`decision_reason_gate_test.go` pins in lockstep.

## Correctness: seeing a carrier never makes it sliceable

Widening extraction changes what is *measured*, never what is *routed*. Every
newly-measured carrier is still refused, by one of two independent gates that
both fire before the cost thresholds:

- `RangeWindowNative`, `RangeWindowResample`, `AbsentOverTime` are absent from
  `chplan`'s slice-invariant registry → `allSliceInvariant=false` → gate (1).
- `RangeBucketFanout`, `StepGrid` are registered slice-invariant but
  `ReanchorRange` clones their grid verbatim, so `recordGridCarrier` sets
  `sawNonRangeWindowSpine` for any non-reanchorable carrier with `Step > 0` →
  gate (1b).

`sawInstantWindow` is gated on `reanchorable` so an instant-shaped broadcast
subtree cannot newly refuse a plan that routes today, and `hasWindow` still
requires `Step > 0`, which no non-reanchorable carrier can reach without having
already tripped gate (1b). The new `extraction-failed` gate sits between the
`instant` guard and gate (1) and is route-neutral: both paths refuse, so it
changes only the recorded reason.

## Tests

- `TestCarrierGeometry_ExtractsFeaturesForEveryKind` — table over **all seven**
  `GridCarrier` kinds with a distinct lookback per kind, asserting the exact
  `NAnchors` / `OuterRange` / `CumulativeD` / `Fanout` / `Step`, and asserting
  that every non-reanchorable kind is still refused at both `Plan` and the
  threshold-free `Eligible` seam.
- `TestCarrierGeometry_CoversEveryCarrier` /
  `TestShapeModifiers_CoverEveryGridCarrier` — parse `internal/chplan` for the
  `(Start, End, Step)` grid signature (the same key chplan's own completeness
  ratchet uses; its registry is unexported) and fail if a kind exists with no
  table row. A carrier added upstream fails here instead of shipping an
  invisible shape.
- `TestCarrierGeometry_NativeIsMeasuredButNeverSliceable` — the correctness pin:
  `sum(rate(...))` lowered natively extracts non-zero features **and** is refused
  by `Plan`, by `Eligible`, and is not registered slice-invariant.
- `TestCarrierGeometryOf_UnclassifiableCarrierFailsClosed` — the default arm
  contributes no features and sets the fail-close flag.
- `TestClassify_NoCarrierRecordsExtractionFailed` — a carrier-free plan on a
  range request reports `extraction-failed` with an empty grid.
- `TestShapeModifiers_EveryGridCarrierGetsAToken` /
  `TestShapeModifiers_DistinguishesFanoutFromBareProject` — every carrier
  contributes its token; fan-out + union + join no longer collides with a bare
  aggregate.

## Golden churn

`test/spec/**` `-- chplan --` / `-- sql --`: **zero** — no lowering, optimizer or
emitter behaviour changes.

`test/perf/solver-decision-baseline.json`: **one row**,
`matrix_selector_top_level` (`up[5m]`), `below-threshold` → `extraction-failed`.
`routed=false, K=0` are unchanged, so this is not a routing move. The fixture's
plan is `Project → Filter → Project → Filter → Scan` with no grid carrier
anywhere: a raw matrix selector returns raw samples and owns no anchor grid, so
its zero cost grid was always absence of measurement rather than a cheap plan,
and the row now says which.

### DIFF
--- a/docs/operations.md
+++ b/docs/operations.md
@@ -201,9 +201,11 @@ on any diff vs reference Prometheus.
 additive `X-Cerberus-Route-Decision` header reporting the per-request
 classification regardless of mode: `routed` (took route B),
 `below-threshold`, `instant`, `instant-join`, `not-sliceable`, `high-D`,
-`now64`, `grid-mismatch`, `incommensurate`, `scalar-heavy`, or
+`now64`, `grid-mismatch`, `incommensurate`, `scalar-heavy`,
 `routing-disabled` (the reason recorded under `single`, where no threshold is
-consulted at all). The header is **omitted**
+consulted at all), or `extraction-failed` (no grid carrier was measurable, so
+the reported cost grid is zero for want of data rather than because the plan is
+cheap). The header is **omitted**
 for non-PromQL heads and when the solver is fully off (nil). It is purely
 diagnostic — observe it to see what the solver would do (under `single`) or
 did (under `auto`) without changing the wire body.
--- a/docs/router-rules.md
+++ b/docs/router-rules.md
@@ -259,7 +259,7 @@ scalar) — context, never a gate, so no inline tolerance number is needed.
 `decision_reason` is the shadow-header value the solver records to explain each
 routing decision (`routed`, `below-threshold`, `not-sliceable`, `instant`,
 `instant-join`, `high-D`, `now64`, `grid-mismatch`, `incommensurate`,
-`scalar-heavy`, `routing-disabled`; see
+`scalar-heavy`, `routing-disabled`, `extraction-failed`; see
 [`internal/solver/decision.go`](../internal/solver/decision.go)). Its primary
 use is **grouping / attribution**: the failure rules group by it so the operator
 can see *which solver path* produced the failure and pick the right lever (shard
--- a/docs/solver.md
+++ b/docs/solver.md
@@ -40,10 +40,17 @@ The signals, each gathered in the one pass:
    marker is proven.
 2. **Routable spine family.** Re-anchoring rewrites the grid carried by the
    `RangeWindow` matrix family and the `RangeLWR` bare-selector
-   last-with-respect-to family. A `RangeBucketFanout` or `StepGrid` spine
-   carries its own eval grid that re-anchoring clones verbatim, so a plan whose
-   spine bound-carrier is one of those fails closed to route A (every shard
-   would otherwise emit stale bounds).
+   last-with-respect-to family. Every other grid carrier —
+   `RangeWindowNative`, `RangeWindowResample`, `RangeBucketFanout`, `StepGrid`,
+   `AbsentOverTime` — carries its own eval grid that re-anchoring clones
+   verbatim, so a plan whose spine bound-carrier is one of those fails closed to
+   route A (every shard would otherwise emit stale bounds).
+
+   Measurement is separate from admissibility: the signal walk records the cost
+   grid of **every** carrier kind, routable or not, because the corpus needs to
+   know how expensive the refused plans were. Being seen never makes a carrier
+   sliceable — that is decided solely by the marker in signal 1 and by whether
+   re-anchoring knows the node's grid.
 3. **Pinned bounds.** Both `Start` and `End` must be pinned (non-zero) on the
    outermost windowed node — it anchors the whole grid. An inner subquery node
    may be unpinned (both bounds zero, the shape the re-anchor fills); a
@@ -94,8 +101,8 @@ consulted when none was.
 Every classification — routed or not — produces a `Decision` carrying the
 reason (`routed`, `below-threshold`, `not-sliceable`, `instant`, `instant-join`,
 `high-D`, `now64`, `grid-mismatch`, `incommensurate`, `scalar-heavy`,
-`routing-disabled`) for the shadow header, alongside the plan's cost grid
-(`N`, `F`, `D`, `OuterRange`, `Step`).
+`routing-disabled`, `extraction-failed`) for the shadow header, alongside the
+plan's cost grid (`N`, `F`, `D`, `OuterRange`, `Step`).
 
 The grid is populated on **every** Decision, including the refusals. The signal
 walk that derives it makes no routing decision and mutates nothing, so it runs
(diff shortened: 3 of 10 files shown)


tsouza:
Adversarial review of this PR's **own tests** finished after it merged and confirmed two test-honesty gaps. Both are now closed on branch `test/solver-carrier-honesty-gaps` (`a5f41fc`), pushed and awaiting a follow-up PR. **No production behaviour changed** — the mechanisms shipped here are correct; only their coverage was hollow.

**1. `internal/engine/plan_shape_id.go` — the four new `nodeKind` arms had no test.** Every carrier fixture wraps its node in a `Project` (deliberately, so the carrier reads as a *modifier*), so `nodeKind` was only ever invoked with `*chplan.Project`. Deleting all four arms (`lwr` / `rbf` / `aot` / `sg`) left the whole suite green. A carrier **is** the plan root in real lowerings — `absent_over_time(...)` roots on `AbsentOverTime` in six txtar fixtures, a negative-offset selector on `RangeLWR` — so a missing arm degrades those ids to the generic `cerb:node` fallback. Added a test calling `nodeKind` on each **unwrapped** carrier plus the end-to-end `planShapeID`, and one pinning the fallback so the per-kind assertions provably discriminate against it.

**2. `internal/solver/carrier_geometry_test.go` — the fail-close was only proven at depth 0.** All seven rows of the correctness half place the carrier at depth 0 as the plan's sole carrier, the one position where the `reanchorable` fail-close is redundant with the outer-grid path. Narrowing `recordGridCarrier` to `if !geom.reanchorable && step > 0 && depth == 0` passed the entire table. Added two placements per non-re-anchorable kind — nested one level under a routable `RangeWindow` spine, and beside one under a `StepAligned` `VectorJoin` — asserting refusal at the threshold-free `Eligible` seam as well as `Plan`, and asserting the cumulative D so a fixture that never reached the carrier cannot pass vacuously. A second test pins that for `RangeBucketFanout` / `StepGrid` (which chplan **does** register slice-invariant) gate (1) is silent, making the depth-free fail-close the entire barrier.

Both mutations were reproduced against a copy of the tree and are now killed: the `nodeKind` deletion fails 4 subtests; the `depth == 0` narrowing lets a nested `RangeBucketFanout` route B at K=8 (caught), with nested `StepGrid` caught at the `Eligible` seam.

tsouza:
Adversarial review of this PR found two test-honesty gaps; both are now fixed in `46c1d0f6` (tests only, **no production change**):

1. **`internal/engine/plan_shape_id_carrier_test.go`** (new) — the `nodeKind` carrier arms were unwrapped: deleting all four left the suite green. The new table pins each arm.
2. **`internal/solver/carrier_geometry_test.go`** — narrowing `recordGridCarrier` to `&& depth == 0` passed the whole existing table, i.e. nothing asserted that carriers nested below the root are recorded. Now covered.

Both mutations were reproduced against the pre-fix tests and confirmed killed by the new ones.

tsouza:
Correction to my previous comment: this PR had already merged when `46c1d0f6` was pushed, so that commit never reached `main`. The two test gaps now ship as #1406 (same commit, rebased onto `main`).

PR #1407, opened 2026-07-31
test(compat): seed the 3 unseeded PromQL metrics and gate corpus-vs-seed drift

tsouza:
## What

The PromQL compatibility corpus referenced three metric families the seeder never wrote:

- `demo_api_request_duration_seconds_bucket`
- `demo_batch_last_success_timestamp_seconds`
- `demo_intermittent_metric`

Both backends returned empty for every query over them, the upstream tester scored "no diff" as PASS, and the corpus header block called that "acceptable noise for the compatibility baseline". It was a hollow green: those cases counted toward the score while testing nothing at all.

This PR seeds all three on **both** backends, deletes the header block that tolerated the hole, and adds a `test/regression/` gate so the hole cannot re-open.

## The three families

**`demo_api_request_duration_seconds`** (`otel_metrics_histogram`) — classic histogram, 2 instances x 2 methods. `ExplicitBounds = [0.1, 0.5, 1, 2.5, 5, 10]`, per-step increments `[5, 20, 60, 45, 40, 29, 1]` (cumulative `[5, 25, 85, 130, 170, 199, 200]`). Every quantile the corpus asks for — 0.1/0.5/0.75/0.9/0.95/0.99, i.e. ranks 20/100/150/180/190/198 — lands **strictly inside** a finite bucket, so none degenerates onto an exact bound and hides an interpolation bug. `BucketCounts` is cumulative over time, so `_bucket` behaves as a counter under `rate()`. The per-series scale factor `mult` in {1,2,3,4} keeps quantiles (a shape property) identical across series while `rate()` values differ, so a label mix-up still shows.

**`demo_batch_last_success_timestamp_seconds`** (`otel_metrics_gauge`) — 3 instances; the gauge **value is a unix timestamp**, which is the shape `time() - max(...)` and the seven date functions are written against. Batch interval 20m, so 3 runs land in the 1h window (`changes(...[1h])` sees 2 transitions) and `time() - max(...)` sweeps 0..1200s, crossing the corpus's `< 1000` threshold in both directions instead of sitting on one side. Instances are staggered 60s so the trailing two fall on the previous day — `hour` / `day_of_month` / `day_of_week` are then non-constant across series, which is the only thing that makes the 21 date-function cases discriminate anything. Derived from the same anchor string as `TimeUnix`, parsed in the same server timezone, so sample time and value can't diverge.

**`demo_intermittent_metric`** (`otel_metrics_gauge`) — 3 instances on a 45s cadence (inside the 5m lookback, so it exercises repeat-last-sample) with one contiguous 12m blackout (longer than the lookback, so both backends must report a genuine staleness gap). Samples sit at 45s multiples while the tester's grid is 10s-aligned, so grid points *exactly* 300s after a sample do occur — the fixture hits Prometheus's exclusive lookback boundary, not only its interior.

## The remote_write mirror

The reference Prometheus has no scrape config; the seeder mirrors the CH fixture into it over remote_write. Two additions:

1. **The two new gauges** join `fixtureSources`. This closed a **live drift**: the list had already fallen behind `fixtureInserts` — exactly what the new gate now prevents.
2. **A histogram arm** (`histogramFixtureSources` + `readHistogramFixtureSeries`) that fans one CH row out into `_bucket` / `_count` / `_sum`.

Two deliberate choices in that arm:

- **Wire names come from production, not a private suffix list.** `histogramWireNames` enumerates with `promql.HistogramSyntheticNames(base, schema.Metrics)` and classifies each result with `schema.Metrics.HistogramCompanionColumn`. If a lowering change adds a fourth companion or renames one, the mirror fails loudly instead of silently under-mirroring. The bare base name is never mirrored — cerberus's lowering doesn't serve it, so writing it would manufacture a diff the fixture invented.
- **`le` labels and cumulative counts are computed in ClickHouse**, with the same expressions `internal/promql/histogram_bucket.go` emits (`if(i > length(ExplicitBounds), '+Inf', toString(ExplicitBounds[i]))` and `arraySum(arraySlice(BucketCounts, 1, i))`). Re-deriving them in Go would put a second float-to-label formatter in the harness, and a `"0.5"` vs `"0.50"` disagreement would hard-diff every bucket series for a reason the fixture invented. Computing them once, server-side, makes the two sides equal by construction.

The mirror also needed a `(metric name, label set)` series key rather than a label-set-only one: `_count` and `_sum` share an attribute set off the same CH row, so the old key folded them into a single series.

## The gate

`test/regression/compat_promql_seed_corpus_test.go` (default lane, no ClickHouse needed) makes the invariant enforced rather than conventional:

- **`TestCompatCorpusReferencesOnlySeededMetrics`** reads `fixtureInserts` out of the seeder's Go source with `go/ast`, maps each entry to the PromQL series names it can serve (histogram table to `HistogramSyntheticNames`; gauge/sum table to the name itself; any other table to `t.Fatalf`, "teach this test"), then expands the corpus's template placeholders over a **closed** substitution table, parses every query with the upstream PromQL parser, and fails when a query names a family the seeder cannot produce.
- **`TestCompatSeedMirrorsEveryFixture`** pins the CH-side `fixtureInserts` list against the Prom-side `fixtureSources` + `histogramFixtureSources` lists as `(name, table)` sets. The seeder's own comment already said "keep them in lock-step"; nothing enforced it, and it had already drifted.

The one exemption is `nonexistent`-prefixed names, and it is asserted **in both directions**: such a name is never seeded, and an unseeded name must carry it. It names no specific metric and grows no entries, so it is a contract, not an allow-list.

Anti-vacuity guards, because a gate that silently inspects nothing is worse than no gate: non-empty servable set, non-empty referenced set, at least one `_bucket` expansion (so the histogram arm demonstrably fired), at least one `nonexistent` reference (so the exemption arm isn't dead), every substitution-table key actually consumed, every corpus case non-empty after decode, and at least one query parsed. An unparseable query is tolerated only when the corpus itself marks it `should_fail: true`.

## Deviations from the plan

1. **Batch-timestamp epoch expression.** The plan wrote `intDiv(toUnixTimestamp64Milli(toDateTime64({anchor:String}, 9)), 1000)`, which mixes an `Int64` with `UInt64` step arithmetic (ClickHouse does not reliably supertype that) and buries a bare `1000` unit conversion. Replaced with `toInt64(toUnixTimestamp(toDateTime({anchor:String})))`, and both step-derived terms are wrapped in `toInt64(...)` so the whole sum is signed. Same instant, no magic constant, no mixed-sign arithmetic.
2. **Histogram numerics stay inline in SQL.** The plan contradicted itself here (it proposed Go consts built with `fmt.Sprintf`, then concluded "inline in SQL"). Went with inline: declaring Go consts nothing reads would fail the unused-symbol lint, and `fmt.Sprintf`-templating SQL is the pattern we avoid. The geometry is documented in full in a comment block above the fixture instead — the cumulative ladder, the landing ranks, and why 509.5 was chosen (exactly representable in binary, so `_sum` carries no float drift).
3. **`intermittentInstanceOffset` lifted to a named const.** The plan inlined `instance_idx * 1000`. That is a meaning-bearing literal, so it is now a const whose name explains it, passed as a bound query parameter.
4. **`fixtureSources` grew two entries the plan didn't call out.** The plan only described the histogram arm of the mirror; the two new gauges also had to be added to the existing Value-shaped list, which had silently drifted behind `fixtureInserts`.

## Conflict resolution against `main`

This branch sat behind `main` long enough to collide with #1566 / #1571, which landed the same two-sample `rate` floor on the classic-histogram bucket-fanout path independently.

**The collision.** Both sides gate an anchor on how many samples its `[range]` window holds, with incompatible IR:

| | this branch | `main` (#1566 / #1571) |
|---|---|---|
| `chplan.RangeBucketFanout` field | `MinSamplesPerSeries int` + `SeriesKey Expr` | `MinSamples int` |
| emit | two-level collapse, `HAVING count() >= N` per source series | single collapse, `HAVING uniqExact(TimestampCol) >= N` |
| `sum_over_time` floor | not differentiated (hardcoded 2) | `stalenessMinSamples = 1` vs `rateMinSamples = 2` via `windowFor` |
| native-histogram paths | not covered | covered |

**Resolution: `main`'s side, wholesale.** It is merged, pinned by #1566's test, correctly differentiates the `rate`/`increase` floor from `sum_over_time`'s, and covers the native-histogram paths this branch's version did not. #1571 additionally rewrote the surrounding bucket-layout group-key code that this branch's version was built on. `internal/chsql/range_bucket_fanout_test.go` (a new file here, asserting only the dropped `MinSamplesPerSeries` design) was deleted with it.

The branch's own histogram_quantile **wire-name** fix was re-applied on top of `main`'s files: `histogramQuantileMatcherPredicate` in `internal/promql/histogram_quantile.go` plus `metricNamePredicateOn` in `internal/promql/lower.go`. An all-pinned selector folds to byte-identically the predicate that existed before, which is why the regenerated goldens show zero churn.

`main`'s version does leave a residual gap this branch's version closed: with `sum by (le)`, `histogramAggGroupBy` returns a nil group-by, so `uniqExact(TimestampCol)` counts timestamps across the **union** of source series rather than per series. Verified against the recorded golden (`GROUP BY anchor_ts, ExplicitBounds HAVING uniqExact(TimeUnix) >= 2`) — `ExplicitBounds` is a layout key, not a series key. That is out of scope for a seeding PR and is filed as **#1584**, citing the deleted test as its reproducer spec.

**Seed-mirror merge.** `compatibility/prometheus/cmd/seed/prom_remote.go` was the only hand-resolved file. `main` had added `demo_resource_latency_seconds` through a `histogram bool` discriminator on `fixtureSource` and a Go-side resource-attribute projection; this branch had rebuilt the same area around a separate `histogramFixtureSources` list. Kept this branch's architecture — `TestCompatSeedMirrorsEveryFixture` parses those lists with `go/ast` and requires two-field positional literals, which `main`'s three-field keyed form breaks — and folded `main`'s fixture into it as a second `histogramFixtureSources` entry. `main`'s Go-side resource-label projection moved server-side into `histogramFixtureSelect`, using cerberus's own `[^a-zA-Z0-9_]` sanitiser rather than the narrower `.`-to-`_` rewrite, so the two sides stay equal by construction. `main`'s corpus case (`histogram_quantile(0.5, sum by (k8s_namespace_name, le) (rate(demo_resource_latency_seconds_bucket[1m])))`) is retained.

**Generated baselines.** None were hand-resolved. This branch never touched `test/perf/solver-decision-baseline.json`, `test/perf/cardinality-baseline.json`, `test/rejection-parity/catalogue.json`, or `compatibility/parity-baseline.json`, so every merge took `main`'s side wholesale; all four are byte-identical to `origin/main`. `just update-golden` (which chains `update-solver-decision-baseline` before and `update-cardinality-baseline` after the golden pass) was then run on the merged tree and produced **zero diff**: cardinality 814 records before and after, solver-decision 458 before and after — zero lost, zero duplicated. Nothing under `test/spec/` moved either, which is the expected result of a wire-name fix that only alters the unpinned-`__name__` arm.

**Semantic (non-textual) conflict.** `main` #1579 added `test/regression/mirror_inventory_test.go` with its own package-level `sortedKeys`. Git merged both files without a marker; the redeclaration only surfaces at typecheck. This branch's helper is renamed `sortedSetKeys`.

**The PR's purpose is intact.** Taking `main`'s `MinSamples` reinstates no vacuous-pass path: the floor governs which anchors emit rows, not whether a metric is seeded. The seeds for all three families, the deleted allow-list prose in the corpus header, and both gate tests are fully retained. `TestCompatCorpusReferencesOnlySeededMetrics` was negative-controlled after the merge — appending an unseeded `sum(demo_totally_unseeded_metric)` to the corpus makes it fail with "corpus queries reference metric families the seeder cannot serve", so the gate is enforcing rather than hollow.

**Resource-label mirror scope (12 real compatibility diffs, fixed at the source).** Moving `main`'s resource-attribute projection server-side generalised it from the one fixture `main` ran it on (`demo_resource_latency_seconds`, whose `ResourceAttributes` is solely `k8s.namespace.name=prod`) to every histogram fixture — including `demo_api_request_duration_seconds`, whose resource layer is `service.name=demo`. The mirror then wrote a `service_name="demo"` label onto the reference Prometheus that cerberus never emits, because the resource arm excludes keys backed by a dedicated top-level column (`service.name` → `ServiceName`) and the seeder deliberately leaves that column unpopulated. Result: 12 genuine diffs (726/738), every one with **identical values** and a label set differing by exactly that label — the `{job="demo", __name__!~"demo_cpu_usage_seconds.*"}` selector plus all 11 `histogram_quantile` cases over `demo_api_request_duration_seconds`. All 28 differing series on the reference side carried `service_name`; none on the cerberus side did; no other metric family appeared. Fixed by filtering the fold through the production predicate `promql.DedicatedResourceLabelExcluded` rather than a hardcoded key list, so the mirror promotes exactly what cerberus's resource arm promotes. No allow-list, no baseline relaxation, no tolerance. `demo_resource_latency_seconds` is unaffected — `k8s.namespace.name` has no dedicated column, so it is still promoted on both sides, which is the point of that fixture.

Follow-up tracked in #1584.


tsouza:
Seeding the three metrics did exactly what it was supposed to do: it exposed real divergences the empty-vs-empty scoring was recording as passes.

\`compatibility/prometheus\` now reports **REGRESSED — 13 cases** the parity baseline records as agreeing with reference Prometheus. From the run-30619769095 report artefact, two shapes:

1. **Eleven `histogram_quantile(φ, rate(demo_api_request_duration_seconds_bucket[1m]))` cases** (plus the `scalar(vector(0.9))` and `{__name__=~"demo_api_request_duration_seconds_.+"}` variants). Cerberus emits **two extra leading samples** at the first two steps that reference Prometheus does not produce; the remaining 357 samples are identical. That is the signature of the classic-bucket rate fan-out emitting at steps whose lookback window holds fewer than two samples, where Prometheus's `rate` requires ≥2. This is **not** a seed misalignment — `prom_remote.go` reads timestamps straight from ClickHouse via `toUnixTimestamp64Milli(TimeUnix)`, so both backends receive identical timestamps.

2. **`{job="demo", __name__!~"demo_cpu_usage_seconds.*"}`** — the reference returns `demo_api_request_duration_seconds_count` series that cerberus omits, pointing at a raw selector not enumerating the synthetic histogram companion names.

Per the no-allow-list rule these get fixed at the source; the parity baseline is not an option for REGRESSED rows. Tracked as a blocking follow-up before this PR can merge.

Issue #1535, opened 2026-08-01
promql: instant-mode histogram_quantile agg path missing rate's two-sample floor


tsouza:
## What

#1424 gives the **range-mode** aggregated histogram path a minimum-sample floor: `chplan.RangeBucketFanout.MinSamples`, emitted as `HAVING uniqExact(<TimestampCol>) >= <MinSamples>` in `internal/chsql/range_bucket_fanout.go`. It carries reference PromQL's rule that `rate` / `increase` need two points in `[range]` to span a delta and emit **nothing** at an anchor holding fewer.

The **instant-mode** siblings — `lowerHistogramQuantileAgg` (`internal/promql/histogram_quantile.go:577`) and `lowerHistogramQuantileNativeAgg` (`:884`) — apply `stalenessLowerBoundExpr` to bound the window but then collapse through a plain `chplan.Aggregate`, which has no equivalent gate. So an instant query `histogram_quantile(0.5, sum by(le)(rate(X_bucket[1m])))` evaluated at a timestamp whose `[1m]` window holds a single scrape returns a value where reference Prometheus returns an empty result.

## Why it wasn't fixed in #1424

`chplan` has no `Having` slot on any node — `Having` exists only on `chsql.QueryBuilder`. The range fix could thread `MinSamples` through `RangeBucketFanout` because that node owns its own emitter. Giving `Aggregate` the same capability means a new plan-level field plus emitter support on a node used by every head, which is a materially larger and riskier change than the range-path fix it would accompany.

## Repro shape

An instant-query spec fixture seeding one scrape inside the `[1m]` window and asserting zero rows would fail today. Note the fixture harness's range window is fixed at `[2026-01-01 00:00:00, 00:05:00]` (`internal/promql/lower_test.go`), so this needs the instant lane rather than `-- range_step --`.

## Suggested shape

Add a min-distinct-timestamps field to `chplan.Aggregate` (or a small dedicated node), emit it as `HAVING uniqExact(<ts>) >= n` the way `emitRangeBucketFanout` now does, and thread `histogramAggShape.windowMinSamples` into both instant lowerings — the matcher already computes the value.

Found while fixing the range-mode divergence in #1424.

Issue #1486, opened 2026-08-01
k3d Grafana surface-crawl inventory has never been bootstrapped — the stack closest to production has zero crawl coverage


tsouza:
The Grafana surface crawler was generalised into a two-stack framework (compose + k3d). The k3d half looks shipped — it has a stack config, its own inventory file, and its own CI matrix leg — but its inventory is empty and the leg that would surface that is skipped on PR and on push.

**Evidence on `main`:**

- `test/e2e/playwright/crawl/grafana-surface-inventory.k3d.json` is `{"stack":"k3d","surfaces":[]}`. `git log --oneline` on that path shows one commit (the one that created it) — it has never been bootstrapped.
- `test/e2e/playwright/crawl/lib.ts` `assertInventoryBootstrapped` throws whenever `surfaces.length === 0` and `CERBERUS_UPDATE_INVENTORY` is unset. It is called from `crawl/crawl.spec.ts` at two sites.
- `.github/workflows/e2e.yml` gates the k3d crawl shard on `RUN_SHARD: ${{ matrix.crawlStack != 'k3d' || github.event_name == 'schedule' || github.event_name == 'workflow_dispatch' }}` — a no-op on `pull_request` and on `push`.

**Consequence.** Every k3d crawl run can only throw the bootstrap error, and the shard that would throw it doesn't run on PR or on merge. The repo presents a per-stack ratchet as if both stacks were ratcheted; in fact k3d — the environment closest to how cerberus is actually deployed — has zero Grafana surface-crawl coverage. Any surface regression that manifests only under k3d (datasource provisioning, dashboard rendering against the k3d-deployed cerberus) is invisible to CI, and the emptiness of the inventory is what keeps it that way.

This is a hollow-green shape: the mechanism is correct, the gate exists, and it measures nothing because its input set is `[]`.

**Fix:** run the crawl once under k3d with `CERBERUS_UPDATE_INVENTORY` set, commit the resulting inventory, and confirm the shard's skip condition is the intended one once there is something for it to assert.

Found independently by two sweep passes (chunk 5, S5-052/053; pass 3 chunk 4, PRs #772/#773). Verified against `origin/main`; not addressed by any open PR.

tsouza:
Duplicate of #1467 — same defect: `test/e2e/playwright/crawl/grafana-surface-inventory.k3d.json` has an empty `surfaces` array, so `assertInventoryBootstrapped()` has nothing to assert and the k3d crawl shard emits no coverage signal.

[...]

PR #1550, opened 2026-08-01
ci(forbid-deferral): enforce the no-deferrals rule as a required status check

tsouza:
## What

`forbid-deferral` — a new required status check, sibling to `forbid-skip`, that
makes the repository's no-deferrals rule **enforced** rather than conventional.

`forbid-skip` rejects a test that declines to assert. This rejects a change that
names work it is not doing and walks away. The rule it enforces is already in
`CLAUDE.md`: when work surfaces that genuinely belongs outside a change's scope,
you `gh issue create` — you do not write a sentence about it into a description,
a commit message, or a comment and move on. A sentence is not a resting place;
once the pull request merges it appears in no list, no gate reports on it, and
nobody is assigned. An issue stays open until somebody closes it.

## Scope — the change's own additions, and nothing else

Three surfaces, all of them things this change itself introduced:

1. the pull request description,
2. the commit messages in `BASE_SHA...HEAD_SHA`,
3. the added (`+`) lines of that diff.

The tree at large is **not** scanned, and that is the load-bearing design
decision. Measured on `main`: 145 occurrences of one of these words, 86 of
another, 74 of a third, 23 and 18 of two more — and the overwhelming majority
are legitimate architecture prose. `internal/schema/otel.go` describing a
boundary as being outside the unknown-name fan-out is a factual statement about
a system, not a parked task. A gate that fired on those would be unusable, and
an unusable gate gets routed around, which is worse than no gate.

This is scoping, not an allow-list. There is no tolerance file, no exemption
list, and no way to park a violation. The single accepted resolution is to open
the issue.

### Why the commit-message surface exists (measured, not assumed)

A sweep of all 1326 commits on `main` found **217 commits whose bodies carry
text of this class**, of which **178 carry it ONLY in intra-branch commit
messages — absent from the pull request description entirely**. A
description-only gate would miss roughly **82%** of the real population. That is
the reason surface 2 is in the list.

The same sweep is what calibrated the table. The bulk of the RAW regex hits were
false positives of three shapes, and all three are pinned as explicit non-match
tests in `forbid-deferral.test.mjs`:

| False-positive shape | Why it is not a deferral | How the table avoids it |
| -------------------- | ------------------------ | ----------------------- |
| Go's `defer` statement | `defer rows.Close()` is resource cleanup | every bare word is `\b`-anchored; only the English participle matches |
| the phrase that records COMPLETED work | "…&nbsp;to #1234" names work that already happened | only the LABEL forms match — a heading, or the colon that introduces a list nobody has taken |
| a change that DELETES a marker line | removing stale prose is the fix, not the offence | markers are read from `+` lines only; removed lines have no position in the new file |

Two further rows are qualifier-scoped for the same reason: the bare phrase about
scope and the bare phrase about state are ordinary prose (the first appears in
the hundreds on `main`), so only the variants that name **this change** as the
place the work did not happen match.

## The marker table

Ten rows, in one exported constant (`DEFERRAL_MARKERS` in
`.github/scripts/forbid-deferral.mjs`) so it is testable and reviewable in
isolation. Matching is case-insensitive and word-boundaried.

| id | what it catches |
| -- | --------------- |
| `code-work-marker` | a conventional source-comment work marker |
| `unfixed-here` | a finding named as unresolved by this change |
| `followup-label` | a heading or label introducing work nobody has taken |
| `not-yet-here` | a capability named as still missing after this change |
| `deferral-to-later` | work postponed instead of done |
| `left-for-later` | work handed to some later change |
| `later-pr` | work assigned to a pull request that does not exist yet |
| `punted` | a decision put off rather than taken |
| `re-visit` | a promise to look at something again |
| `pr-scoped-exclusion` | work excluded by naming this pull request's scope |

Each row's exact pattern is in the module; the test asserts every row fires on a
real example (a row that can trigger on nothing gates nothing) and that no row
matches the source text, id, or description of any row — including its own,
which is what lets this pull request pass its own gate.

## The requirement

Every match must be accompanied, **in the same paragraph** (prose) or **within
`CITATION_WINDOW_LINES` either side** (diff), by a reference to an issue in this
repository — `#<n>` or a full `github.com/tsouza/cerberus/issues/<n>` URL. Each
number is resolved through the API and must be

- an **issue**, not a pull request (GitHub's issues endpoint returns both; the
  `pull_request` key discriminates), and
- **open**.

Failure emits one `::error::` per violation naming the surface, the exact
location, the marker class and the marker text, plus the remedy.

## Anti-vacuity

A gate that inspects nothing reports the same green as a gate that inspects
everything. So the scanner asserts positively that it did work, and every
missing or malformed input fails loudly:

```text
### PR_BODY unset on a pull_request run
::error title=forbid-deferral::forbid-deferral: PR_BODY is unset on a pull_request run — the description surface would go uninspected. Wire `PR_BODY: ${{ github.event.pull_request.body }}` into the step env.
exit=1

### GITHUB_TOKEN unset
::error title=forbid-deferral::forbid-deferral: GITHUB_TOKEN is unset — resolving a cited issue needs issues:read
exit=1

### empty commit range (base == head)
::error title=forbid-deferral::forbid-deferral: be58f3aac87a..be58f3aac87a resolved to zero commits — the gate would inspect no change at all
exit=1

### unresolvable base with no parent to fall back to
::error title=forbid-deferral::forbid-deferral: no base commit: BASE_SHA does not resolve and e9058f102faf has no parent — the gate would scan an undefined range
exit=1
```

An empty marker table is rejected on the same grounds, and every pattern is
compiled before the scan so a malformed row cannot silently match nothing.

## Proof that it activates

Not "the code looks right" — real end-to-end runs of the CLI, against real
issue numbers in this repository.

### FAIL — all three rejection arms, one run

A throwaway repository seeded with an unlinked work marker on an added line, a
commit message citing a **closed issue** (`#1486`), and a description citing a
**merged pull request** (`#1143`):

```text
- description: the pull_request event payload (57 chars)
- commits: 1 in e9058f102faf..be58f3aac87a
- diff: 1 file(s) at --unified=3
- marker table: 10 rows
- markers found: 3, issue numbers resolved: 2
::error title=forbid-deferral::PR description line 3: a heading or label introducing work nobody has taken ("Follow-up:") is not tracked by an open issue within the same paragraph — #1143 is a pull request, not an issue. Remedy: open an issue for the work (`gh issue create`) and cite it as #<n> beside the marker. …
::error title=forbid-deferral::commit be58f3aac87a message line 3: a heading or label introducing work nobody has taken ("Follow-up:") is not tracked by an open issue within the same paragraph — #1486 is a closed issue, so the work it names is untracked. …
::error title=forbid-deferral::a.go:5: a conventional source-comment work marker ("TODO") is not tracked by an open issue within ±3 lines — it cites no issue at all. …
::error title=forbid-deferral::forbid-deferral: 3 untracked deferral(s) across 3 surface(s).
exit=1
```

All three fixtures are live: `#1143` really is a merged pull request, `#1486`
really is a closed issue, `#1535` really is an open one.

### PASS — a properly-linked deferral, and the bare phrase

Same repository, same shapes, but the postponement cites the **open** issue
`#1535`, and both the code comment and the description use the **bare**
scope phrase that must not match:

```text
- description: the pull_request event payload (106 chars)
- commits: 1 in e9058f102faf..3803bf8caf02
- diff: 1 file(s) at --unified=3
- marker table: 10 rows
- markers found: 1, issue numbers resolved: 1
::notice title=forbid-deferral::forbid-deferral: 1 marker(s) across 1 commit(s), 1 file(s) and the description — every one tracked by an open issue.
exit=0
```

`markers found: 1` is the point: the bare phrase in the comment and in the
description produced **no** candidate, while the tracked postponement in the
commit message produced one and was cleared by `#1535`.

### PASS — this pull request, against itself

The gate was run over this branch's own range and description before the branch
was pushed. It passes: the module, its test, and the documentation all name
marker CLASSES rather than reproducing marker text, and every pattern is written
so it cannot match its own source (a bare word is spelled `\b`-prefixed, whose
literal `b` denies the leading word boundary; a phrase is joined by `\s+` so the
literal phrase never appears). This is the same move `forbid-skip.mjs` makes
when it writes its own banned pattern as an escaped regex.

### Unit guard

`node --test .github/scripts/forbid-deferral.test.mjs` — 24 tests, run as the
step BEFORE the gate in the same job: table reachability, the three
false-positive shapes, the two qualifier rows, citation extraction, fenced-block
handling, paragraph scoping, diff parsing (including new-file line numbering and
the window), and every verdict arm.

## Wiring

- **`.github/workflows/ci.yml`** — new job named exactly `forbid-deferral`, on
  `pull_request` **and** `push` (ci.yml already triggers on `main` +
  `release/*.x`, which is what makes the context eligible as a required check and
  what stops `release-gate-drift.yml` reporting it as lane drift). Job
  `permissions:` grant `issues: read` + `pull-requests: read`. No `setup-node`
  (ubuntu-latest ships node), no `continue-on-error`.
- **`.github/workflows/release.yml`** — added to `RELEASE_REQUIRED_CHECKS`.
  Skipping this is the exact trap hit when `perf-guards` was promoted in #1446:
  `release-gate-drift.yml` would report the new context as protection drift.
- **`test/regression/release_required_checks_test.go`** — added to
  `branchProtectionContexts`, keeping the totality pin honest.
- **`docs/test-strategy.md`** — new row in the CI-gate inventory, and the
  required-check prose corrected (it claimed fifteen; the live set is eighteen,
  nineteen once this is promoted).
- **`.github/scripts/README.md`** — module + env-contract inventory entry.
- **`CLAUDE.md`** — count corrected from eighteen to nineteen, the new context
  added to the list, and the no-deferrals rule restated as enforced.

Branch protection is promoted to include `forbid-deferral` once this lane posts
green here.


tsouza:
### CI-side activation, on this pull request

The evidence in the description is from local runs of the CLI. For completeness,
the lane itself on this branch (`5cda44d`, job
[91373133149](https://github.com/tsouza/cerberus/actions/runs/30701478832/job/91373133149)):

```text
1 success  Set up job
2 success  Run actions/checkout@v7
3 success  Assert the deferral scanner matches, and does not over-match
4 success  Reject untracked deferrals
8 success  Post Run actions/checkout@v7
9 success  Complete job
```

Both steps executed — the unit guard first, then the gate over this pull
request's own description, its commit message, and the `+` lines of its
eight-file diff. `conclusion: success`.

That the gate actually inspects rather than waves through was shown twice
before this run: it FAILED this branch on its first attempt, on two real
violations in the additions themselves — the module header used a banned
participle in a sentence of its own prose, and the `CLAUDE.md` bullet being
edited quoted four marker literals inline. Both were reworded to name marker
CLASSES instead, which is the same discipline the module, its test and the docs
all follow. The gate found them; nothing else would have.

### Out-of-scope finding, filed rather than parked

Reconciling the required-check inventory for this change surfaced that
`strict-scan` and `CodeQL` are branch-protection required on `main` but appear
in neither `RELEASE_REQUIRED_CHECKS` nor `RELEASE_INFORMATIONAL_CHECKS`, so a
maintenance publish never waits for them. Filed as #1554 — which is, fittingly,
the rule this pull request is making enforceable.

Issue #1591, opened 2026-08-02
histogram_quantile: first-bucket interpolation ignores a non-positive lowest explicit bound


tsouza:
## What

`histogram_quantile` interpolates from a hard-coded lower edge of `0.0` when the
target rank falls in the **first** bucket. Prometheus does not: when the first
bucket's upper bound is itself `<= 0` — a histogram with negative or zero
explicit bounds, which OTel permits — it returns that upper bound directly
instead of interpolating up from zero.

For a histogram whose bounds start at, say, `-10`, cerberus returns a value
interpolated across `[0, -10]` (a negative-width bucket) where Prometheus returns
`-10`. A wrong number, not an empty result.

## Where

Reference — `promql/quantile.go`, `bucketQuantile`, in the switch after the rank
search:

```go
case b == 0 && buckets[0].UpperBound <= 0:
    return buckets[0].UpperBound
```

(line 154 of the `tsouza/prometheus` `cerberus-parser` fork's copy.)

Cerberus — `internal/chsql/histogram_quantile.go`, in
`histogramQuantileValueFrag`:

```go
boundLo := If(Eq(idx(), InlineLit(1)), zeroF, boundAt(true))
cumLo   := If(Eq(idx(), InlineLit(1)), zeroF, cumAt(true))
```

`idx() == 1` is exactly upstream's `b == 0`. Cerberus takes the `zeroF` branch
unconditionally; upstream first checks the sign of `buckets[0].UpperBound` and
short-circuits the whole interpolation when it is non-positive.

## Fix sketch

The guard is one more `If` in the same Frag tree — the value branch becomes

```text
if(idx = 1 AND ExplicitBounds[1] <= 0, ExplicitBounds[1], <existing interp>)
```

composed with typed Frags (`And` / `Lte` / `Subscript`), no raw SQL.

## Acceptance

A `test/spec/promql/` TXTAR fixture seeded with a classic histogram whose
`ExplicitBounds` start at a negative value, with `-- expected_rows --` matching
what reference Prometheus returns for the same data. `compatibility/prometheus`
is the arbiter.

Found while fixing #1569 / #1570; out of that PR's scope because it is a
different divergence (interpolation edge case, not bucket-layout merging) with
its own seed shape.

Issue #1590, opened 2026-08-02
histogram_quantile: parameterised aggregations over classic buckets still fold to empty


tsouza:
## What

`histogram_quantile(phi, <agg> by(le) (rate(<classic>[5m])))` returns an empty
result — where Prometheus returns rows — when `<agg>` is one of the four
**parameterised** aggregations: `quantile`, `topk`, `bottomk`, `count_values`.

This is the residue of #1569. The reducer family (`sum`, `avg`, `min`, `max`,
`count`, `group`, `stddev`, `stdvar`) is fixed by the classic-bucket
layout-merge change, which folds every contributing row's ladder in the
cumulative per-`le` domain over the union of bounds. The four ops above are
rejected one step earlier and never reach that fold.

## Where

`internal/promql/histogram_quantile.go`, in `matchHistogramAggIdiom`:

```go
if a.Param != nil {
    return histogramAggShape{}, false
}
```

`ok=false` makes the caller fall through to `unwrapVectorSelector`, and the plan
degenerates to `&chplan.Filter{Predicate: &chplan.LitBool{V: false}}` over the
gauge/sum tables — an empty result rather than an error.

## Why it is not one `case` in the fold switch

The four ops do not share a shape, and only one of them is a fold at all.

- **`quantile(phi, ...)` IS a per-group reducer.** It collapses each `le` group
  to one value exactly like `min` / `stddev` do, so it fits
  `classicBucketLadderFold` directly as
  `arrayReduce('quantile(<phi>)', rungs)`. The only new work is threading the
  scalar `Param` through `histogramAggShape` and requiring it to be a
  `*parser.NumberLiteral` (a non-constant param has no CH counterpart, since the
  aggregate-function name is a literal string in `arrayReduce`).
- **`topk(k, ...)` / `bottomk(k, ...)` are SELECTORS, not reducers.** They return
  the original input series with their full label sets intact, filtered to the
  top/bottom `k` within each `by(le)` group. So the output of
  `topk by(le)(k, rate(x_bucket[5m]))` is not one series per `le` — it is a
  subset of the input series, still carrying `job` / `instance` / … , and
  `histogram_quantile` then groups those by every label except `le`. A series can
  survive at some `le` values and not others, producing a genuinely partial
  ladder. Nothing in the current merge shape expresses that.
- **`count_values("label", ...)` MINTS a new label** whose values are sample
  values, so it changes output series identity in a way `histogram_quantile`'s
  grouping has no representation for.

[...]

Issue #1584, opened 2026-08-02
histogram_quantile(phi, sum by (le) (rate(...))) admits source series below the two-sample rate floor


tsouza:
## Symptom

For `histogram_quantile(phi, sum by (le) (rate(<classic hist>[R])))`, cerberus admits bucket counts from source series that hold **fewer than two samples** in the `[R]` window, while reference PromQL's `rate()` drops those series entirely. The result is a quantile computed over a bucket ladder the reference never assembled.

## Mechanism

`histogramAggGroupBy` (internal/promql/histogram_quantile.go:812) folds `sum by (le)` — and `sum by ()` — to `return nil, nil, emptyAttrsMap()`: `le` is dropped from the by-list because the bucket distribution lives in the `BucketCounts` array rather than in an `Attributes` key, and nothing is left. `chplan.RangeBucketFanout.GroupBy` is therefore empty of any series-identity key.

`internal/chsql/range_bucket_fanout.go:179` then attaches the two-sample floor to that collapse:

```go
if r.MinSamples > fanoutNoMinSampleFilter {
    collapse.Having(Gte(Call("uniqExact", Col(r.TimestampCol)), InlineLit(int64(r.MinSamples))))
}
```

The recorded golden for `test/spec/promql/histogram_quantile_classic_agg_rate_min_samples.txtar` shows the emitted shape:

```sql
GROUP BY `anchor_ts`, `ExplicitBounds` HAVING uniqExact(`TimeUnix`) >= 2
```

`ExplicitBounds` (added by #1571 so ladders with different layouts never blend) is a *layout* key, not a *series* key. So `uniqExact(TimeUnix)` counts distinct timestamps across the **union** of every source series sharing an anchor and a bucket layout.

Concretely, with series A holding samples at `t1, t2` inside the window and series B holding only `t1`:

- reference PromQL emits `rate(A)` alone;
- cerberus sees `uniqExact(TimeUnix) = |{t1, t2}| = 2`, passes the HAVING, and `sumForEach(BucketCounts)` folds **both** A and B into the ladder that feeds the interpolation.

The correct floor is per source series, not per group.

## Scope

Only bites when the group key carries no series identity — i.e. `sum by (le)` / `sum by ()`. `agg == nil`, `sum without (…)`, and `sum by (<real labels>, le)` all bind a series-identity key (`histogramIdentityExpr` / `attributeLookupExpr`), so `uniqExact` is already per series there.

[...]

Issue #1569, opened 2026-08-01
histogram_quantile: non-ladder-preserving aggregations over classic buckets still fold to empty


tsouza:
## What

`histogram_quantile(phi, <agg> by(le) (rate(<classic>[5m])))` returns an empty result for every aggregation except `sum` and `avg`. The affected set is `min`, `max`, `count`, `group`, `stddev`, `stdvar`, `quantile`, `topk`, `bottomk`, `count_values`.

## Where

`internal/promql/histogram_quantile.go` — `matchHistogramAggIdiom` gates on `classicBucketAggOpIsLadderPreserving(a.Op)`, which returns true only for `parser.SUM` and `parser.AVG`. Anything else returns `ok=false`, the caller falls through `unwrapVectorSelector`, and the plan degenerates to `&chplan.Filter{Predicate: &chplan.LitBool{V: false}}` — an empty result rather than an error.

`sum` and `avg` were widened from sum-only in the PR that adds this issue; the remaining ops are the residue.

## Why the residue is not a one-line widening

OTel-CH classic histograms store **per-bucket** counts in `BucketCounts` alongside `ExplicitBounds`. The current lowering reduces them element-wise with `sumForEach` and lets `arrayCumSum` build the ladder afterwards. That is only correct for reducers that **commute** with the per-bucket → cumulative running total:

- `sum` commutes: `cumsum(Σ_rows c[i]) == Σ_rows cumsum(c)[i]`.
- `avg` is `sum` divided by a row count that is constant across every bucket position within a layout group, and `histogram_quantile` interpolates the *ratio* of cumulative counts to the total, so the constant scale cancels.
- `min` / `max` do **not** commute. Rows `[10, 0]` and `[0, 10]`: element-wise max is `[10, 10]`, whose cumsum is `[10, 20]`. Prometheus, aggregating already-cumulative per-`le` samples `[10, 10]` and `[0, 10]`, gets `[10, 10]`. Different ladder, different quantile.

The same argument rules out every remaining op — each of them needs the ladder accumulated **before** the reduction, not after.

## What a fix needs

1. A cumulative-ladder input mode in `chplan.HistogramQuantile` (or an equivalent emitter path) so the accumulation happens per-row and the reducer runs over already-cumulative rungs.
[...]

PR #1592, opened 2026-08-02
fix(promql): merge classic bucket layouts in the cumulative per-le domain

### LINKED ISSUE #1570: histogram_quantile: mismatched ExplicitBounds layouts group separately instead of union-merging by le
## What

When a classic-histogram aggregation group contains rows carrying **different** `ExplicitBounds` layouts, cerberus now emits **one row per layout** — each quantile computed against the ladder it was actually measured on. Reference Prometheus instead merges the layouts into a single ladder by taking the union over `le` and emits **one** row.

Both answers are defensible; they are not the same answer. This is a real, verified parity divergence against reference Prometheus.

## Where

`internal/promql/histogram_quantile.go` — `classicBucketAggGroupBy` appends `ExplicitBounds` to the aggregation's group keys (used by both the instant path in the same file and `lowerHistogramQuantileClassicAggRange` in `histogram_quantile_range.go`).

## Why it is like this

Before, the lowering reduced with `sumForEach(BucketCounts)` while picking bounds with `any(ExplicitBounds)`. Position `i` of one layout was added to position `i` of another and the interpolation ran against a ladder matching neither — a silently wrong number, no error, no empty result. That is fixed by making the layout a group key.

Grouping was chosen over failing loudly because it never produces a wrong number and is a no-op in the single-layout case (no existing fixture's computed values changed). The honest cost is the divergence above.

## The observable difference

`test/spec/promql/histogram_quantile_sum_by_le_mixed_layouts.txtar` seeds two hosts with disjoint layouts (`[1, 2, 3]` and `[100, 200, 300]`) and asks `histogram_quantile(0.5, sum by(le) (rate(...[5m])))`. Cerberus emits two rows, `0.5` and `300`, both with an empty label set — i.e. **two rows with identical labels**, which the Prometheus data model does not otherwise produce. Reference Prometheus would emit one row against the union ladder.

## What settling this needs

The `compatibility/prometheus` lane, which is the source of truth for reference semantics. It was unavailable when this was found (Docker Hub pull quota), so the reference behaviour here was established by reading upstream `BucketQuantile` / `coalesceBuckets` in `promql/quantile.go` rather than by running the harness. Confirm against the lane, then decide:

- match upstream by union-merging over `le` before interpolating, or
- keep per-layout grouping and add a distinguishing label so the output stays a valid Prometheus result.

Real producers emit one stable layout per series, so this only bites when a producer changes its bucket boundaries mid-window, or when `by (...)` drops the label that told two differently-configured series apart.

tsouza:
Fixes two classic-histogram defects that turn out to be one mechanism seen from
two sides, so they ship together rather than regenerating the same goldens twice.

## Verification first — both defects were live on `origin/main`

Confirmed at each issue's own cited position before writing anything (#1571 had
already landed in this exact area, so "already fixed" was a real possibility):

- **#1569** — `internal/promql/histogram_quantile.go`,
  `classicBucketAggOpIsLadderPreserving` returned `op == parser.SUM || op ==
  parser.AVG`, and `matchHistogramAggIdiom` gated on it. Empirically:
  `histogram_quantile(0.5, max by(le) (rate(<classic>[5m])))` lowered on
  `origin/main` to `Filter predicate=false` over `Aggregate
  groupBy=[Attributes["le"] AS gkey_0] funcs=[max(Value) AS Value]` over
  `Scan(merge(otel_metrics_gauge|otel_metrics_sum))` — no `HistogramQuantile`
  node in the plan at all, and `[]` rows out of chDB.
- **#1570** — `classicBucketAggGroupBy` appended `ExplicitBounds` as a grouping
  key. Empirically, `histogram_quantile_sum_by_le_mixed_layouts.txtar` returned
  **two** rows for one requested series.

## Root cause (shared)

Cerberus aggregated classic-histogram rows in the **positional per-bucket array**
domain — `sumForEach(BucketCounts)`, with `arrayCumSum` applied afterwards —
while Prometheus aggregates in the **cumulative per-`le`** domain, where each
input series is one already-cumulative float sample per `le`.

Positional reduction equals Prometheus's only when both hold:

1. every row in a group shares one bucket layout, and
2. the reducer commutes with `arrayCumSum`.

#1571 enforced (1) by promoting the layout to a grouping key — which is exactly
#1570: the layout leaked into series identity and split one requested series into
one output series per layout. And (2) was enforced by a `sum`/`avg` whitelist —
which is exactly #1569: every other aggregation fell off the histogram path
entirely and degenerated to `Filter predicate=false` against the gauge/sum
tables, i.e. an empty result where Prometheus returns rows.

Neither restriction is needed once the fold happens in the cumulative domain.

## The fix

Three stages, all composed as typed `chplan` exprs / `chsql` Frags:

1. **`Aggregate`** collects each contributing row's layout and counts verbatim —
   `groupArray(ExplicitBounds)` + `groupArray(BucketCounts)` — under the
   **user's** grouping keys only. The bucket layout never enters series identity.
2. **First `Project`** builds the union of every row's bounds
   (`arraySort(arrayDistinct(arrayFlatten(...)))`), evaluates each row's
   cumulative count at every union bound, restricts each rung to the rows whose
   own layout actually carries that bound (a series with no `le=u` bucket
   contributes no series at that `le`, same as Prometheus), and folds those rungs
   with the operator the user asked for: `arraySum` / `arrayAvg` / `arrayMin` /
   `arrayMax` / `length` / `arrayReduce('stddevPop'|'varPop')` / the constant
   `group` writes. One `+Inf` rung is appended from each row's total.
3. **Second `Project`** applies Prometheus's
   `ensureMonotonicAndIgnoreSmallDeltas` — a prefix maximum — which a
   cross-layout merge can genuinely need.

`chplan.HistogramQuantile.BucketCountsCumulative` then tells the emitter the
array is already a ladder, so `internal/chsql/histogram_quantile.go` skips
`arrayCumSum` and reads the observation total off the ladder's top rung instead
of summing the array.

`sum by(le)` now legitimately leaves the inner `Aggregate` **keyless** (`le` is
dropped — the distribution lives in the parallel arrays, not in `Attributes` —
and the layout is no longer a key). The empty-input guarantee therefore moves
from "there is always a grouping key" to `DropEmptyOnNoGroup`'s `count() > 0`
guard. The affected tests are re-pinned to assert that guard reaches the emitted
SQL, not relaxed.

## Before / after — the `expected_rows` evidence

A SQL-shape-only fixture cannot distinguish "returns empty" from "returns
correctly", which is the entire defect, so each fixture carries `-- seed --` and
`-- expected_rows --`.

| fixture | `origin/main` | this PR |
|---|---|---|
| `histogram_quantile_max_by_le` (new) | `[]` — plan root is `Filter predicate=false`, no `HistogramQuantile` node | one row, `1` |
| `histogram_quantile_count_by_le` (new) | `[]` — same shape | one row, `0.5` |
| `histogram_quantile_sum_by_le_layout_change_midwindow` (new) | **two** rows: `{} 0.5` and `{} 15` | **one** row, `15` |
| `histogram_quantile_sum_by_le_mixed_layouts` (existing) | **two** rows: `{} 0.5` and `{} 300.0` | **one** row, `1` |

The first two are #1569 (one series each, seeded across two hosts with the same
layout). The third is #1570 in its purest form: a **single** series `host=a`
whose layout changes mid-window — `[4,0,0,0]`/`[1,2,3]` at `t0`, then
`[0,6,0,0]`/`[10,20,30]` at `t1`. One series in, one series out. The fourth is
the fixture #1571 added, now returning the merged answer instead of one row per
layout.

## Value churn on existing fixtures — why, and its exact boundary

Eight existing fixtures changed their `expected_rows` values. Every one of them
seeds `len(BucketCounts) == len(ExplicitBounds)`, which violates the node's own
documented invariant (`len(BucketCounts) == len(ExplicitBounds) + 1`, for the
mandatory `+Inf` overflow bucket). The merge now always emits `len(bounds)+1`
rungs, so those malformed seeds read as `+Inf` count `0`.

Hand-verified on `histogram_quantile_sum_by_le`: the old path's
`arrayFirstIndex` hit `idx == length` and returned the top bound `1.0`; the new
path interpolates `0.5 + 0.5 * 5.2/6 = 0.9333…`, which is exactly what
Prometheus returns for the well-formed equivalent seed.

**Every well-formed fixture — the `(4,3)` and `(3,2)` seeds: `plateau`,
`duplicate_bounds`, `zero_count_plateau`, `plateau_overflow`,
`classic_agg_rate_min_samples`, `classic_bare_rate_min_samples` — is
byte-identical in `expected_rows`.** The single well-formed exception is
`mixed_layouts`, which is the target defect.

## Generated baselines

Regenerated with `just update-golden`; never hand-edited. Record-level diff keyed
on fixture name:

| baseline | before | after | added | removed | duplicated | changed |
|---|---|---|---|---|---|---|
| `test/perf/cardinality-baseline.json` | 814 | 817 | 3 (exactly the new fixtures) | 0 | 0 | 1 |
| `test/perf/solver-decision-baseline.json` | 458 | 461 | 3 (same three) | 0 | 0 | 0 |

The one changed cardinality record is
`promql/histogram_quantile_agg_empty.peak_intermediate` `0 → 1`: the keyless
aggregate now emits the `_cerb_n` count-guard subquery, so one intermediate row
exists before the outer `WHERE` drops it. In scope, and it is the same guard the
re-pinned unit test asserts.

## Scope

Closes #1570.

Addresses the reducer half of #1569 — `min` / `max` / `count` / `group` /
`stddev` / `stdvar` now return the right series alongside `sum` / `avg`. The four
**parameterised** aggregations (`quantile` / `topk` / `bottomk` /
`count_values`) are rejected earlier, at `a.Param != nil` in
`matchHistogramAggIdiom`, and do not share this shape — `topk` / `bottomk` are
selectors that keep their inputs' full label sets, and `count_values` mints a new
label, so neither collapses an `le` group to a single value. Filed as **#1590**
rather than left as prose here, so #1569 is intentionally **not** auto-closed.

Also filed **#1591** while reading upstream's `bucketQuantile`: cerberus's
first-bucket interpolation hard-codes a `0.0` lower edge and has no counterpart
to upstream's `b == 0 && buckets[0].UpperBound <= 0` short-circuit, so a
histogram with a non-positive lowest explicit bound gets a wrong number. Verified
at `internal/chsql/histogram_quantile.go`'s `boundLo` / `cumLo` Frags; a
different divergence with its own seed shape, so it is filed rather than folded
in here.

**#1584** (the two-sample `rate` floor evaluated over the merged group rather
than per source series) is *not* fixed here, and the interaction is worth stating
exactly rather than waving off. The floor itself is untouched: it remains the
`uniqExact(TimeUnix) >= MinSamples` HAVING at
`internal/chsql/range_bucket_fanout.go:179`, and that fix needs a
per-source-series pre-aggregation, a different mechanism.

What *does* move is the group the floor is evaluated over.
`lowerHistogramQuantileClassicAggRange` used to key the fanout with
`classicBucketAggGroupBy(userGroupBy, …)`, which appended `ExplicitBounds`; it now
passes `userGroupBy` unchanged, because keying on the layout is defect #1570.
Fanout groups are therefore strictly coarser, so the floor now counts distinct
timestamps over a superset. The direction is the same as this PR's: a series
whose layout changed mid-window used to be split into two fanout groups that
could each fall under the two-sample floor and vanish; now it is one group and
survives. #1584's actual complaint — that the floor is per *group* and not per
*source series* — is unchanged in both severity and mechanism.

## The guard invariant is pinned, and the pin was watched failing

Dropping `ExplicitBounds` as a grouping key is what turns `sum by (le)` keyless,
and `DropEmptyOnNoGroup` is what stops CH's 1-row-of-zeros for a keyless
aggregate from surfacing as a synthesised quantile. So this PR makes
"**the guard appears iff an Aggregate has zero grouping keys**" load-bearing —
and until now nothing enforced it. A later change that re-keyed or de-keyed one
of these aggregates would only show up as a `` `_cerb_n` `` appearing in or
vanishing from regenerated migration explain goldens, which reads as
"regenerate them", not as a semantic change.

`TestHistogramQuantile_EmptyGuardTracksGroupingKeyArity`
(`internal/promql/histogram_quantile_empty_test.go`) asserts the biconditional
structurally — over the plan and the emitted SQL, not over golden text. It
counts the Aggregates the emitter *would* guard (`len(GroupBy) == 0 &&
DropEmptyOnNoGroup`, the condition verbatim from
`internal/chsql/emit_node.go`) and requires the SQL to carry exactly that many
`` `_cerb_n` > 0 `` guards. The table spans both directions:

| case | collapse node | guards |
| --- | --- | --- |
| `sum by (le)` (instant) | `Aggregate`, keyless | 1 |
| `max by (le)` (instant) | `Aggregate`, keyless | 1 |
| `sum by (le, job)` (instant) | `Aggregate`, keyed by `job` | 0 |
| `sum by (le)` (range) | `RangeBucketFanout` | 0 |

Each case declares which collapse node it expects and the test fatals if the
plan used the other one, so a "0 guards" expectation can never pass against a
plan that never reached the shape under test — the vacuity mode #1569 itself
presented as (`Filter predicate=false`, zero rows, assertion trivially true).

The range row's rationale is worth stating precisely, because my first pass got
it wrong: the range path does **not** keep an anchor column as an `Aggregate`
grouping key. It has no `Aggregate` at all — it collapses through
`chplan.RangeBucketFanout` (`internal/promql/histogram_quantile_range.go:649`),
keyed by (series, anchor) structurally. `DropEmptyOnNoGroup` is an
Aggregate-only concept, so the guard cannot arise on that path. The test
deliberately does *not* read the fanout's own `GroupBy`, which is legitimately
empty for `sum by (le)` — that emptiness is #1584's subject.

### Mutation 1 — invert the emitter guard

`internal/chsql/emit_node.go:520`, `len(a.GroupBy) == 0` → `len(a.GroupBy) > 0`.
Three of four cases redden, **in both directions**: the two keyless rows lose a
guard they should have, and the keyed row gains one it should not.

```text
=== NAME  TestHistogramQuantile_EmptyGuardTracksGroupingKeyArity/instant_sum_by_le_job_keeps_the_user_label
    histogram_quantile_empty_test.go:347: emitted SQL for "histogram_quantile(0.9, sum by (le, job) (rate(http_server_request_duration_bucket[5m])))" carries 1 `_cerb_n` > 0 guard(s) but the plan has 0 keyless Aggregate(s) — guard presence must track grouping-key arity and nothing else (`job` survives the `le` drop, so the aggregate is keyed and CH emits no synthetic row)
=== NAME  TestHistogramQuantile_EmptyGuardTracksGroupingKeyArity/instant_sum_by_le_is_keyless
    histogram_quantile_empty_test.go:347: emitted SQL for "histogram_quantile(0.9, sum by (le) (rate(http_server_request_duration_bucket[5m])))" carries 0 `_cerb_n` > 0 guard(s) but the plan has 1 keyless Aggregate(s) — guard presence must track grouping-key arity and nothing else (`le` is dropped (the distribution lives in the parallel arrays) and the layout is no longer a key, so nothing remains)
=== NAME  TestHistogramQuantile_EmptyGuardTracksGroupingKeyArity/instant_max_by_le_is_keyless
    histogram_quantile_empty_test.go:347: emitted SQL for "histogram_quantile(0.9, max by (le) (rate(http_server_request_duration_bucket[5m])))" carries 0 `_cerb_n` > 0 guard(s) but the plan has 1 keyless Aggregate(s) — guard presence must track grouping-key arity and nothing else (the non-sum rung fold takes the same keyless shape as sum)
--- FAIL: TestHistogramQuantile_EmptyGuardTracksGroupingKeyArity (0.00s)
    --- FAIL: .../instant_sum_by_le_job_keeps_the_user_label (0.00s)
    --- FAIL: .../instant_sum_by_le_is_keyless (0.00s)
    --- PASS: .../range_sum_by_le_collapses_through_the_fanout (0.00s)
    --- FAIL: .../instant_max_by_le_is_keyless (0.00s)
FAIL	github.com/tsouza/cerberus/internal/promql	0.006s
```

(`SQL:` dumps elided; the assertion prints the full emitted query.)

### Mutation 2 — the range row is not decorative

The range row *passes* under mutation 1, which is correct — it has no Aggregate
to guard — but that would leave it indistinguishable from an inert assertion.
So I injected the exact regression it exists to catch: a keyless guarded
`Aggregate` under the fanout
(`internal/promql/histogram_quantile_range.go:650`,
`Input: rawSide` → `Input: &chplan.Aggregate{Input: rawSide, DropEmptyOnNoGroup: true}`),
which is precisely how a stray `` `_cerb_n` `` would appear in a range panel's
explain golden.

```text
=== CONT  TestHistogramQuantile_EmptyGuardTracksGroupingKeyArity/range_sum_by_le_collapses_through_the_fanout
    histogram_quantile_empty_test.go:330: plan for "histogram_quantile(0.9, sum by (le) (rate(http_server_request_duration_bucket[5m])))" carries 1 Aggregate node(s) alongside the RangeBucketFanout; the range path is expected to collapse structurally, so a keyless Aggregate appearing here would silently change what the guard count means
--- FAIL: TestHistogramQuantile_EmptyGuardTracksGroupingKeyArity (0.00s)
    --- FAIL: .../range_sum_by_le_collapses_through_the_fanout (0.00s)
FAIL	github.com/tsouza/cerberus/internal/promql	0.005s
```

### Mutation 3 — the control that matches the actual risk

Mutations 1 and 2 are emitter-side and plan-shape-side. The regression this test
actually exists to catch is *lowering*-side: a change that makes a keyed fold
keyless. So `internal/promql/histogram_quantile.go:713` was mutated to build the
bucket Aggregate with an empty `GroupBy`, which is exactly "a keyed aggregate
silently became keyless".

```text
    histogram_quantile_empty_test.go:338: plan for "histogram_quantile(0.9, sum by (le, job) (rate(http_server_request_duration_bucket[5m])))" has 1 guard-eligible (keyless + DropEmptyOnNoGroup) Aggregate(s), want 0 — `job` survives the `le` drop, so the aggregate is keyed and CH emits no synthetic row
--- FAIL: TestHistogramQuantile_EmptyGuardTracksGroupingKeyArity (0.00s)
    --- PASS: .../instant_sum_by_le_is_keyless (0.00s)
    --- FAIL: .../instant_sum_by_le_job_keeps_the_user_label (0.00s)
    --- PASS: .../range_sum_by_le_collapses_through_the_fanout (0.00s)
    --- PASS: .../instant_max_by_le_is_keyless (0.00s)
```

The point is what *else* reddens. Across `internal/promql` + `internal/chsql`,
exactly two things do:

```text
--- FAIL: TestHistogramQuantile_EmptyGuardTracksGroupingKeyArity
    --- FAIL: .../instant_sum_by_le_job_keeps_the_user_label
--- FAIL: TestLower
    --- FAIL: TestLower/histogram_quantile_classic_agg_plateau
    --- FAIL: TestLower/histogram_quantile_le_not_in_by
```

`TestLower` is a golden. It reddens the way goldens do — a diff — and the repair
anyone reaches for is `just update-golden`. The new test reddens naming the
query and the arity, so the same regression presents as a violated invariant
rather than as churn to re-baseline. That contrast is the argument for the test
existing at all.

Mutation 3 was found by the peer agent that regenerated the tier-0 goldens
(`cd8d5d27`); I re-ran it independently against this test rather than taking the
finding on trust, and the two-failure result above is my own run.

All three mutations were reverted; `git status` shows only the test file, and
`go test ./internal/promql/ ./internal/chsql/` is green on `75375c76`.

## Discipline

- No raw SQL. Every new shape is typed Frags / typed `chplan` exprs. The
  `sb.Write` / `writeSQL` / `strings.Builder` self-check over `internal/chsql/`
  surfaces only the known-allowed sites (`emit_node.go`'s pre-rendered subquery
  splice, `emit.go`'s output buffer); `internal/chsql/histogram_quantile.go`,
  the only chsql file this PR touches, surfaces nothing.
- No magic constants — `promGroupSampleValue`, the alias constants
  (`_hq_bounds_list` / `_hq_counts_list` / `_hq_ladder`), the lambda-parameter
  names, and `emptyAggGuardAlias` are all named.
- No `t.Skip`, no soft assertions, no tolerance files. The new unit assertions
  fail on `origin/main`'s shape (they pin `AggFuncs` names, the absence of a
  layout key in `GroupBy`, `DropEmptyOnNoGroup`, and the `` `_cerb_n` > 0 ``
  guard in the emitted SQL).


tsouza:
Pushed `cd8d5d27` — regenerates the Tier-0 migration explain goldens, which is what `lint` step 6 was red on:

```
Then each explain report matches committed golden byte byte   # then_explain.go:23   [FAILED]
```

**Why `just update-golden` didn't catch it.** It regenerates `test/spec/**` plus the two perf baselines, but not the migration explain reports — those live behind a separate `just migration-golden`. That gap is [#1573](https://github.com/tsouza/cerberus/issues/1573), and I'm chaining the two recipes in a companion PR so this stops being a hand-fix.

**What regenerated.** `just migration-golden` changed five `sql:` lines across four archetypes and nothing else:

| archetype | panel | expr |
|---|---|---|
| `already-otel` | dashboard `Database latency/A` | `histogram_quantile(0.99, sum by (le) (rate({__name__="db.client.operation.duration_bucket"}[5m])))` |
| `already-otel` | rule `otel:http_server_request_duration:p95` | `histogram_quantile(0.95, sum by (le) (rate({__name__="http.server.request.duration_bucket"}[5m])))` |
| `mimir-cortex` | rule `tenant:cortex_request_duration_seconds:p99` | `histogram_quantile(0.99, sum by (le, tenant) (rate(cortex_request_duration_seconds_bucket[5m])))` |
| `regulated-airgapped` | rule `audit:session_duration:p95` | `histogram_quantile(0.95, sum by (le) (rate(session_duration_seconds_bucket[5m])))` |
| `saas-repatriation` | rule `saas:checkout_latency:p95` | `histogram_quantile(0.95, sum by (le) (rate(checkout_latency_seconds_bucket[5m])))` |

Every one is a classic-histogram query. No non-histogram panel moved, and no field other than `sql:` moved on any panel.

**Audit — each delta traced to this PR's mechanism**, not blind-accepted. Token-count deltas per changed line:

- `sumForEach` 1 → 0 and `arrayCumSum` 15 → 0. The positional per-bucket-array reduction is gone, and `chplan.HistogramQuantile.BucketCountsCumulative` tells the emitter the array is already a ladder so it skips the prefix sum — exactly as the commit message describes.
- `groupArray` 0 → 2, new aliases `_hq_bounds_list` / `_hq_counts_list` / `_hq_ladder`. The `Aggregate` collects each contributing row's bounds and counts verbatim; the two new `Project` stages build the union bounds, fold the rungs, and apply the `ensureMonotonicAndIgnoreSmallDeltas` repair.
- `arrayFlatten` / `arraySort` / `arrayDistinct` / `has` / `arrayMax` appear (0 → 2/2/2/1/1): union-of-bounds construction, per-bound membership when evaluating a row's cumulative count at a union bound, and the prefix-maximum repair. `arraySum` 13 → 4 as the positional sums give way to the per-rung fold.
- `GROUP BY anchor_ts, `ExplicitBounds`` → `GROUP BY anchor_ts`. The bucket layout no longer enters series identity — the #1570 half of the fix, visible directly in the golden.
- On the three `sum by (le)` rules, `ExplicitBounds` was the *only* grouping key, so dropping it leaves the aggregate with none and the emitter adds the `DropEmptyOnNoGroup` guard: `count() AS `_cerb_n` … WHERE `_cerb_n` > 0`. This is the one delta that isn't purely the array fold, and it's a direct consequence of the grouping-key removal — PromQL `sum` over empty input yields no series, while a SQL aggregate with no `GROUP BY` yields one row.
  - Control: `mimir-cortex` groups by `(le, tenant)`, still has a real grouping key after `le` is dropped, keeps its `GROUP BY`, and correctly shows **no** `_cerb_n` guard. The guard appears exactly where the theory predicts and nowhere else.

**Nothing unexplained.** No golden changed for a reason outside this PR's scope.

`just migration-tier0` (assert mode, no `MIGRATION_UPDATE_GOLDENS`) is green locally: 8 scenarios, 58 steps, including `explain every corpus query offline`.

Issue #1598, opened 2026-08-02
fix(traceql): dotted attributes colliding with intrinsic names lower to the intrinsic column


tsouza:
Cerberus's TraceQL parser resolves an intrinsic for the **dotted** attribute form (`.name`, `.duration`, `.status`, …). Upstream Tempo deliberately does not: the dotted form is a user attribute lookup, and only the *bare* identifier is the intrinsic. Every dotted attribute whose key happens to collide with an intrinsic identifier is therefore silently routed to the wrong ClickHouse column.

## Evidence

**Upstream treats `.x` as a plain attribute, always.** `pkg/traceql/expr.y:499` (module cache `grafana/tempo@v1.5.1-0.20260508211128-2f74ea818de1`):

```
attributeField:
    DOT IDENTIFIER END_ATTRIBUTE  { $$ = NewAttribute($2) }
```

and `pkg/traceql/ast.go:1267` hard-codes the intrinsic away:

```go
func NewAttribute(att string) Attribute {
	return Attribute{
		Scope:     AttributeScopeNone,
		Parent:    false,
		Name:      att,
		Intrinsic: IntrinsicNone,   // <-- always
	}
}
```

Upstream reaches intrinsics only through the dedicated grammar productions (`expr.y:459` `NAME { NewIntrinsic(IntrinsicName) }`), i.e. for the bare `name` spelling.

**Cerberus resolves the intrinsic in the dotted path.** `internal/traceql/ast/parser.go:861-867`:

```go
func (c *cursor) parseAttributeField() Attribute {
	scope := c.advance().kind
	switch scope {
	case tokDot:
		name := c.expect(tokIdentifier).str
		c.expect(tokEndAttribute)
		return NewAttribute(name)
```

and `internal/traceql/ast/attribute.go:45-47`:

```go
func NewAttribute(name string) Attribute {
	return Attribute{Name: name, Intrinsic: intrinsicFromString(name)}
}
```

`intrinsicFromString` is exactly what upstream's `NewAttribute` refuses to call. Note that `tokDot` is the **only** callsite of `NewAttribute` in the parser (bare identifiers go through the intrinsic productions and `NewIntrinsic`), so this intrinsic resolution is never useful and always wrong.

[...]

PR #1599, opened 2026-08-02
fix(tempo): populate span name when the query references the name intrinsic

### LINKED ISSUE #1518: Tempo search: spanSets[].spans[].name is empty unless the query uses | select(name)
Reference Tempo populates the wire field `spanSets[].spans[].name` for any span matched by a condition on the `name` intrinsic — its fetch layer attaches the value and `asTraceSearchMetadata` lifts `IntrinsicName` into `Span.Name`. Cerberus fills it only when the query carries an explicit `| select(name)`.

**Evidence on `main`:**

- `internal/api/tempo/handler.go:1089` — `searchKeySelName = "__cerberus_sel_name"`, the only route by which a span's name reaches the response.
- `internal/api/tempo/handler.go:1284` — `SpanName → __cerberus_sel_name (populates SpanSetSpan.Name, …)`.
- `internal/api/tempo/handler.go:1881` — `observeSpan`, whose comment that "reference Tempo emits `name: \"\"` there" holds only for queries that never touch `name`.

**Measured divergence** (harness run against reference Tempo):

| query | tempo | cerberus |
| --- | --- | --- |
| `{ name = "GET /api/checkout/17" }` | `name="GET /api/checkout/17"` | `name=""` |
| `{ name != nil }` | names on all 5 matched spans | `""` on all 5 |
| `{ resource.service.name = "checkout" } && { status = ok }` | `""` | `""` (agrees) |

**Consequence.** Grafana's Tempo search results table renders the span name column from this field, so a search that filters *by name* comes back with a blank name column — the one query where the user has already told us which name they care about. It is a wrong wire value returned with a 200, not an error.

**Why no gate caught it.** The Tempo differ compares `spanSets` only when a corpus case sets `spss`, and no name-referencing case did. That is a corpus gap in its own right: the field is compared, and nothing that exercises the field opts into the comparison.

**Fix:** teach the sample-projection wrap to emit `__cerberus_sel_name` for any branch whose predicate references `SpanName` — per union arm, for set operations. This is an emitter + handler change with golden churn across the traceql search corpus. Add a corpus case with `spss` set and a `name` condition so the differ can see it.

Routed out of PR #1422 with the note that it "lands as its own PR; it is not parked or allow-listed here". Verified against `origin/main`; not addressed by any open PR.

tsouza:
## The divergence, verified against upstream

`/api/search` returned `spanSets[].spans[].name` as `""` for every query except `| select(name)`. Reference Tempo emits the span name for a much wider class.

The reference rule, read off the vendored snapshot (`grafana/tempo@v1.5.1-0.20260508211128-2f74ea818de1`) rather than inferred, has two halves:

1. **`pkg/traceql/engine.go` `asTraceSearchMetadata`** sets the wire field only when the `name` intrinsic is among the span's attributes:

   ```go
   atts := span.AllAttributes()
   if name, ok := atts[NewIntrinsic(IntrinsicName)]; ok {
       tempopbSpan.Name = name.EncodeToString(false)
   }
   ```

2. **`pkg/traceql/ast_conditions.go:3` `RootExpr.extractConditions`** decides what lands in `AllAttributes`, and it folds **every stage of the query into a single `FetchSpansRequest`**:

   ```go
   func (r RootExpr) extractConditions(request *FetchSpansRequest) {
       r.Pipeline.extractConditions(request)
       ...
   ```

Together: reference populates `name` **iff the query referenced the `name` intrinsic anywhere**, and the scope is query-wide rather than per-spanset-side.

That reconciles the apparent contradiction in the code being changed. `observeSpan`'s comment claimed reference emits `name: ""` on plain spanset-filter queries, and that claim is *true* — for the queries it was measured on, which never named the intrinsic. It was then generalised into a rule that also suppressed the name on queries that do name it. Both halves of the reference rule are load-bearing; having one without the other is a per-span `field_mismatch` in the differ, in one direction or the other.

Two consequences worth calling out, because they are the non-obvious part:

- **Either side of a structural operator counts.** `{ name = "a" } >> { .x = 1 }` returns descendant spans that were matched on `.x`, yet they carry `name` — conditions are collected query-wide, so the storage layer materialised the name column for every candidate span.
- **`span:name` counts.** Upstream's grammar folds the scoped spelling to the same intrinsic (`expr.y:480`, `SPAN_COLON NAME { NewIntrinsic(IntrinsicName) }`), and so does cerberus's parser (`parser.go:821`). `ScopedIntrinsicSpanName` exists as an enum value in both codebases but neither parser ever constructs it.

## The change

`ast.RootExpr.ReferencesIntrinsic(Intrinsic) bool` (new, `internal/traceql/ast/references.go`) walks the query for a named intrinsic. Its scope mirrors `extractConditions` deliberately — whole pipeline plus the metrics first stage, no per-side narrowing — and that reasoning is written down at the top of the file so the next reader doesn't "tighten" it back into a divergence. `MetricsSecondStage` is skipped explicitly: `MetricsFilter` / `TopKBottomK` / `ChainedSecondStage` carry no `Attribute` at all.

Threading it to the shaper needed no new mechanism and no SQL change:

- `traceqlLang.Parse` computes the bit — the only place the parsed AST is still in hand — and stashes it on `engine.Meta.Extra` beside the existing `metaKeySearchTraceLimit`, which is the established pattern for exactly this.
- `toTraceSummaries` / `ToTraceSummaries` take the `engine.Meta` of the result that produced the rows, and `observeSpan` falls back to `Sample.MetricName` when the bit is set and the `select(name)` slot is empty.
- **Both transports get it for free.** `SearchResult` (`structural_two_phase.go:102`) is the single shared HTTP+gRPC entry point and returns the `Meta` threaded through from `Parse`, including on the two-phase and empty-phase-A routes.
- **No projection change.** `canonicalSampleProjections` (`handler.go:1263`) already aliases the `SpanName` column to `MetricName` on every search projection, so the value was on the row the whole time and simply wasn't read.

`| select(name)` keeps precedence over the fallback: that value is what the projection actually selected. `/api/search/recent` builds its chplan directly with no TraceQL query, so its `Meta` carries no bit and its spans keep `name` unset — which is the correct reference shape for a query that names no intrinsic.

No TXTAR fixture: the change is response shaping only, so the emitted SQL and the `-- chplan --` IR are byte-identical. `test/spec/traceql/` snapshots would not move.

## Proof the tests fail without the fix

`TestSearch_SpanNamePopulatedWhenQueryNamesIntrinsic` drives all six cases through the real HTTP handler with an identical stub row, so only the query text differs and the assertion isolates the shaping decision. Reverting **only** the behavioural change in `observeSpan` (keeping the plumbing and both tests):

```text
--- FAIL: TestSearch_SpanNamePopulatedWhenQueryNamesIntrinsic (0.00s)
    --- FAIL: TestSearch_SpanNamePopulatedWhenQueryNamesIntrinsic/bare_name_equality (0.00s)
        search_span_name_intrinsic_test.go:139: spanSets[0].spans[0].name = "", want "GET /home" (the query names the intrinsic, so reference emits the span name)
    --- FAIL: TestSearch_SpanNamePopulatedWhenQueryNamesIntrinsic/scoped_span_name (0.00s)
        search_span_name_intrinsic_test.go:139: spanSets[0].spans[0].name = "", want "GET /home" (span:name is the same intrinsic (expr.y:480 folds it to IntrinsicName))
    --- FAIL: TestSearch_SpanNamePopulatedWhenQueryNamesIntrinsic/name_not_nil (0.00s)
        search_span_name_intrinsic_test.go:139: spanSets[0].spans[0].name = "", want "GET /home" (an existence condition on name still fetches it)
    --- FAIL: TestSearch_SpanNamePopulatedWhenQueryNamesIntrinsic/name_on_one_side_of_structural (0.00s)
        search_span_name_intrinsic_test.go:139: spanSets[0].spans[0].name = "", want "GET /home" (conditions are query-wide, so the returned descendants carry name too)
FAIL
FAIL	github.com/tsouza/cerberus/internal/api/tempo	0.019s
```

Exactly the four name-referencing cases fail. The two non-referencing cases — `{ resource.service.name = "frontend" }` and `{ status = error }` — pass **both before and after**, which is the half that matters for not trading one `field_mismatch` for its mirror image: the fix does not over-populate.

The test guards against a hollow pass by asserting the emitted SQL really does project `SpanName` as `MetricName`, rather than trusting the stub's hand-written row to be representative.

`internal/traceql/ast/references_test.go` unit-covers the walker across 27 parsed queries plus the hand-built AST shapes the parser never emits (value-form `SpansetFilter`, an ordinary attribute *spelled* `name` carrying `IntrinsicNone`). Those are new-code coverage rather than regression triggers — pre-fix the function doesn't exist, so "failing first" there is a compile error, not a signal.

## Found while verifying, filed separately

Cerberus's parser resolves an intrinsic for the **dotted** attribute form, which upstream deliberately never does — `expr.y:499` routes `.x` to a `NewAttribute` that hard-codes `Intrinsic: IntrinsicNone`, while cerberus's `NewAttribute` (`internal/traceql/ast/attribute.go:46`) calls `intrinsicFromString(name)`. Measured effect:

```text
{ .name = "x" }      => WHERE (`SpanName` = ?)          -- should be SpanAttributes['name']
{ .duration = "x" }  => WHERE (`Duration` = ?)          -- should be SpanAttributes['duration']
{ .status = "x" }    => WHERE (`StatusCode` = ?)        -- should be SpanAttributes['status']
{ .foo = "x" }       => WHERE (`SpanAttributes`[?] = ?) -- correct
```

Wrong answers, silently, for anyone with a span attribute whose key collides with an intrinsic name. Tracked in #1598 — deliberately not bundled, because it changes lowering and SQL goldens whereas this PR changes only response shaping, and the two need different validation. The shaper here is correct with respect to the AST it is handed; #1598 is that the parser hands it an AST where `.name` claims to be the intrinsic. When #1598 lands, `{ .name = "x" }` stops populating the wire field with no change needed here.

## Mutation coverage for the new walker

The walk folds its two-sided recursions with `||`, so a query naming the intrinsic on exactly **one** side is what separates a correct walk from a `&&` one. The original 27-query table had no such case for the nested-spanset, scalar-filter, scalar-arithmetic or metrics first-stage arms, and the `phase4-traceql-parser` leg went red on it. Reproduced locally with the leg's own flags, scoped to the new file:

```text
       LIVED INVERT_LOGICAL at references.go:52:41     (ScalarFilter LHS/RHS)
       LIVED INVERT_LOGICAL at references.go:74:42     (nested SpansetOperation)
 NOT COVERED INVERT_LOGICAL at references.go:76:41     (ScalarFilter as a spanset operand)
 NOT COVERED INVERT_LOGICAL at references.go:90:41     (ScalarOperation LHS/RHS)
       LIVED CONDITIONALS_NEGATION at references.go:126:34  (avg_over_time attribute)
       LIVED INVERT_LOGICAL at references.go:128:15    (compare with a nil filter)

Killed: 17, Lived: 4, Not covered: 2
Test efficacy: 80.95%   Mutator coverage: 91.30%
```

`references_mutation_test.go` adds exactly those shapes. After:

```text
Killed: 23, Lived: 0, Not covered: 0
Test efficacy: 100.00%   Mutator coverage: 100.00%
```

Both NOT-COVERED arms were checked for reachability before writing a test rather than assumed live: `ScalarFilter` satisfies `SpansetExpression` (`pipeline.go:121`) and `parseParenPipeline` returns it unwrapped when the parens hold no `|` chain, so `(max(name) > 1s) >> { .x = 1 }` reaches line 76 — the test asserts the parse really produced a `SpansetOperation` so it cannot silently drift onto the `Pipeline` arm again. Neither arm was dead code, so neither was deleted.

The second commit is the CI-side half. The two `./internal/traceql/ast` legs partition the package by `exclude_files` — the parser leg owns `parser.go` alone, the ast leg owns the rest — and a *new* file is caught by neither exclude, so `references.go` was being mutated by both. `TestMutationLegsPartitionEveryScopedFile` gates precisely that and was the `check-test` failure; adding `references` to the parser leg's exclusion alternation hands the file to the ast leg, matching the split's intent.

Closes #1518


tsouza:

PR #1644, opened 2026-08-03
fix(security): sanitize request-derived values before logging (CodeQL go/log-injection)

tsouza:
Closes CodeQL alerts #75-88 - 14 instances of go/log-injection across the prom, loki, and tempo handlers.

## What broke

Each site logged a request-supplied query string (or, for label_values, a path parameter) directly as a structured slog attribute:

    h.Logger.Debug("cerberus loki query", "logql", q, "sql", res.SQL, "args", res.Args)

A client-controlled value can contain a raw newline, carriage return, or other control character. Rendered through slog's TextHandler (or when a control byte reaches any downstream line-oriented consumer of the JSONHandler / otelslog-exported OTel log records internal/config/config.go can wire), that forges a fake log line or corrupts structured-log parsing - the same shape whether the query lands via /query, /query_range, /label/<name>/values, /index/stats, /index/volume, /patterns, /detected_fields, tempo search, or any of the tempo metrics_query_range paths.

## Fix

Added telemetry.SanitizeForLog (internal/telemetry/logsafe.go): replaces every unicode.IsControl rune with a readable Go-syntax escape, so the value stays legible for debugging instead of being dropped, with a fast path for the common case of a query with no control characters at all.

internal/telemetry was already an import every affected package either had or could take without introducing a new dependency edge, so one helper closes all 14 sites rather than each handler growing its own escaping logic.

## Verification

- go build ./... clean.
- go vet ./internal/api/... ./internal/telemetry/... clean.
- go test ./internal/api/{loki,prom,tempo}/... ./internal/telemetry/... - all pass.
- New internal/telemetry/logsafe_test.go: empty string, no-control fast path (byte-identical return, not just equal), newline/CR/tab, multiple control chars in one value, an unnamed control char falling back to a hex escape, and a C1 control char.
- golangci-lint run ./... clean (staticcheck flagged a raw control byte literal embedded directly in my first draft of the C1 test case; fixed to use the proper Go escape sequence in source instead).

### DIFF
--- a/internal/api/loki/detected_fields.go
+++ b/internal/api/loki/detected_fields.go
@@ -17,6 +17,7 @@ import (
 	"github.com/tsouza/cerberus/internal/chclient"
 	"github.com/tsouza/cerberus/internal/chsql"
 	"github.com/tsouza/cerberus/internal/schema"
+	"github.com/tsouza/cerberus/internal/telemetry"
 )
 
 // defaultDetectedFieldsLineLimit caps the number of log rows the
@@ -143,7 +144,7 @@ func (h *Handler) handleDetectedFields(w http.ResponseWriter, r *http.Request) {
 		h.respondError(w, &apiError{Kind: ErrInternal, Err: err, Status: http.StatusInternalServerError})
 		return
 	}
-	h.Logger.Debug("cerberus loki detected_fields", "logql", q, "sql", sqlStr, "args", args)
+	h.Logger.Debug("cerberus loki detected_fields", "logql", telemetry.SanitizeForLog(q), "sql", sqlStr, "args", args)
 
 	rows, err := h.Client.QueryDetectedFieldRows(r.Context(), sqlStr, args...)
 	if err != nil {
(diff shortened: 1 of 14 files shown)


tsouza:
Reverted 4 commits GitHub's CodeQL Autofix bot pushed onto this branch unprompted (`889f338e`, `42b77c53`, `6756312b`, `3f94c43d` — all authored as me, co-authored by `github-advanced-security[bot]`), back to my own `69751411`.

One of them (`889f338e`) broke the build outright: it deleted `logsafe.go`'s fast-path check and the now-"unused" `strings` import, but missed that `strings.Builder` a few lines below still needs it — `go build` failed with `undefined: strings`, which is what turned every check on this PR red.

The other three replaced `telemetry.SanitizeForLog(q)` at three call sites with an inline `strings.ReplaceAll(q, "\n", "").ReplaceAll(..., "\r", "")` — weaker than the shared helper (misses tabs and other C0/C1 control chars this PR's own tests cover) and duplicated per call site, which is exactly the "fix the class, not per-leg" anti-pattern this repo avoids elsewhere.

Restored and re-verified from a clean worktree: `go build ./...`, `go vet ./internal/api/... ./internal/telemetry/...`, and `go test` across all four affected packages — all clean. Force-pushed `69751411` back as the branch tip.

Issue #1568, opened 2026-08-01
Verify the .gitattributes -merge gate is honoured by GitHub's server-side merge


tsouza:
Spun out of #1567 (found while fixing the silent-blend class; deliberately **not** fixed there).

## Context

#1567 marks every generated baseline `-merge` in `.gitattributes` so git refuses to line-merge them and raises a conflict instead. That closes the class for merges performed by a **git client** — a local `git merge`/`git rebase`/`git pull` reads the repo's `.gitattributes` and falls back to the binary merge driver.

## The gap

Cerberus's actual merge paths are mostly **server-side**:

- the "Update branch" button (merge-commits) on a stale PR,
- `gh pr merge --squash`, which computes the merge on GitHub's side,
- GitHub's mergeability precomputation that decides whether a PR shows as conflict-free.

`-merge` is a *built-in* driver rather than a custom program, so there is no inherent reason GitHub cannot honour it — but I did not verify that it does, and the whole value of the gate rests on that answer. If GitHub's server-side merge ignores `.gitattributes`, a blend of the #1422 shape can still land through the merge button while the gate silently protects only local merges.

Note that the original #1422 corruption reached `main`, which is itself weak evidence about which path performed the blend.

## What would close this

1. Determine empirically whether GitHub's server-side merge honours `-merge` — e.g. on a scratch repo, two branches that each insert records at nearby offsets into a `-merge`-marked sorted JSON array, merged via the web UI / `gh pr merge`, and check whether it conflicts or blends.
2. If it **is** honoured: record that in `.gitattributes` / `docs/test-strategy.md` so the next reader does not have to re-derive it.
3. If it is **not** honoured: add a defence-in-depth CI check that does not depend on the merge implementation. The cheap version is a structural validator over the array-shaped baselines (uniform record shape, sorted, no duplicates, roster-internal consistency); the strong version re-runs the generator on the merge commit and diffs, which is what `perf-guards` already has the substrate for.

## Why it matters

A ratchet fed a silently blended baseline reports green while measuring nothing real, which is worse than having no ratchet. The gate in #1567 is the right mechanism; this issue is about confirming it actually fires on the path the repository merges through.

PR #1668, opened 2026-08-04
DO NOT MERGE: throwaway experiment for issue #1568 (.gitattributes -merge server-side verification)

tsouza:
Throwaway experiment for #1568. Both branches insert a new record at a different, non-overlapping line offset into the -merge-marked array test/perf/solver-decision-baseline.json. Checking whether GitHub's server-side merge honours the -merge attribute (refuses / conflicts) or falls back to a normal text merge (silently blends). Will be closed/deleted without merging into main.

### DIFF
--- a/test/perf/solver-decision-baseline.json
+++ b/test/perf/solver-decision-baseline.json
@@ -2520,6 +2520,16 @@
     "outer_range": "1h0m0s"
   },
   {
+    "query": "mmm_experiment_branch_b_record",
+    "routed": false,
+    "k": 0,
+    "reason": "not-sliceable",
+    "n_anchors": 10,
+    "fanout": 0,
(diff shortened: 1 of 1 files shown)


tsouza:
Experiment concluded. Empirical result for #1568: local `git merge` honours the .gitattributes `-merge` attribute and refuses (CONFLICT, binary merge driver) on this exact file/branch pair, while GitHub's server-side mergeability computation reports `mergeable: MERGEABLE` for the identical pair — i.e. it does NOT honour `-merge` and would silently blend the two non-overlapping record insertions via the "Update branch" / squash-merge / merge-commit paths. Closing without merging; branches deleted. Filed the fix in a follow-up PR.

Issue #1549, opened 2026-08-01
promql: unknown-name selectors still never reach the exp-histogram table, and a bare `{job="x"}` reaches neither histogram table


tsouza:
The regex-name histogram fan-out landed for the **classic** histogram table only. Two halves of the original gap remain, plus the doc comment that justified the whole deferral is now false.

**What landed**

`internal/promql/regex_histogram_lower.go` (new) builds the per-arm `UnionAll` lowering: `buildRegexMetricArm` + one `buildRegexHistogramCompanionArm` per companion suffix + `buildRegexHistogramBucketArm`, unioned in `lowerRegexHistogramSelector`. So `{__name__=~"http_.*"}` now reaches classic-histogram rows.

**Residue 1 — the exp-histogram table is still unreachable**

`internal/promql/regex_histogram_lower.go:57`:

```go
scan := &chplan.Scan{Table: s.HistogramTable}
```

`s.ExpHistogramTable` appears nowhere in the file. `lowerRegexHistogramSelector` builds arms for the classic histogram only, so a regex `__name__` matcher over an exponential histogram still returns HTTP 200 with zero series.

**Residue 2 — a selector with no `__name__` matcher at all reaches neither histogram table**

`internal/promql/lower.go:341` gates the new lowering on:

```go
if metricName == "" && hasUnpinnedMetricNameMatcher(v.LabelMatchers) && s.HistogramTable != "" {
	return lowerRegexHistogramSelector(v, s, ctx)
}
```

and `hasUnpinnedMetricNameMatcher` (`regex_histogram_lower.go:16-23`) returns true only when a `__name__` matcher is *present* with a non-`MatchEqual` type. A selector carrying **no** `__name__` matcher — `{job="api"}` — falls through to `internal/promql/lower.go:367` `tables := s.TablesForUnknownName()`, which returns `{GaugeTable, SumTable}`. Both histogram tables stay invisible for that shape.

**Residue 3 — the doc comment is now stale and asserts the opposite of the code**

`internal/schema/otel.go:583-588` still reads:

> The histogram / exp-histogram tables cannot join this candidate set: their row shape (Count / Sum / BucketCounts, no Value column) is disjoint from the Sample contract the `merge()` fan-out requires, so surfacing classic-histogram series under regex name matchers **needs the per-arm UnionAll lowering** (see the `_count` / `_sum` companion path) — **out of scope** for the unknown-name fan-out.

[...]

Issue #1758, opened 2026-08-05
compose-smoke/dashboard: showcase-traceql rootName panel returns 1 series, declared empty


tsouza:
## What's broken

The \`compose-smoke-shard (shard-kiosk)\` and \`dashboard-shard\` Playwright jobs in \`e2e.yml\` are failing on **push to main** (not just on feature branches), with:

\`\`\`
Error: dashboard=showcase-traceql panel="rootName (trace-scoped)" target=A expr="{ rootName = "GET /api/checkout" }" (declared=empty): declared-empty panel returned 1 series (the feature it showcases broke)
\`\`\`

This panel is provisioned to showcase a query that is expected to return **zero** series (a `declared-empty` panel), but it now returns 1 series — the trace-scoped `rootName` equality filter is over-matching.

## Evidence this predates PR #1746

Confirmed on `push: main` runs, independent of any single feature branch:

- Run `30989602789` (push, main, 2026-08-05T08:34:45Z) — job `compose-smoke-shard (shard-kiosk...)` (id `92252402844`) fails with the identical error.
- Run `30988146540` (push, main, 2026-08-05T08:13:28Z) — same failure class.
- The same error reproduces identically on unrelated PR branches (e.g. `fix/anchor-regex-label-matchers-1741`), which only touches `internal/chsql/builder.go`'s regex-anchoring (`=~`/`!~`) — it never touches the `rootName` equality (`=`) path.

So this is not caused by the regex-anchoring fix; it looks related to the recently-merged `048d7bb1 fix(traceql): compute trace-scoped root-identity filters and fix NOT-evaluation parity (#1740)`, which changed how trace-scoped root-identity intrinsics (`rootName` / `rootServiceName`) are lowered (`internal/traceql/lower.go`, `internal/chplan/metrics_compare.go`, `internal/traceql/metrics_compare.go`).

## Why it matters

- `compose-smoke` and `dashboard` are release gates (named in `release.yml`'s `RELEASE_REQUIRED_CHECKS`) — they aren't required PR branch-protection checks, but they DO block a release from publishing since they must post green on the commit being shipped.
- The showcase-traceql dashboard panel is provisioned production-facing content — a customer-visible query returning a spuriously non-empty "should be empty" result is a real correctness bug in trace-scoped `rootName` filtering, not a test artifact.

## Suggested next step

[...]

PR #1781, opened 2026-08-05
fix(promql): add exp-histogram companion arms to the regex __name__ path

tsouza:
## What was broken

An unpinned/regex `__name__` matcher over histogram-shaped metrics — `lowerRegexHistogramSelector` in `internal/promql/regex_histogram_lower.go` — unions a metric arm with classic-histogram `_count`/`_sum` companion arms (`buildRegexHistogramCompanionArm`) and a `_bucket` fan-out arm, but never built an arm reading `schema.Metrics.ExpHistogramTable`. The pinned-selector path (`expHistogramSelectorRouting` in `internal/promql/lower.go`) already routes a literal `<base>_count` / `<base>_sum` name to `ExpHistogramTable`, so a query like `{__name__=~".*latency_exp_hist.*"}` silently dropped exponential-histogram series that the equivalent literal-name query would return — a regex-vs-pinned inconsistency.

## Root cause

`internal/promql/regex_histogram_lower.go`, `lowerRegexHistogramSelector`: the per-arm `UnionAll` was built from `regexHistogramScan(s.HistogramTable, …)` and its two companion-arm helpers only ever scanned `s.HistogramTable`. No code path in that file ever referenced `s.ExpHistogramTable`.

## What changed

- Generalized `buildRegexHistogramCompanionArm` into a table-parameterized `buildRegexHistogramCompanionArmTable(table, …)`, kept as a thin wrapper for the classic-histogram case, and added `buildRegexExpHistogramCompanionArm` which scans `s.ExpHistogramTable` instead — consuming the same `_count`/`_sum` suffix set via `schema.Metrics.HistogramCompanionColumn`/`HistogramCompanionSuffixes()` the pinned path already uses, rather than re-deriving a fourth independent copy of the wire policy.
- `lowerRegexHistogramSelector` now appends the two new exp-histogram companion arms to the union whenever `s.ExpHistogramTable != ""`.
- No exp-histogram `_bucket` arm: an exp-histogram row has no `ExplicitBounds` ladder to fan (it carries `Scale`/`PositiveBucketCounts`/`NegativeBucketCounts` instead), so `<base>_exp_hist_bucket` is not wire-reachable either way — matching the pinned path, which never builds one.
- Updated the stale doc comment on `schema.Metrics.TablesForUnknownName` that used to call out this exact gap.

`WireArms` (the already-merged resolver from #1763/#1775) needed no changes — the name/scan/`le` split in this file already consumes it; this PR is purely additive (two new union arms).

## How it's proven

- Regenerated the affected TXTAR goldens via `just update-golden`: 6 existing regex-`__name__` fixtures bump `UnionAll arms=4` → `arms=6` (new `Scan(otel_metrics_exponential_histogram)` arms for `_count`/`_sum`).
- Added a new fixture, `test/spec/promql/regex_name_matcher_scans_exp_histogram_table.txtar`, that seeds real rows into both `otel_metrics_exponential_histogram` and a decoy `otel_metrics_histogram` row under the same base metric name, and asserts (via the chDB round-trip) that the query returns distinct `_count`/`_sum` rows from *each* table — proving the two arms don't clobber each other and that the SQL actually executes correctly against live data, not just structurally.
- `go test ./internal/promql/...`, `./test/rejection-parity/...`, `./test/surface-parity/...` all green.
- No new `fmt.Errorf`/`errors.New` rejection site is introduced, so no `test/rejection-parity/catalogue.json` entry is needed — `TestInventoryIsRegenerable` stayed clean with zero diff under `CERBERUS_UPDATE_INVENTORY=1`.
- Baseline churn is the expected, reviewed consequence of the two new union arms, not incidental drift:
  - `test/perf/cardinality-baseline.json`: `regex_name_matcher_scans_sum_table`'s `scan_rows`/`peak_intermediate` moves 1 → 5 (more union arms scanned), plus a new entry for the added fixture.
  - `test/perf/solver-decision-baseline.json`: new entry for the added fixture.
  - `test/e2e/migration/archetypes/kube-prometheus-stack/expected/explain.txt`: the `tables:` line for a regex-`__name__` migration-corpus query now lists `otel_metrics_exponential_histogram` alongside the existing three tables.

All three regenerated via `just update-golden`'s standard chain, per the documented workflow — no hand-editing.

Addresses #1549 (residue 1). Residue 2 is explicitly out of scope (funded but deferred per the maintainer decision recorded on that issue) and is not touched here.


tsouza:
Fixed the `probe` failure (run 31020022643, 12 tests, all `Code: 47 … Unknown expression identifier 'Count'`) in 70fc3e77.

**The non-obvious part:** this presented as `UNKNOWN_IDENTIFIER`, not `UNKNOWN_TABLE`, which is the tell that the exp-histogram table *existed* but had the wrong shape. chdb-go shares one catalog across the whole test process, and `handler_chdb_range_mode_test.go:1092`'s `expHistogramDDL` declares neither `Count` nor `Sum` — every other exp-hist fixture in the package does. So whichever shape that fixture left behind is what the next test read: the metadata-surface seeds (`metaShapedGauge/Sum/Histogram`) never created the table at all, and whether they saw a missing table or a Count-less leaked one was pure test-ordering luck. The companion arms this PR adds are correct — upstream OTel-CH really does have `Count UInt64` — so the defect was entirely fixture-side, but it was order-dependence rather than a simple omission.

Two changes, one per defect:

- **Fixtures** — added `metaShapedExpHistogramDDL` and folded all four tables into one `metaShapedMetricsDDL` bundle now used by every metadata-surface seed (16 call sites). Metadata selectors fan out across tables the query never names, so "which tables does this seed need?" is not answerable from the selector alone; seeding the full set removes the question.
- **Backfill registry (the class fix)** — `Count`/`Sum` joined `backfilledColumns` in both `internal/chclienttest/rewrite.go` and `test/spec/roundtrip_prep.go`, **scoped to the two histogram tables**. They are deliberately not injected into gauge/sum, where production has no such columns and a synthetic one would mask a query wrongly projecting a companion column off a table that cannot have one. `test/spec/backfill_columns_test.go` pins both directions (verified it fails if the scoping is dropped).

Also seeded a real exp-histogram row and asserted on it, so the new arm is pinned to *return data* rather than merely to parse — `_count`/`_sum`, no `_bucket` (no ExplicitBounds ladder to fan), and no bare base name. That legitimately widened the negated-regex expectation, which now covers both histogram layouts.

Verification: `go test -tags chdb -count=1 ./internal/api/prom/` green three times including `-shuffle=on` (the bug is order-dependent, so a narrow `-run` filter is not sufficient evidence); `./test/spec/...` green. The two bound/projection guards flagged as at-risk (`metadata_scan_bound_guard_chdb_test.go`, `metadata_distinct_projection_chdb_test.go`) pass unchanged — the fourth empty table did not widen any asserted scan bound.

tsouza:
The \`compose-smoke\` red on this PR is **#1758**, not this change: the failing assertion is \`showcase-traceql\` panel "rootName (trace-scoped)" returning 1 series where the dashboard declares empty. Same failure is on `main` (e2e.yml has been red on b7c376ad, dc083d71, 62389b07, c421cb50, 097a6fca), and this PR is PromQL-only — it touches `internal/api/prom` and the regex `__name__` fan-out, nothing in the Tempo head. `compose-smoke` is a release gate rather than one of the required 20, so it does not block this merge, but it does block publishing until #1758 lands.

The `probe` red **was** this change, and is fixed in 19ee087f: the regex `__name__` companion fan-out reaches both histogram tables for any selector at all, so five `drilldown-*` consumer-corpus replay entries began scanning `otel_metrics_exponential_histogram`, which the simplified chDB seed never created (UNKNOWN_TABLE). The seed now creates it with only the kind-specific columns; `Count` / `Sum` / `ResourceAttributes` / `ServiceName` come from `chclienttest.backfilledColumns`. Verified locally: `go test -tags chdb -run TestConsumerCorpus_Replay_ChDB ./test/consumer-corpus/` passes.

PR #2003, opened 2026-08-09
fix(spec): isolate each fixture's chDB session

### LINKED ISSUE #1987: test/spec: the chDB session is shared across fixtures, so a fixture's round-trip can depend on a table another fixture created
Found while enrolling the promql corpus against the real Prometheus engine
(parity wave, stacked on #1983). Unrelated to parity itself — it is a property of
the chDB round-trip lane that the enrolment work happened to surface.

## Evidence

`test/spec/promql/regex_name_matcher_underscored_matches_dotted.txtar` seeds only
`otel_metrics_gauge`. Its query `{__name__=~".*container_cpu_usage.*"}` lowers to a
UNION that also reads `otel_metrics_histogram` and
`otel_metrics_exponential_histogram`.

Run it alone:

```
go test -tags chdb -count=1 -run '^TestLower$/^regex_name_matcher_underscored_matches_dotted$' ./internal/promql/
```

```
Code: 60. DB::Exception: Unknown table expression identifier 'otel_metrics_histogram'
in scope SELECT Attributes, Count, MetricName, ResourceAttributes, ServiceName,
TimeUnix FROM otel_metrics_histogram. (UNKNOWN_TABLE)
```

Run the whole lane and the same fixture's round-trip passes.
`info_regex_name_matcher_unions_info_metrics` behaves identically. So the fixture
only executes because some earlier fixture created that table in the same session
and nothing dropped it.

## Root cause

`OpenChDB` calls `sql.Open("chdb", "")`, which is a process-wide in-process session
rather than a fresh database per fixture. `ApplySeed` compensates by rewriting
`CREATE TABLE` to `CREATE OR REPLACE TABLE` (`testsql.PromoteCreateTable`, whose
own comment names the shared session and `TABLE_ALREADY_EXISTS` on a second run).
That makes each fixture's OWN tables fresh — and does nothing at all for a table
the fixture never declares. Those persist, with the previous fixture's rows in
them.

## Why it matters

Two things, in increasing order of seriousness.

A fixture's executability depends on lane composition and ordering. Narrowing to
one test with `-run` — which invariant 5 requires when reproducing a red check —
gives a different result from the full lane, and the difference presents as a
ClickHouse error that looks like a lowering bug.

More seriously, the residue is DATA, not just schema. A fixture whose query unions
a table it did not seed reads whatever rows the previous fixture left there, and
its `expected_rows:` was regenerated by `just update-golden` in that same polluted
state. I did not find a fixture whose recorded answer is demonstrably wrong because
of this — in both cases above the residual rows do not match the query's selector,
so the answer is unaffected today. But nothing prevents it, and the golden layer
would absorb it silently if it happened, which is precisely the failure mode #1983
was built to close.

## Suggested shape of a fix

Give each fixture its own database and qualify the session to it, or drop every
table not declared by the current seed before executing. Either makes the lane's
result independent of ordering. Whatever is chosen, the check that it worked is
that every fixture passes when run alone under `-run '^TestLower$/^<name>$'`,
which is a property a test can assert directly over a sample of the corpus.

tsouza:
## Summary

- `test/spec/runner_chdb.go`'s `OpenChDB` shares one process-wide chDB engine across every fixture in a test binary. A fixture whose seed didn't declare every table its lowered query unions against could still pass, silently reading rows an earlier fixture's table left behind — so a fixture's own executability depended on which siblings happened to run before it, and the narrowed `-run '^TestName$/^subtest$'` reproduction CLAUDE.md invariant 5 requires could disagree with the full-lane result.
- `resetChDBSession` tears the native engine down (chdb-purego result Free + temp-dir removal) in `t.Cleanup` after every fixture, forcing the next `OpenChDB` call through chdb-go's own from-scratch session path. Cost is small: ~48ms for a fresh session vs ~16ms reusing a warm one, paid once per fixture.
- Isolating sessions surfaced two independently real, previously-masked bugs:
  - `internal/testsql/seed_ddl.go`'s `parseMetricsCreate`/`parseTracesCreate` searched for a `CREATE TABLE` statement's column-list parens in the **raw** seed text instead of the comment-stripped text, so a fixture whose seed comment contained a literal `(` could corrupt the column backfill. This had already shipped a genuinely broken `otel_metrics_gauge` table for `subquery_bare_aggregate_anchor_ts.txtar`, invisible for as long as the shared session let a sibling fixture's correctly-backfilled table answer the query instead.
  - 10 fixtures had queries that union across `otel_metrics_{gauge,sum,histogram,exponential_histogram}` but seeds that declared only some of those tables. Added the missing tables as empty `CREATE TABLE` statements, matching the existing `pi_constant.txtar` pattern for a query whose extra arms are intentionally empty.

## Test plan

- [x] Both `regex_name_matcher_underscored_matches_dotted` and `info_regex_name_matcher_unions_info_metrics` (the two fixtures #1987 named) now pass when run alone via `-run`, not just as part of the full lane.
- [x] Full `internal/promql` (336s) and `test/spec/promql` (242s) suites pass with zero failures.
- [x] `TestParityOracleImportsNoCerberusLowering`, `TestParitySectionsParse` pass.
- [x] `go build -tags chdb ./...` clean.

Closes #1987

### DIFF
--- a/internal/testsql/seed_ddl.go
+++ b/internal/testsql/seed_ddl.go
@@ -139,6 +139,10 @@ func BackfillTracesColumns(stmts []string) []string {
 // [parseMetricsCreate]).
 func parseTracesCreate(stmt string) (table string, colNames []string, rewritten string, ok bool) {
 	trimmed := stripLeadingNoise(stmt)
+	// See [parseMetricsCreate]'s identical prefix/trimmed split: every
+	// paren search below must run against trimmed, never stmt, or a `(`
+	// inside the stripped leading comment corrupts bodyCols.
+	prefix := stmt[:len(stmt)-len(trimmed)]
 	rest, ok := createTableTail(trimmed)
 	if !ok {
 		return "", nil, "", false
@@ -147,15 +151,15 @@ func parseTracesCreate(stmt string) (table string, colNames []string, rewritten 
 	if name != tracesTableName {
 		return "", nil, "", false
 	}
-	open := strings.IndexByte(stmt, '(')
+	open := strings.IndexByte(trimmed, '(')
 	if open < 0 {
 		return "", nil, "", false
 	}
-	closeParen := matchParen(stmt, open)
+	closeParen := matchParen(trimmed, open)
 	if closeParen < 0 {
 		return "", nil, "", false
 	}
-	bodyCols := stmt[open+1 : closeParen]
+	bodyCols := trimmed[open+1 : closeParen]
 	defs := splitTopLevelCommas(bodyCols)
 	declared := make(map[string]bool, len(defs))
 	names := make([]string, 0, len(defs))
@@ -187,7 +191,7 @@ func parseTracesCreate(stmt string) (table string, colNames []string, rewritten 
 	for _, ddl := range missing {
 		newDefs = append(newDefs, " "+ddl)
 	}
-	rewritten = stmt[:open+1] + strings.Join(newDefs, ",") + stmt[closeParen:]
+	rewritten = prefix + trimmed[:open+1] + strings.Join(newDefs, ",") + trimmed[closeParen:]
 	return name, names, rewritten, true
 }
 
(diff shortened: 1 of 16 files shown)


tsouza:
Added a follow-up commit: `test/spec/promql/fixture_isolation_chdb_test.go` pins the issue's own acceptance bar directly — a sample of 7 fixtures (the 2 named in the issue plus 5 more the investigation surfaced) each get run through a nested `go test -run '^TestRoundTripChDB$/^<name>$'` subprocess, so a future regression that reintroduces cross-fixture leakage fails this test even though the whole-package walk stays green.

Verified locally:
- `go build -tags chdb ./...` and `go vet -tags chdb ./...` clean.
- `go test -tags chdb -count=1 ./test/spec/promql/...` — 260s, all pass (includes the new regression test).
- `go test -tags chdb -count=1 ./internal/promql/...` — 336s, all pass.
- `go test -tags chdb,agpl_oracle,chdb_agpl_oracle -count=1 ./test/spec/logql/... ./internal/logql/...` — 70s/75s, all pass (no regression from the shared runner_chdb.go / testsql changes).
- `go test -tags chdb -count=1 ./test/spec/traceql/... ./internal/traceql/...` — 32s/32s, all pass.
- `go test ./internal/testsql/...` (untagged unit tests) — pass, confirms the parseMetricsCreate/parseTracesCreate paren fix didn't regress existing coverage.

PR #2141, opened 2026-08-14
test(parity): enrol classic histogram duplicate bounds

### LINKED ISSUE #2139: promql: classic histogram quantile mishandles duplicate bucket bounds
Found while advancing #1986 classic-histogram parity.

histogram_quantile_classic_duplicate_bounds produces a material mismatch against Prometheus after the parity input adapter derives the omitted Count and the fixture has a valid +Inf bucket. This is distinct from the final-bit rate-window class: it concerns duplicate-bound reconciliation in classic histogram quantile lowering/emission.

Done when the duplicate-bound answer matches Prometheus under exact parity and the fixture is enrolled in #1986.

tsouza:
Closes #2139

Enrol the classic histogram duplicate-bound quantile fixture in full exact Prometheus parity and ratchet the enrolment floor.

Verification: just spec-chdb.


tsouza:
Superseded by #2145, rebuilt cleanly from current main without the unrelated accumulated changes or unresolved generated-artifact conflicts.

Issue #1956, opened 2026-08-08
label_replace: a NULLABLE capture group sharing a name is still an unexpressible divergence


tsouza:
Successor to #1768. Tracks the residue of one divergence: a capture-group NAME
several groups share, where which group `$name` expands to turns on **which
group took part in the match** rather than on what any group captured.

## The gap

Go's `Regexp.expand` resolves `$name` by scanning `SubexpNames()` for the first
index `i` with `match[2*i] >= 0` — the first carrier that **took part**.
ClickHouse's `extractGroups` renders "did not take part" and "took part, matched
empty" identically, both as `''`, so participation is not directly observable
from SQL.

Upstream: **ClickHouse/ClickHouse#114733**. re2 carries the bit —
a non-participating group comes back with `data() == nullptr` — and
`src/Functions/extractGroups.cpp` discards it by calling `insertData(...)`
unconditionally into a hard-coded `Array(String)`. If that function grows a
nullable-returning form, this whole class collapses and the guard here can be
deleted rather than narrowed.

## Current boundary

The guard has been narrowed three times, each time against an exhaustive corpus
with an exact differential against Go's `ExpandString`.

- **#2116** narrowed it from "any nullable carrier" to real participation
  ambiguity, using three positional facts: a non-nullable carrier can never be
  the culprit; a carrier every match passes through truncates the search; and
  two carriers in different branches of one alternation never co-occur.
- **#2124** narrowed it again by rewriting the regex rather than reasoning about
  it: a nullable carrier with a non-nullable, always-consumed ancestor gets that
  ancestor wrapped in a synthetic probe group, whose emptiness reports the
  carrier's participation.
- **#2147** narrowed it a third time with a *negative* sibling probe: a carrier
  alone in a nullable alternation branch has nothing above it to probe, but if
  every other branch of that alternation is non-nullable, probing those siblings
  instead works — the carrier took part exactly when the alternation was
  entered and no sibling probe fired. This requires the alternation to be
  entered at most once per match (non-reenterable).

What is rejected today is a carrier alone in a nullable alternation branch where
the alternation itself is *repeated* (reenterable) — the negative-sibling-probe
mechanism (#2147) only proves a sibling took part at some point across the
repetitions, not whether this carrier took part on the pass Go's leftmost search
would have stopped at — or a carrier under a quest/star with no non-nullable
ancestor to probe (a shape #2124's mechanism cannot reach either).

```promql
[...]

PR #2147, opened 2026-08-14
fix(promql): probe mandatory sibling alternations

tsouza:
## Summary

Implements the negative sibling-probe mechanism for the safe duplicate-name label_replace case tracked by #1956. A nullable carrier in a mandatory, non-repeated alternation branch is selected when every non-empty sibling probe is empty.

The planner carries sibling conditions into LabelReplaceSegment; the ClickHouse emitter renders ordered multiIf selection. PromQL and LogQL retain a repeated-alternation divergence witness because sibling probes cannot identify a particular repetition.

## Verification

- Regenerated migration, parity, solver, promql, logql, traceql, chsql, optimizer, codegen, and cardinality artifacts with just update-golden.
- PromQL chDB fixture covers empty and non-empty duplicate captures.
- Added qlcommon, PromQL/LogQL lowering, and chDB emitter coverage.


tsouza:
Lane B claimed the failed required gremlins check. I will reproduce the reported mutation failure from its CI log, add the narrow regression coverage, and push a repair.

tsouza:
🔒 claimed by scheduler-lane-b

PR #2155, opened 2026-08-14
fix(promql): preserve classic quantile interpolation order

### LINKED ISSUE #2144: fix(promql): match classic histogram quantile interpolation for min-sample rates
## Problem

Exact Prometheus parity for classic `histogram_quantile(0.5, rate(...[1m]))` differs by one ULP in the min-sample fixtures after rate's extrapolation factor is divided before multiplication.

## Reproduction

Run `go test -tags chdb -run '^TestLower$/^(histogram_quantile_classic_agg_rate_min_samples|histogram_quantile_classic_bare_rate_min_samples)$' ./internal/promql`.

Prometheus returns `0.29999999999999993` where Cerberus returns `0.30000000000000004`; the bare-series fixture also has reciprocal final-bit differences around `0.06666666666666667`.

## Scope

Trace the remaining difference through classic histogram quantile interpolation and match Prometheus's floating-point operation order exactly. Do not add a tolerance or expected-failure exemption.

tsouza:
## Summary

Match Prometheus classic histogram quantile arithmetic by dividing the rank offset before multiplying by the bucket width.

Regenerate the dependent SQL and execution snapshots, including the rate min-sample fixture that exhibited the final-bit mismatch.

## Verification

- `go test -run '^TestHistogramQuantileValueFrag_DividesRankBeforeWidth$' ./internal/chsql`
- `just update-golden migration parity solver promql logql traceql chsql optimizer codegen cardinality`
- `node .github/scripts/forbid-sql-raw.mjs`

Fixes #2144.


tsouza:
Merged current main into this branch to include the compose crawl slice repair from #2154. CI has been re-triggered without force-pushing.

PR #2159, opened 2026-08-15
fix(traceql): implement the rest of the reference's static-validation pass

### LINKED ISSUE #2035: traceql: the parser implements one rule of the reference static-validation pass, not the rest
cerberus's in-house TraceQL parser reimplements the reference grammar and
its `impliedType` inference, but not the reference's static-validation pass.
#2034 added one rule of that pass — the binary-operand type match, which
closed #2033 — and left the rest. Each remaining rule is a query cerberus
**accepts** and reference Tempo **rejects** with 400, so this is a
wrong-acceptance divergence in every case.

## Verified

Each query below was run through `ast.Parse` on `main` + #2034 and is
accepted; the reference verdict is the rule that rejects it in
`pkg/traceql/ast_validate.go` (module cache copy of
`tsouza/tempo@v0.0.5-cerberus-accessors`).

| query | cerberus | reference rule |
|---|---|---|
| `{ } \| max(name) > "a"` | accepts | `Aggregate.validate` — "aggregate field expressions must resolve to a number type" (`ast_validate.go:121-124`) |
| `{ } \| avg(name)` | accepts | same rule |
| `{ .foo && 3 }` | accepts | `Operator.binaryTypesValid` — `TypeInt` is not valid under `&&` (`enum_operators.go:77-104`), "illegal operation for the given types" |
| `{ duration }` | accepts | `SpansetFilter.validate` — a span filter must resolve to a boolean (`ast_validate.go:156-159`) |
| `{ 3 }` | accepts | same rule |
| `{ -name = "a" }` | accepts | `Operator.unaryTypesValid` — unary `-` requires a numeric operand (`enum_operators.go:130-133`) |
| `{ span.x =~ "[" }` | accepts | `BinaryOperation.validate` — the regex is compiled at validation time (`ast_validate.go:222-245`), "invalid regex" |

The last one is the odd member: an uncompilable regex is not a type error,
but it is rejected in the same pass and for the same reason — the query
cannot match anything on any span, so it is a mistake in the text rather
than a query with an empty answer.

## Why this is not just tidiness

The rejection-parity catalogue (`test/rejection-parity/`) only pins
rejections cerberus *makes*; there is no layer that catches cerberus
**accepting** what the reference declines. `compatibility/tempo`'s corpus
cannot express the pairing either — `-- expect_status --` requires 4xx on
both sides, and a body diff requires 2xx on both — which the corpus itself
documents at `compatibility/tempo/driver/corpus/smoke.txtar:311-319`, where
exactly this shape (tempo 400, cerberus 200) had to be left out. So every
row above is invisible to CI today.

## Shape of the fix

Extend `internal/traceql/ast/validate.go` with the remaining rules, one
per commit so each rule's acceptance change is reviewable on its own. Two
of the reference's rules must NOT be adopted, because cerberus
deliberately answers queries the reference declines:

- `UnaryOperation.validate`'s `intrinsic=nil` rejection — `{ kind != nil }`
  is in cerberus's own corpus and lowers to a real predicate.
- `Attribute.validate`'s `parent` unsupported-error — `{ parent = "<hex>" }`
  lowers against `ParentSpanId`.

Those two are cerberus answering MORE, which is the safe direction; the
seven rows above are cerberus answering a query that cannot mean anything.

Found while fixing #2033; not covered by any open issue.

tsouza:
## Summary

- Implements the remaining rules of the reference TraceQL parser's static-validation pass (`internal/traceql/ast/validate.go`), which previously enforced only the binary-operand type-match rule from #2034.
- Each new rule closes a wrong-acceptance divergence documented in #2035: cerberus answered 200 with whatever the query happened to lower to, where reference Tempo answers 400.

New rules, mirroring the reference's own wording:

- `Operator.binaryTypesValid` — an operator's operand types must be valid for that operator (`.foo && 3`: `TypeInt` is not valid under `&&`).
- `SpansetFilter`'s boolean-result rule (`{ duration }`, `{ 3 }`).
- `Aggregate`'s number-result rule for `max`/`min`/`sum`/`avg` (`max(name)`).
- `Operator.unaryTypesValid` (`{ -name = "a" }`: unary `-` needs a numeric operand).
- Compile-time regex validation for `=~` / `!~` (`{ span.x =~ "[" }`).

Two of the reference's rules are deliberately **not** adopted, per the issue's own scope: the intrinsic-nil rejection (`{ kind != nil }` stays accepted) and the parent-attribute unsupported-error (`{ parent = "<hex>" }` stays accepted). The bare `parent` intrinsic is a cerberus-only structural marker with no reference equivalent, so `matchOperands` exempts it from the new operator-validity check to keep `{ parent && ... }` lowering unchanged.

## Incidental fixes required by the new rules

- `ParseIdentifier` wrapped a bare identifier in a synthetic `{ ... }` to reuse the grammar; that wrapper is never a real span filter. It now goes through a new `parseTree` helper that runs the grammar without `validate()`, instead of tripping the new boolean-result rule on `{ duration }`.
- `test/rejection-parity/catalogue/internal__traceql__lower.go.json`'s `lowerAttributeExpr#782c14d3` entry had trigger query `{ nestedSetParent }`, which the new span-filter boolean rule now catches one layer up, before it can reach the lowering-level guard it existed to exercise. Updated the (curated) trigger to `{ -nestedSetParent = 2 }`, which still reaches the same guard.

## Test plan

- [x] `go test ./internal/traceql/...` — new and existing tests pass, including new coverage for each of the 5 rules (acceptance + rejection ratchets) in `validate_test.go`.
- [x] `go test ./test/rejection-parity/... ./test/regression/... ./internal/api/tempo/... ./internal/migrateverify/... ./test/spec/... ./internal/solver/... ./internal/optcorpus/... ./internal/testsql/... ./test/surface-parity/...`
- [x] `go build ./...` and `go build -tags chdb ./...`
- [x] `node .github/scripts/forbid-skip.mjs` (`CHECK=all`), `node .github/scripts/forbid-sql-raw.mjs`
- [x] `gofumpt -l` / `goimports -l` clean on touched files

Closes #2035


tsouza:
🔒 claimed by scheduler-lane-a

tsouza:
Backing off: `fix/traceql-static-validation-2035` is already checked out in an active `scheduler-b-issue2035` worktree (git refuses a second worktree on the same branch) — this is scheduler-lane-b's own in-progress work. Leaving my claim marker in place only as a note that I looked; not actually working this. scheduler-lane-b should continue it directly.

PR #2178, opened 2026-08-15
fix(promql): fix native-histogram quantile domain order and NaN-Sum rank base

### LINKED ISSUE #2067: promql: empty native histogram answers NaN for out-of-domain phi where Prometheus answers -Inf/+Inf
## Summary

For a native histogram with no observations, cerberus answers
`histogram_quantile(phi, ...)` with `NaN` for every phi. Reference Prometheus
answers `-Inf` for `phi < 0` and `+Inf` for `phi > 1`, because it checks the
out-of-domain phi BEFORE the empty-histogram case.

Reference, `promql/quantile.go`'s `HistogramQuantile` (pinned fork
`github.com/tsouza/prometheus@v0.0.1-cerberus-parser`, lines 222-232):

```go
if q < 0 { return math.Inf(-1), nil }
if q > 1 { return math.Inf(+1), nil }
if h.Count == 0 || math.IsNaN(q) { return math.NaN(), nil }
```

Cerberus, `internal/chsql/histogram_quantile_native.go` (the `core`
expression):

```go
core := If(Eq(w.total(), InlineLit(0)), nan,
    If(Lt(w.phi(), InlineLit(0)), negInf,
        If(Gt(w.phi(), InlineLit(1)), posInf,
            ...
```

The `total = 0` guard is outermost, so it shadows both out-of-domain branches.

| input | reference Prometheus | cerberus |
| --- | --- | --- |
| empty histogram, `phi = -0.5` | `-Inf` | `NaN` |
| empty histogram, `phi = 1.5` | `+Inf` | `NaN` |
| empty histogram, `0 <= phi <= 1` | `NaN` | `NaN` (agrees) |

## Fix

Swap the nesting so the `phi < 0` / `phi > 1` tests are outermost and the
`total = 0` test sits directly above the `phi == 0` / `phi == 1` edges. The
oracle (`test/property/oracle/promql/histogram_native.go`
`nativeHistogramQuantileValue`) already orders these correctly and needs no
change.

## Why no existing layer catches it

`TestPromQL_Property_NativeHistogram`'s generator deliberately never draws an
empty histogram — every native-histogram function is a division by `Count` or a
walk over a zero-total array, so the empty case is its own semantic axis
(documented on `ExpHistogramDataset` in `test/property/gen/exp_histogram.go`).
No `test/spec/promql/` fixture seeds an all-zero exponential histogram either.

Worth adding both when this is fixed: an empty-histogram seed with an
out-of-range phi, ideally `-- parity --` annotated.

### LINKED ISSUE #2072: promql: native-histogram quantile ranks against the bucket sum, so NaN-observation histograms answer wrongly
## Summary

`histogram_quantile()` over a native (OTel exponential) histogram derives the
rank base from the **bucket arrays**, never from the histogram's stored
`Count`, and cannot see its `Sum` at all. For a histogram that observed NaN
values — where `Sum` is NaN and `Count` exceeds the sum of the buckets —
cerberus answers a plausible-looking number that is not the reference answer,
and in the upper tail returns a number where reference Prometheus returns NaN.

This is a *different* bug from #2066 (walk direction), which is why it was not
fixed there: #2066 is about which bucket the walk stops on, this is about what
the walk is ranking against.

## Reproduction

Verified directly against the pinned fork's own `promql.HistogramQuantile`
(`github.com/tsouza/prometheus@v0.0.1-cerberus-parser`). Positive buckets
`[2, 2]` at scale 0, offset 0 — bucket sum 4 — with two NaN observations, so
`Count = 6` and `Sum = NaN`:

| phi | reference Prometheus | cerberus |
| --- | --- | --- |
| 0.25 | 3.3635856610148585 | 1.4142135623730951 |
| 0.5 | 2.82842712474619 | 2 |
| 0.75 | **NaN** | 2.82842712474619 |
| 0.9 | **NaN** | 3.4822022531844965 |

The same histogram with `Count = 4, Sum = 10` (no NaN observations) agrees to
within the 1-5 ULP of #2024, confirming the divergence is the NaN-observation
contract and nothing else.

## Cause

`promql/quantile.go`'s `HistogramQuantile` treats a NaN `Sum` as the marker for
"this histogram counted observations that landed in no bucket", and three
behaviours hang off it:

1. `rank` is taken against `h.Count`, the STORED count, which is larger than
   the bucket sum.
2. The walk is forced FORWARD regardless of phi
   (`if math.IsNaN(h.Sum) || q < 0.5`), because a reverse walk cannot find a
   percentile whose mass is not in the buckets.
3. When the walk exhausts every bucket without reaching the rank, the result is
   NaN plus a `NativeHistogramQuantileNaNResultInfo` annotation; a histogram
   whose bucket total is below `h.Count` also earns a
   `NativeHistogramQuantileNaNSkewInfo` annotation.

cerberus models none of these. `internal/chsql/histogram_quantile_native.go`
computes `total = cum[length(cum)]` — the bucket sum — and
`chplan.HistogramQuantileNative` (`internal/chplan/histogram_quantile_native.go:58-82`)
carries no `SumColumn` and no `CountColumn`, so neither value reaches the
emitter. The columns themselves exist and are already used by other functions:
`internal/schema/otel.go:343-344` declares `CountColumn: "Count"` and
`SumColumn: "Sum"`, and `histogram_sum()` projects `argMax(Sum, TimeUnix)`.

## Scope

1. Add `SumColumn` and `CountColumn` to `chplan.HistogramQuantileNative` and
   set them at the four lowering sites
   (`internal/promql/histogram_quantile.go:1716`, `:1903`,
   `internal/promql/histogram_quantile_range.go:592`, `:709`), including the
   aggregation path, where both have to survive the scale-fold merge.
2. Rank against the stored `Count` and force the forward walk when `Sum` is
   NaN, alongside the phi-based direction choice `reverseWalkPhi` already
   selects.
3. Emit reference's NaN result when the walk exhausts the buckets below the
   rank.
4. Seed a `-- parity --` fixture on a NaN-`Sum` histogram so the reference
   engine arbitrates the shape, which today no fixture does.

The annotations in 3 are a separate surface from the value and can land with
whatever mechanism cerberus grows for Prometheus warnings; the wrong NUMBER is
what this issue is about.

tsouza:
## Summary

Both issues are the same bug class in the same function: `internal/chsql/histogram_quantile_native.go`'s `core` `If`-chain (the native-histogram-quantile value expression) checks the wrong things in the wrong order and ranks against the wrong quantity. Batching them avoids two PRs racing to edit the same lines.

- **#2067** — the domain checks (`phi < 0` → `-Inf`, `phi > 1` → `+Inf`) were nested *inside* the empty-histogram check, so an out-of-domain phi on an empty histogram answered `NaN` instead of `±Inf`. Reference Prometheus (`quantile.go:222-232`) checks domain first. Fixed by reordering the `If`-chain.
- **#2072** — the rank walk derived its base from the bucket arrays (`arrayCumSum(...)`) instead of the histogram's own stored `Count`, and never consulted `Sum`. For a histogram that observed NaN values (`Sum` is NaN, `Count` exceeds the bucket-array sum), cerberus answered a plausible-looking number where reference answers `NaN` in the upper tail. Fixed by:
  - Adding `CountColumn` / `SumColumn` to `chplan.HistogramQuantileNative`, threaded through all four lowering sites (bare instant/range, aggregated instant/range) — the aggregation path sums `Count`/`Sum` across the merge, mirroring how `ZeroCount` already does.
  - Ranking against the stored `Count` (`rankBase`) instead of the bucket-derived total.
  - Forcing the forward walk when `Sum` is NaN, regardless of phi (reference's `IsNaN(h.Sum) || q < 0.5`).
  - Answering `NaN` when the walk exhausts every bucket without reaching the rank (`idx = 0`, `arrayFirstIndex`'s not-found sentinel) — this is the NaN-observation case where `Count > bucket total`.

## Why one PR

Both fixes touch the *same* `core` chain in the *same* emitter function; fixing them separately would mean the second PR rebasing through the first's diff of the same lines. The scope note in #2072 itself says the emptiness/domain-order fix has to land alongside the rank-base fix for the `core` chain to read coherently.

## Test plan

- Two new fixtures pin the exact reproductions from each issue:
  - `test/spec/promql/histogram_quantile_native_empty_phi_below_domain.txtar` / `..._above_domain.txtar` — empty histogram, phi outside `[0,1]`, expect `-Inf` / `+Inf` (#2067).
  - `test/spec/promql/histogram_quantile_native_nan_sum_rank.txtar` (phi=0.25, answerable) / `..._nan_sum_exhausted.txtar` (phi=0.75, walk exhausts → `NaN`) — pin the exact reference values from #2072's own reproduction table (verified against the pinned fork's `promql.HistogramQuantile`).
- All 30 existing native-quantile fixtures' seed tables gained `Count`/`Sum` columns (required now that the columns are non-optional) with `Count` set to the exact pre-existing bucket-array total so their pinned answers are unchanged.
- `just update-golden migration parity solver promql logql traceql chsql optimizer codegen cardinality` regenerated every stale shard; reviewed the diff.
- Extended `internal/chsql/histogram_quantile_native_test.go`'s `TestEmit_HistogramQuantileNative_WalkDirection` and `gremlins_kill_test.go`'s `TestEmitHistogramQuantileNative_ComputedPhiNaNGuard` for the new `isNaN(Sum)` guard that now rides alongside the existing `isNaN(phi)` one.
- `go test ./...` green across the whole tree.

Closes #2067
Closes #2072


tsouza:

tsouza:

tsouza:

Issue #2172, opened 2026-08-15
parity oracle: spec-lane pre-wrap SQL hides TimeUnix for array-reduce subquery reducers


tsouza-squid:
Found while enrolling the "counter" (rate()/increase()/delta-temporality) bucket of #1986's parity-closure effort.

## What's blocked

`test/spec/promql/subquery_avg_over_time_rate.txtar` and `subquery_over_rate_range.txtar` (both `avg_over_time(rate(http_requests_total[1m])[2m:10s])` in range mode) cannot be enrolled against the Prometheus parity oracle. They're declared `-- parity_exempt --` (`reason: no-comparable-oracle`) instead, pending this fix.

## Root cause

In range mode, `avg_over_time`/`sum_over_time`/`count_over_time`-style array-reduce outer `RangeWindow` reducers emit raw SQL that projects `(Attributes, anchor_ts, Value)` — no `TimeUnix` alias (see the array-reduce path in `internal/chsql/range_window.go`, e.g. the `arrayAvg(window_vals)` emission). The sibling "selection" reducers (`max_over_time`, `min_over_time`) happen to alias `anchor_ts AS TimeUnix` in their own raw SQL as an incidental side effect of their `tupleElement`-based emission — that's why `subquery_max_over_time_rate_range.txtar` enrols cleanly while the array-reduce siblings don't.

This is **not a production bug**: `internal/api/prom/handler.go`'s `wrapWithSampleProjection` unconditionally renames `anchor_ts` to `TimeUnix` for every `chplan.IsDerivedShape` plan on the way out of the real HTTP handler, so `/api/v1/query_range` returns correct timestamps for these queries today.

It only blocks the **parity test harness**: `test/spec/parity_chdb.go`'s `RunParity` re-executes the spec lane's own pinned `-- sql --` section (the pre-`wrapWithSampleProjection` plan `internal/promql/lower_test.go` emits) to learn the projection's column names via `locateSampleColumns`. Because that raw SQL lacks the `TimeUnix` alias for this reducer family, and a range query's output timestamps participate in the comparison (see `comparesTimestamps`), `RunParity` cannot check output timestamps for these two fixtures at all.

## Suggested fix shape

Either:
[...]

Issue #2025, opened 2026-08-10
parity oracle: metric-name translation applies no OTel normalisation and synthesises no _sum/_count series


tsouza:
`labelsFromSeededRow` in `test/spec/parity_chdb.go` builds the reference engine's label set
from a seeded row. It maps `Attributes`, `ResourceAttributes` and `ServiceName` correctly
(#1997), but it takes `MetricName` **verbatim** and it synthesises no series a metric row does
not literally carry. Cerberus does both, so on those two axes the oracle is shown different
data and "disagrees" for a reason that has nothing to do with the query.

Four exponential-histogram fixtures were left out of #2022's enrolment for exactly this:

**No OTel-to-Prometheus name normalisation.** Cerberus normalises `container.cpu.usage` to
`container_cpu_usage` before matching, so `{__name__=~".*container_cpu_usage.*"}` selects it.
The oracle, handed the raw dotted name, matches nothing.

- `regex_name_matcher_underscored_matches_dotted` — reference returns 0 samples, cerberus 1
- `not_regex_name_matcher_excludes_dotted` — reference returns 2, cerberus 1 (the negated
  matcher fails to exclude the dotted name)

**No `_sum` / `_count` series over a native histogram.** Cerberus exposes a classic-style
`<name>_sum` / `<name>_count` for an exponential histogram — `histogram_mean_latency_exp`'s
own plan projects `toFloat64(Sum) AS Value` straight off `otel_metrics_exponential_histogram`
for the `_sum` selector. Vanilla Prometheus has no such affordance: a native histogram exposes
one histogram-valued series and you reach its sum with `histogram_sum()`.

- `histogram_mean_latency_exp` (`rate(latency_exp_hist_sum[1m]) / rate(latency_exp_hist_count[1m])`)
  — reference returns 0 samples, cerberus 1
- `regex_name_matcher_scans_exp_histogram_table` — reference 5 samples, cerberus 6

## Why this is delicate rather than a patch

Name normalisation **is part of the system under test**. Hand-copying cerberus's rule into the
oracle produces exactly the mirror `TestParityOracleImportsNoCerberusLowering` exists to
prevent: the two sides would agree about normalisation by construction and could never
disagree about it. The same argument applies to synthesising `_sum` / `_count`.

The honest framing is that both are **data-model** facts, like the label mapping in
`labelsFromSeededRow` — which is already written out independently in the oracle, with a doc
comment explaining that an independent expression of a documented mapping can still catch a
[...]

Issue #1985, opened 2026-08-09
promql: ClickHouse atan2 and Go math.Atan2 disagree by 1 ULP, so cerberus answers a different float than Prometheus


tsouza:
Found while enrolling the promql fixture corpus against the real Prometheus engine (parity wave, stacked on #1983). `test/spec/promql/binop_atan2_scalar_vector.txtar` is the one fixture in 225 whose disagreement is a genuine difference in the answer rather than a limitation of the oracle.

## Evidence

Query: `2 atan2 up`, seeded gauge `up`:

| labels | seed value | reference Prometheus | cerberus (ClickHouse) |
|---|---|---|---|
| `{job="api", instance="a"}` | 0.4 | `1.3734007669450157` | `1.373400766945016` |
| `{job="web", instance="b"}` | 0.7 | `1.2341215074081693` | `1.2341215074081695` |

The other two series in the same fixture agree exactly. The differing pairs are
adjacent doubles — a 1 ULP difference — so this is `atan2` itself, evaluated by
ClickHouse's libm on one side and Go's `math.Atan2` on the other, not a lowering
or a rounding step cerberus introduces.

## Why it matters despite being 1 ULP

Invariant 7 says every diff against a reference backend is a real bug to fix at
the source, and there is no tolerance file. That is the right posture here: the
value cerberus returns for `atan2` is not the value Prometheus returns, and a
Grafana panel comparing the two backends during a migration will show it. It also
means this fixture — and any future fixture over a transcendental function
(`atan2` is the one with an explicit two-argument libm call; `sin`/`cos`/`tan`/
`asin`/`acos`/`atan`/`exp`/`ln`/`log2`/`log10`/`sqrt` are all candidates) — cannot
be parity-enrolled, so the corpus is blind to real bugs in exactly those lowerings.

## What needs deciding

Whether cerberus should compute these functions the way Prometheus does rather
than the way ClickHouse does. That is a real trade (pushing the computation
client-side costs the pushdown), so this issue is the decision, not a foregone fix.
Whatever is decided, the outcome should end with the affected fixtures either
parity-enrolled or the divergence recorded as a bounded, named fact rather than an
un-enrolled fixture that looks the same as an unexamined one.

## Reproduce

Add to `test/spec/promql/binop_atan2_scalar_vector.txtar`:

```
-- parity --
oracle: prometheus
endpoint: /api/v1/query
scope: full
```

[...]

PR #2187, opened 2026-08-15
test(spec): close remaining promql parity-oracle coverage gaps

### LINKED ISSUE #1986: EPIC: the parity comparator only reads the canonical gauge Sample shape, so 331 promql fixtures cannot be enrolled
Found while enrolling the promql fixture corpus against the real Prometheus engine
(parity wave, stacked on #1983). That wave took enrolment from 1 fixture to 215.
This issue is the honest accounting of the remaining 331 and their single shared
root cause: `readSeededSeries` + `referenceShapeOfExpectedRows` in
`test/spec/parity_chdb.go` model exactly one shape — a gauge table read as
(MetricName, Attributes, TimeUnix, Value), compared against an `expected_rows:`
row of the same four columns. Everything below is a fixture that is correct, is
executable, and is invisible to its oracle because it is not that shape.

These are deliberately one issue and not five. The remedy in each case is the same
kind of work — teach the oracle's input model one more thing the corpus already
expresses — and splitting them would file four more leaves of one class.

## The classes, with counts measured over the 559-fixture corpus

**1. Counter fixtures — 152.** Seed `otel_metrics_sum`. The reference engine needs
`AggregationTemporality` / `IsMonotonic` to know what a `rate()` over these rows
means; `seededMetricsTables` reads neither. Adding the columns churns each
fixture's `SELECT *`-expanded golden, which is why the enrolment wave scoped them
out. Biggest single block, and the one covering `rate()`/`increase()`/`resets()` —
the lowerings most worth an oracle.

**2. Histogram fixtures — 100.** 50 seed `otel_metrics_histogram`, 49
`otel_metrics_exponential_histogram`, 1 both. `readSeededSeries` does not read
bucket columns at all, and `flatten` deliberately rejects a histogram-valued
reference answer rather than silently comparing its float field. Note
`ScopeExceptZeroBucket` already exists for the exp-histogram zero-threshold axis,
so the contract anticipated this.

**3. Non-Sample projections — 49.** 48 fixtures project two columns, one projects
three, and three more project four columns that are not the Sample shape
(`subquery_bare_aggregate_anchor_ts`, `subquery_over_sum_over_time`,
`subquery_time_minus_aggregate_offset` all emit `(Attributes, ts, ts, Value)`).
`referenceShapeOfExpectedRows` errors rather than skipping, which is the right
call — but it means the whole `*_over_time` two-column family is unreachable.

**4. Metric-name normalisation and resource-attribute promotion — at least 1
today, a cap on everything after.** `labelsFromAttributesJSON` builds the oracle's
label set from `Attributes` plus a raw `MetricName`. It ignores `ResourceAttributes`
and `ServiceName` entirely, and does not apply the OTel-to-Prometheus name
normalisation. `regex_name_matcher_underscored_matches_dotted` is the fixture that
exposes it: cerberus matches `{__name__=~".*container_cpu_usage.*"}` against the
seeded metric `container.cpu.usage` because it normalises the name first; the
oracle, handed the raw name, matches nothing and returns empty. Cerberus is right
and the oracle is wrong here — but the fix is delicate, because that normalisation
IS part of the system under test, and hand-copying it into the oracle is exactly
the mirror hazard `TestParityOracleImportsNoCerberusLowering` exists to prevent.
Needs a design, not a patch.

**5. No way to declare the evaluation window — 1 today.** `at_start_basic` is
lowered by `LowerAt(expr, schema, start, end)` over the fixture-wide
`[Unix(100), Unix(500)]` window so `@ start()` resolves to `Unix(100)`, but
`RunParity` passes `instantEval` for both bounds because the fixture carries no
`range_step:` section. The oracle therefore evaluates `up @ start()` as an instant
query in 2026, where `start()` is 2026, finds the only sample (1970) outside the
lookback, and returns empty. Cerberus is correct; the parity contract simply has
no key with which a fixture can say which window it is lowered over. Every future
`@ start()` / `@ end()` fixture inherits this.

## Not in scope here

Three fixtures (`absent_aggregate_no_series`, `absent_binary_no_series`,
`hour_default`) seed no rows at all, and `RunParity`'s guard correctly refuses
them — a reference engine with zero input series agrees with any answer. That is
the guard working, not a gap. Two more (`limitk_by_job_truncates`,
`limitk_computed`) disagree on which series they return, which is `limitk`'s
documented arbitrary-selection contract on both sides. Neither class needs work.
The `atan2` divergence is #1985.

## Definition of done

Each class above ends either with its fixtures parity-enrolled and the floor in
`TestParityEnrolmentFloor` ratcheted, or with a named, mechanically-checked reason
the shape cannot be compared — never with a fixture that is simply absent from the
enrolment count and indistinguishable from one nobody looked at.

tsouza:
## Summary

- Closes the parity-comparator scope gap `#1986` catalogued: the oracle's `readSeededSeries` /
  `referenceShapeOfExpectedRows` modeled exactly one shape (a gauge `(MetricName, Attributes,
  TimeUnix, Value)` read), so 331 correct, executable promql fixtures were invisible to
  `TestPromQLParityCoverageIsComplete` — indistinguishable from fixtures nobody had looked at.
- Every one of the five classes the epic named (counter, histogram, non-Sample projection,
  metric-name normalisation, evaluation-window declaration) now ends either parity-enrolled against
  real Prometheus, or carries a `-- parity_exempt --` section with a closed-vocabulary `reason:` and
  a reviewed `detail:` — never silently absent.
- `TestPromQLParityCoverageIsComplete` / `TestParityVocabulariesAreClosed` /
  `TestParityExemptVocabulariesAreClosed` all pass: promql fixtures are 592 parity-enrolled + 16
  mechanically exempt = 608/608 covered. `parityEnrolmentFloor` (promql + logql + traceql) moves to
  658.
- Enrollment uncovered and fixed four real production bugs along the way (below); none of the fixes
  were "just make the test pass" — each traces to a genuine SQL-emission or evaluation defect.

## Before / after

- promql: 485/575 enrolled at the epic's last accounting → **592/608 enrolled, 608/608 covered**
  (16 exempt) now. The corpus itself grew from 575 to 608 via unrelated merged PRs in the meantime;
  the 331-fixture gap this epic opened against is fully closed.
- Combined `parityEnrolmentFloor` (promql + logql + traceql): 658.

## Class-by-class

| Class | Status | Notes |
|---|---|---|
| Counter (`rate`/`increase`/DELTA-temporality) | Shipped | Oracle’s `AggregationTemporality`-aware DELTA-to-cumulative fold already existed; enrolled 22 fixtures (incl. 3 new `native_rate_temporality_{cumulative,delta,mixed}` dual-emit pins) against real Prometheus. The array-reduce `anchor_ts` SQL-shape gap affected two more fixtures; `#2192` fixed it and both are now parity-enrolled. |
| Histogram (classic + exponential bucket columns) | Shipped | Fixed a real bug in `internal/chsql/histogram_quantile.go`; enrolled the remaining classic/native bucket-interpolation fixtures against real Prometheus, with a ULP-tolerant oracle comparator for float rounding differences between ClickHouse and Go math (mirrors the already-closed `#1985` atan2 precision class). |
| Non-Sample projection | Shipped | Taught `RunParity`'s `locateSampleColumns` to recognise `anchor_ts` as the range answer's timestamp column when `TimeUnix` is absent — mirroring what `wrapWithSampleProjection` already does for real HTTP responses — and enrolled the subquery/topk/bottomk projection family this unblocked. |
| Metric-name normalisation / resource-attribute promotion | **Already shipped before this branch** | This is the class `#1986` flagged as needing "a design, not a patch" (`#2025`, closed 2026-08-14 by `#2133`): the oracle now applies OTel-to-Prometheus name normalisation and `ResourceAttributes`/`ServiceName` label promotion without importing cerberus's own lowering logic, respecting `TestParityOracleImportsNoCerberusLowering`. This branch enrolled the one fixture still missing a round-trip (`resource_attrs_allowlist_lookup_table`, needed a `seed:`/`expected_rows:` section it never had). |
| Evaluation-window declaration (`@ start()`/`@ end()`/pinned range steps) | **Already shipped before this branch** | The `at_start_basic` fixture the epic cited was already enrolled on `main` (`#1999`). This branch's enrollment work exercised the capability harder and found a real bug: `predict_linear(... @ ts, ...)` pinned over a range query wasn't evaluating at each output step. Fixed in `internal/chplan/range_window.go`, `internal/chsql/range_window.go`, `internal/promql/{modifiers,range_fns}.go`. |
| Exemption mechanism | Shipped | `test/spec/parity_exempt.go` defines the closed `reason:` vocabulary (`nondeterministic-selection`, `no-comparable-oracle`, `rejection-only`, `vacuous-empty-input`), mirroring `parity.go`'s own shape. New regression tests (`TestParityExemptSectionsParse`, `TestParityExemptVocabulariesAreClosed`, `TestPromQLParityCoverageIsComplete`) make a promql fixture with neither `parity:` nor `parity_exempt:` a hard failure. |

## Other production bugs fixed along the way

- **`NestMapWhere`** (`internal/testsql/query_rewrite.go`): a `Filter` fused directly onto a
  self-aliased `Map`-column projection collided with `RewriteMapProjections`'s `toJSONString(...)`
  wrap — ClickHouse resolves the `WHERE` identifier against the now-String-typed SELECT alias ahead
  of the FROM column, so the predicate received a `String` and the query failed outright. Fixed by
  pushing the `WHERE` one level below the wrap, mirroring `NestMapOrderBy`'s existing fix for the
  same collision on `ORDER BY`. Production never hit this (the live path has no `toJSONString` wrap);
  it only blocked enrolling `limit_ratio_half`/`limit_ratio_complement` against chDB.
- **`ReasonVacuousEmptyInput`** joins the closed exempt vocabulary: `absent_aggregate_no_series` and
  `absent_binary_no_series` read a selector against a seed with zero rows in any table — the vacuous-
  agreement shape `RunParity`'s own guard correctly refuses — and no existing reason named it.

## Test plan

- [x] `go test -run '^TestPromQLParityCoverageIsComplete$|^TestParityVocabulariesAreClosed$|^TestParityExemptVocabulariesAreClosed$' ./test/regression/...`
- [x] `go test -timeout 15m -race ./test/regression/... ./test/spec/...` (untagged lane)
- [x] `go test -tags chdb -count=1 ./test/spec/...` (`spec-chdb` recipe — the real chDB round-trip lane)
- [x] `go build ./...`
- [x] `gofumpt -l -extra` / `goimports -local github.com/tsouza/cerberus -l` clean on every file this
      effort touched
- [ ] CI green (lint / strict-scan / compat lanes need the real CI substrate per this repo's
      invariant 5)

## Notes for reviewers

Two structurally-scoped classes were already resolved by earlier, independently-merged PRs
(`#2133` for name-normalisation, `#1999` for eval-window) before this branch started — this PR's
job for those two was verifying the capability actually
covers the corpus and closing the one or two fixtures still missing a `parity:`/`parity_exempt:`
section, not re-doing the design work. Every fixture that is not parity-enrolled carries a
mechanically-checked `parity_exempt` reason. The former array-reduce `anchor_ts` SQL-shape gap
(`#2172`) was resolved by `#2192`, which this branch includes through its current-main sync.

Closes #1986


tsouza:

PR #2192, opened 2026-08-15
fix(chsql): alias anchor_ts AS TimeUnix for array-reduce subquery reducers

### LINKED ISSUE #2172: parity oracle: spec-lane pre-wrap SQL hides TimeUnix for array-reduce subquery reducers

tsouza:
## Summary

`avg_over_time`/`sum_over_time`/`count_over_time`-style array-reduce outer `RangeWindow` reducers emitted raw SQL projecting `(Attributes, anchor_ts, Value)` with no `TimeUnix` alias, while the sibling "selection" reducers (`max_over_time`, `min_over_time`) happened to alias it as an incidental side effect of their `tupleElement`-based emission. This blocked the parity test harness (`test/spec/parity_chdb.go`'s `RunParity`) from checking output timestamps for two fixtures, which sat `parity_exempt` (`reason: no-comparable-oracle`) instead of enrolled.

Not a production bug: `internal/api/prom/handler.go`'s `wrapWithSampleProjection` already unconditionally renames `anchor_ts` to `TimeUnix` for every derived-shape plan on the way out of the real HTTP handler.

## Fix

Consolidated the five duplicated "alias `anchor_ts` under `r.TimestampColumn` unless it's already `anchor_ts` itself" branches across `internal/chsql/range_window.go` into one `projectAnchorAsTimestampColumn` helper, which falls back to the canonical schema column name (`TimeUnix`) when `TimestampColumn` is itself the `anchor_ts` alias — the array-reduce-stacked-over-array-reduce subquery shape (`avg_over_time(rate(m[1m])[2m:10s])`). This mirrors `range_window_fused.go`'s `emitFusedMatrixSubquery`, which already had the correct fallback.

Re-enrolled `subquery_avg_over_time_rate.txtar` and `subquery_over_rate_range.txtar` against the Prometheus parity oracle, replacing their `parity_exempt` sections. `parityEnrolmentFloor` raised 606 → 608.

## Test plan

- [x] `go build ./...`, `go vet ./...`
- [x] `go test -tags chdb -run '^TestLower$/<each touched fixture>' ./internal/promql/...` — all 8 pass
- [x] `go test -run '^TestParityEnrolmentFloor$|^TestParitySectionsParse$|^TestParityVocabulariesAreClosed$' ./test/regression/...`
- [x] `just update-golden migration parity solver promql logql traceql chsql optimizer codegen cardinality` — clean, diff limited to the 9 expected files, no unrelated drift
- [x] `gofumpt -extra` / `goimports -local` clean on touched files

Closes #2172


tsouza:

PR #2231, opened 2026-08-16
fix(promql): support native histograms in delta and instant-rate functions

tsouza:
## Summary

- add native-histogram lowering for `delta`, `irate`, and `idelta`
- preserve reference reset, offset, `@`, instant-query, and range-query semantics
- add focused lowering, seeded chDB, wire, property, and compatibility coverage

## Root cause

The native-histogram range lowering only admitted `rate` and `increase`, so the three sibling functions fell through to the explicit exponential-histogram selector rejection even though reference Prometheus returns histogram-valued samples for them.

## Validation

Canonical ten-shard generation completed successfully, including all four PromQL legs and all eight cardinality legs. The generated parity catalogues were reclassified against current source evidence, the remaining public rejection was retargeted to #2225, and the append-only parity-enrolment floor was raised to 668.

Closes #2224

PR #2232, opened 2026-08-16
fix(promql): evaluate native histogram over-time reducers

### LINKED ISSUE #2225: promql: sum_over_time and avg_over_time reject native-histogram ranges
## Problem

Cerberus rejects native-histogram range vectors passed to histogram-aware over-time reducers, while reference Prometheus accepts them and returns histogram-valued samples.

Current main explicitly pins `sum_over_time(demo_latency_exp_hist[5m])` as a rejection. `avg_over_time` follows the same ordinary range-vector lowering path and must be audited alongside it.

## Evidence

The pinned Prometheus evaluator has dedicated histogram branches in both `funcSumOverTime` and `funcAvgOverTime`:

- histogram-only windows are folded with histogram Kahan addition;
- `avg_over_time` divides the merged histogram by the sample count;
- mixed float/histogram windows are removed with a mixed-sample warning;
- incompatible schemas and counter-reset-hint collisions are surfaced through annotations.

Cerberus already has histogram-aware window machinery for rate/increase and histogram-quantile idioms, but the top-level generic `sum_over_time` shape still reaches `expHistogramSelectorRouting` and rejects.

## Acceptance

- `sum_over_time()` and `avg_over_time()` return the reference histogram payload for histogram-only windows.
- Match reference handling of mixed samples, schema incompatibility, counter-reset-hint collisions, one-sample windows, offsets, `@`, and range-query anchoring.
- Pin the 13-column histogram row/wire contract.
- Add seeded execution and focused compatibility coverage.
- Retarget the shared rejection-parity guard to this issue when it becomes the oldest verified remaining trigger.

tsouza:
## Summary

- Implement native-histogram `sum_over_time` and `avg_over_time` as histogram-valued PromQL results.
- Reconcile mixed exponential-histogram scales before compensated component-wise folding while preserving one-sample, offset, pinned-anchor, and wire-shape behavior.
- Add focused lowering, upstream-reference, chDB HTTP, seeded TXTAR, solver, parity, and compatibility coverage.

## Root cause

The native-histogram root-shape dispatcher supported selectors, aggregations, rates, resets, and scalar-binop paths, but did not recognize the two histogram-aware over-time reducers. Their selectors therefore fell through to the explicit unsupported-shape rejection even though Prometheus evaluates them as native histograms.

## Verification

- [x] Rebased onto the #2231 merge on current `main`.
- [x] Regenerated `migration parity solver promql chsql codegen cardinality` canonically.
- [x] Focused lowering and upstream-reference tests pass.
- [x] Parity enrollment ratchet passes at the durable 671 floor.
- [x] Focused chDB HTTP histogram-valued test passes.
- [x] Compatibility corpus includes mixed-scale, offset, and pinned-anchor cases.
- [ ] Required CI is green.

Closes #2225


tsouza:
Ready checkpoint for head `978ed083b`:

- rebased onto current `main` after #2231 merged;
- resolved the overlapping native-histogram support/rejection messaging;
- regenerated every required shard canonically;
- replaced the obsolete issue-named 668 parity checkpoint with the durable capability-named 671 floor;
- passed the focused PromQL implementation/reference tests;
- passed the exact parity enrollment ratchet;
- passed the affected real chDB handler test in 200.7 seconds.

The PR is ready for review. Protected squash auto-merge is armed; CI will be handled fail-fast on the first real red.

tsouza:
Fail-fast CI repair pushed at head `4325930c1`.

Exact failure:

- `forbid-skip` → `doc-counts.mjs` found that canonical rejection-parity regeneration removed the resolved #2225 divergence, leaving 2 live divergence rows while `docs/coverage.md` still claimed 3.

Action taken:

- cancelled all eight still-running current-head workflows immediately;
- reproduced the exact `node .github/scripts/doc-counts.mjs` failure locally;
- updated the durable coverage count from 3 to 2;
- reran the exact gate successfully;
- committed and pushed only that focused repair.

Protected squash auto-merge remains the intended merge path; the new CI run will again be handled fail-fast.

tsouza:
Corrected head `7f28e8b95` is ready, mergeable, and protected squash auto-merge remains armed.

The first red was `compatibility/prometheus-floor`: all six new native-histogram over-time corpus cases passed, but their identities were absent from the committed parity roster. I canceled the remaining prior-head workflows, downloaded that run's authoritative `compat-cases.json`, and regenerated the Prometheus roster with the canonical `compat-baseline-sync.mjs` updater (952 passing cases, six deterministic buckets rewritten, zero stale files).

Exact local verification on the generated roster:

```text
compat-ratchet: compatibility/prometheus matches the committed baseline — all 952 case(s) ran and agreed with the reference, and the run introduced none the baseline does not record.
```

Fresh fail-fast CI monitoring is active on this head.

PR #2240, opened 2026-08-16
fix(promql): reconstruct DELTA first counter level

### LINKED ISSUE #2236: promql: DELTA zero-point clamp loses pre-window cumulative history
## Summary

The ordinary PromQL `rate()` / `increase()` window emitters use the first raw
OTel DELTA observation as the counter level for Prometheus's zero-point
extrapolation clamp. That value is one interval's increment, not the equivalent
cumulative counter level at the first in-window timestamp.

When the series has nonzero observations before `(T-range, T]`, Cerberus can
therefore shorten `duration_to_start` while the prefix-summed CUMULATIVE series
does not. A DELTA stream and its logically equivalent prefix-summed CUMULATIVE
stream then return different `rate()` / `increase()` values.

The fused subquery emitter already handles this distinction: for DELTA it
computes the prefix through the first in-window observation before calling the
shared extrapolation arithmetic. The ordinary instant and fanout-matrix paths
still pass the raw first DELTA value.

## Minimal counterexample

Evaluate at `T = 2026-01-01T00:00:01Z`:

```promql
rate(counter_total[1m])
```

Seed one DELTA-temporality series:

```sql
CREATE TABLE otel_metrics_sum (
    AggregationTemporality Int32,
    MetricName String,
    Attributes Map(String, String),
    TimeUnix DateTime64(9),
    Value Float64
) ENGINE = MergeTree ORDER BY (MetricName, Attributes, TimeUnix);

INSERT INTO otel_metrics_sum VALUES
    (1, 'counter_total', map(), toDateTime64('2025-12-31 23:59:00', 9), 90.0),
    (1, 'counter_total', map(), toDateTime64('2025-12-31 23:59:31', 9), 10.0),
    (1, 'counter_total', map(), toDateTime64('2026-01-01 00:00:00', 9), 40.0);
```

The query window is `(23:59:01, 00:00:01]`. Its raw DELTA observations are
`[10, 40]`, with:

- sampled interval: `29s`
- duration to start: `30s`
- duration to end: `1s`
- raw increase: `40`
- extrapolation threshold: `31.9s`, so neither edge is threshold-clamped

### Current ordinary-path result

The emitter supplies `first_val = 10`, so the counter-zero duration is
`29 * 10 / 40 = 7.25s`. It computes:

- `rate = 0.8563218390804598`
- `increase = 51.37931034482759`

### Independent expected result

The parity adapter converts the complete DELTA seed to the CUMULATIVE series
`[90, 100, 140]` before reference evaluation. The in-window counter readings
are therefore `[100, 140]`. The zero duration is `29 * 100 / 40 = 72.5s`, so
the `30s` start duration is retained. Reference semantics compute:

- `rate = 1.3793103448275863`
- `increase = 82.75862068965517`

The difference is solely the first value supplied to the zero-point clamp.

## Affected paths

- `internal/chsql/range_window.go`
  - ordinary instant `emitWindowedArrayExtrapolated`
  - ordinary fanout matrix `emitWindowedArrayExtrapolatedMatrix`
  - `firstValFrag` feeding `extrapolatedValueExpr`
- `internal/chsql/range_window_fused.go` is the behavioral control: its DELTA
  branch reconstructs the prefix through the first observation.
- `test/spec/parity_chdb.go` is the independent reference control: it
  prefix-sums complete DELTA seeds before Prometheus evaluation.

Existing ordinary DELTA fixtures do not expose the gap. Their first in-window
DELTA value is also the first cumulative value, or the only aged-out value is
zero. The fused-subquery fixtures do include prior nonzero observations and are
why that emitter already carries the prefix correction.

## Acceptance criteria

- A DELTA series and its prefix-summed CUMULATIVE twin produce equivalent
  `rate()` and `increase()` results when nonzero observations precede the query
  window.
- The contract holds for ordinary instant and fanout-matrix execution, while
  preserving the already-correct fused-subquery behavior.
- Live chDB plus independent Prometheus parity fixtures cover:
  - instant `rate()` with nonzero pre-window history;
  - instant `increase()` with nonzero pre-window history;
  - a range/fanout anchor where a nonzero observation has aged out.
- Boundary extrapolation coverage proves the result difference comes from the
  counter zero-point clamp, not the `1.1 * average_interval` edge-gap clamp.
- The fix preserves the fail-closed instant scan-bound invariant and does not
  reintroduce an unbounded retention-wide `groupArray`.

tsouza:
## Summary

- reconstruct the cumulative first counter level for ordinary OTel DELTA `rate()` and `increase()` evaluation
- keep instant `groupArray` scans exactly window-bounded while joining a separately aggregated scalar prefix
- carry a query-envelope-bounded running DELTA level through ordinary range fanout
- add instant rate, instant increase, and aged-out range-anchor Prometheus parity/chDB fixtures

## Root cause

The ordinary emitters passed the first raw DELTA increment to Prometheus's counter-zero extrapolation clamp. Once earlier observations aged out of the range, that increment was smaller than the equivalent cumulative counter level, so `duration_to_start` was shortened incorrectly. The fused subquery emitter already reconstructed the cumulative level.

## Impact

DELTA streams now agree with their prefix-summed CUMULATIVE equivalents across ordinary instant and range evaluation without materializing retention-wide sample arrays or weakening the fail-closed instant scan-bound contract.

## Validation

- canonical migration, parity, solver, PromQL, LogQL, TraceQL, chsql, optimizer, and codegen generation stages passed
- pre-push repository hygiene, forbid-skip, forbid-SQL-raw, feature-discipline, and actionlint gates passed

Closes #2236


tsouza:
Focused acceptance is green at `39d8edbc6`:

- live chDB: `increase_delta_temporality`, `rate_delta_temporality`, and `rate_delta_temporality_range_prefix`
- post-optimizer live chDB plus real Prometheus parity: the same three fixtures
- instant coverage pins the `90 / 10 / 40` pre-window-history counterexample
- range coverage pins the nonzero observation aging out on a left-open anchor
- both shapes keep the start gap below `1.1 * average_interval`, isolating the counter-zero clamp

The implementation is complete; marking ready and arming protected squash auto-merge while the remaining required checks run.

tsouza:
Fail-fast CI update at `ca132e5e1`:

- first red was `lint`: only `funlen` on the two ordinary extrapolation emitters (`88 > 80` and `90 > 80` statements)
- all remaining previous-head workflows were canceled immediately
- extracted the instant scalar-prefix join and bounded matrix running-level construction into stable helpers; emitted SQL is unchanged
- re-ran the exact three live chDB fixtures and the same post-optimizer real-Prometheus parity fixtures: all pass
- pre-push policy gates pass

CI has restarted on the corrected head; protected squash auto-merge remains the target.

tsouza:
Fixed the corrected-head `check-test` failure by adding the append-only parity enrolment checkpoint for the new DELTA prefix fixture count (670).

Focused reproduction now passes:

```text
GOCACHE=/tmp/cerberus-2236-gocache go test -count=1 -run '^TestParityEnrolmentFloor$' ./test/regression/
```

Pushed as `5c0f94c13`; auto-merge remains armed. Resuming fail-fast CI monitoring.

tsouza:
Fixed the first corrected-head red (`perf-guards-shard (2)`) at `dea0d8351`.

Root cause: DELTA prefix reconstruction's running-sum window was applied to every temporality-bearing matrix, including CUMULATIVE-only series. The fan-out lowerer now partitions CUMULATIVE and DELTA inputs before emission, preserving the historical cheap cumulative path while keeping prefix reconstruction isolated to DELTA rows.

Local reproduction and verification:

- Before fix, same-shard `TestScaleWallPin` failed 3/3 at 16.40x, 16.04x, and 14.45x (bound 12x; scan amplification 4.63x).
- After fix, it passed 3/3 at 9.22x, 8.34x, and 9.69x with 1.00x scan amplification.
- Final post-regeneration run passed at 10.94x with 1.00x scan amplification.
- `TestCardinalityBaselineCoversTheCorpus` passes after canonical regeneration added the new range-prefix row.
- Focused fan-out/native temporality partition tests pass.
- Canonical generated shards completed successfully: migration, parity, solver, PromQL, LogQL, TraceQL, chsql, optimizer, codegen, and cardinality.

All other current-head workflows were canceled immediately on the first red. Resuming fail-fast CI on this verified head; protected squash auto-merge remains the intended merge path.

tsouza:
Fixed the first rebased-head red (`agpl-oracle`) at `04ef2fb68`.

Root cause: the COW guard treated a freshly re-anchored `UnionAll` as the single off-spine head. The production slicer remained correct: each union arm still shared its immutable off-spine subtree across shards. The guard now descends union arms and verifies every positional shared head.

Focused verification:

- the exact two AGPL/race failures now pass
- `TestCardinalityBaselineCoversTheCorpus` passes
- perf shard 2 `TestScaleWallPin` passes at 1.00x scan amplification and 11.46x wall ratio
- canonical generated output remained unchanged after the clean rebase

CI is running on the verified head; protected squash auto-merge is armed.

PR #2251, opened 2026-08-16
ci: add evidence-backed shadow merge fence

tsouza:
## Summary

- add a shadow `merge-fence` that recomputes impact selection from trusted changed paths
- require invocation-bound, lane-native evidence with exact SHA, tree, run, attempt, artifact, and nonce correlation
- publish closed native evidence bundles from every enrolled producer workflow
- scope pull-request mutation work to changed production lines with a fail-closed full fallback
- isolate replaceable main and scheduled deep runs without cancelling queue or qualification work
- preserve `quickstart` as its own direct required check and as independent native evidence

## Safety properties

- rollout remains shadow-only; this PR does not change branch protection
- unknown paths, selector errors, absent evidence, duplicate evidence, non-success conclusions, and hollow reports fail closed
- producer identity and evidence digests are independently revalidated by the collector
- projected-trunk and pull-request identities are handled explicitly rather than conflated

## Validation

- 151 focused Node tests across the lane contract, selector, collector, native evidence, concurrency, and mutation suites
- targeted structural Go regressions for merge-queue posting, cancellation safety, and mutation timeout wiring
- `git diff --check`

Part of #2230

Issue #2230, opened 2026-08-16
ci: implement the evidence-backed merge and release test fences


tsouza:
## Objective

Replace Cerberus broad, slow pull-request matrix with an evidence-backed test fence that keeps projected trunk buildable within a p95 20-minute merge SLO, while requiring every finite deterministic source and artifact qualification lane to pass before publication.

## Evidence baseline (2026-08-16)

Across the latest 50 merged final SHAs:

- required-ready p50 was 41.4 minutes and p95 was 59.0 minutes;
- runner usage was p50 190 and p95 310.8 minutes per PR;
- performance, chDB/roundtrip, mutation, compatibility, and E2E/Compose dominated cost;
- ordinary-PR `property` and `coverage` contexts did not execute the work their names implied;
- Compose crawl produced unique failures while non-blocking;
- recent main pushes exposed property, E2E, coverage, and generated-artifact failures after merge.

## Accepted constraints

- Preserve all 14 test layers and independent oracle diversity.
- Merge fence: fixed fast core plus fail-closed impact selection.
- Required contexts converge to `merge-fence`, `quickstart`, `CodeQL`, `pr-body`, and `forbid-deferral`.
- `quickstart` remains a direct required context: every non-documentation PR and projected-trunk queue entry must execute the repository-root README Compose canary, while every push to main and maintenance branches must execute it regardless of path.
- Merge queue validates the exact projected trunk; no admin bypass.
- Unknown paths or selector failures run the full relevant suite.
- Release qualification independently requires fresh `e2e.quickstart/default/source_tree` evidence; another aggregate cannot substitute for either the direct merge check or release evidence.
- Release qualification binds an immutable source SHA/tree and candidate digest, finishes within 120 minutes, and rejects missing, skipped, neutral, cancelled, mismatched, zero-test, or non-executed evidence.
- Publication starts only after source and artifact qualification succeeds. Public-availability checks remain post-public monitors.

## Implementation checklist

- [x] Add the machine-readable lane registry and closed selection/report contract with anti-hollow-green tests.
- [x] Add enrollment guards for tagged tests, mutation packages, property shapes, and per-head live parity.
- [ ] Build `merge-fence` in shadow mode with trustworthy lane-native telemetry.
- [x] Add the direct required one-stack Compose quickstart canary.
- [ ] Add impact routing, consistent chDB caching, split lint/strict seams, and changed-line mutation.
[...]

PR #2257, opened 2026-08-16
test(ci): backtest the merge-fence lane selector against real PR history

tsouza:
## Summary

- Issue #2230's acceptance checklist has an unchecked live "run the shadow comparison over at least 7 days and 50 non-documentation final SHAs" soak. The new selector (`ci-lane-selection.mjs`, PR #2251) is a pure function of `(base SHA, head SHA, registry)` — no live checkout required to evaluate — so this adds `.github/scripts/ci-lane-backtest.mjs`, which replays it against real historical merged PRs pulled from the Actions/PR REST API instead of waiting on the clock.
- Per PR it computes: whether every lane the OLD (always-run) system observed failing was still selected by the NEW selector (a miss is a real "old-only catch", reported not hidden); the NEW-selected lanes' summed actual runner-minutes; and a wall-clock "required-ready" proxy (earliest job start to latest completion among NEW-selected lanes, assuming unconstrained parallelism).
- `contextMatches()`/`jobObservationState()` are reused from `ci-lane-shadow-collector.mjs` (the latter now exports `contextMatches`) so "which job backs this lane" can never drift between the live collector and this replay; the Actions-run discovery itself is intentionally NOT reused — the live collector's strict PR-association binding exists to stop a spoofed live event, which doesn't apply to an already-merged SHA looked up by its own unique head SHA.
- `ci-lane-backtest.test.mjs` covers the comparison/aggregation logic offline with synthetic fixtures (`node --test`, no network).

**Depends on #2251** — this branch was cut from #2251's current state and imports `selectionPolicy`/`ci-lane-contract.mjs` from it, so it can't merge before #2251 does. Not intended to merge until then.

## What it found, run for real against `tsouza/cerberus`'s actual history

| Sample | Non-doc SHAs | Span | Required-ready p50/p95 (target p95 ≤20min) | Runner-min p50/p95 (target p50 ≤95/p95 ≤160) | Coverage |
| --- | --- | --- | --- | --- | --- |
| n=50 | 50 | 1.3 days | 39.0 / 54.6 — **FAIL** | 82.5 / 101.0 — PASS | 0 old-only catches — PASS |
| n=150 | 150 | 7.1 days | 26.6 / 53.3 — **FAIL** | 82.5 / 100.8 — PASS | 0 old-only catches — PASS |

Coverage and runner-cost both clear their bar at both sample sizes (the n=150 run alone satisfies the literal "≥50 SHAs over ≥7 days" wording). Required-ready wall time does not, and the tool's own diagnostics point to two concrete, unfixed causes — reported here per the task's instruction not to silently patch the registry to paper over what the backtest finds:

1. **The selector falls back to the full lane set on ~80% of real PRs** (40/50 in the n=50 sample) because whole directories match no conditional lane's `package_globs` and no `known_nonimpact_glob`, so `unknown_paths` fires and every conditional lane gets selected — same cost as the old always-run system. Top offenders by occurrence count in the n=50 sample: `test/spec` (1819), `compatibility/parity-baseline` (263), `test/property` (30), `.github/scripts` (28), `.github/workflows` (27).
2. **`performance.benchmark` alone runs a 17.4min median / 27.3min max** across the n=150 sample — selected on half of real PRs, it alone leaves little headroom under the 20-minute SLO before any other lane's start-skew is added. Even the 18% of PRs where impact selection successfully narrows to precisely the touched lanes (no fallback) still miss the target (wall-clock p50 25min/p95 52min) — narrowing conditional lanes barely moves this number because the ~27 `always`-posture lanes (and `performance.benchmark` when selected) dominate the critical path regardless.

## Out of scope (needs a live run, not a backtest)

Whether the new artifact-bundling / evidence-collection plumbing (`ci-lane-shadow-collector.mjs`, `ci-lane-native-evidence.mjs`) actually works end-to-end inside a live GitHub Actions run — permissions, cross-job artifact download, etc. A backtest cannot exercise the runner sandbox; that still needs one short live-verification pass once #2251 lands.

## Test plan

- [x] `node --check` on every new/changed `.mjs` file
- [x] `node --test .github/scripts/ci-lane-backtest.test.mjs` — 18/18 passing, offline, synthetic fixtures only
- [x] `node --test .github/scripts/ci-lane-shadow-collector.test.mjs` — 25/25 still passing (regression check on the `contextMatches` export)
- [x] `CHECK=all node .github/scripts/forbid-skip.mjs` clean
- [x] Ran the real tool twice against `tsouza/cerberus` (n=50 and n=150) — results above; full JSON reports available on request (written under the gitignored `build/` tree, not committed)
- [ ] `just lint` / `just test` — not run locally per CLAUDE.md invariant 5 (no speculative pre-flight); this is a Node-only script change with no Go surface

## Notes for reviewers

- The `RUNNER_P50_TARGET_MINUTES` (95) / `RUNNER_P95_TARGET_MINUTES` (160) constants have no other home in the source tree (they're issue #2230's literal acceptance numbers); `MERGE_P95_SLO_MINUTES` (20) is reused from `ci-lane-contract.mjs` rather than duplicated.
- The wall-clock "required-ready" figure is an explicitly-documented optimistic proxy (unconstrained parallelism) — the real number could be worse, never better, so the FAIL above is not an artifact of an unfair measurement.
- Whether/how to close the two findings above (broaden `package_globs` coverage, and/or budget `performance.benchmark` differently) is a judgment call for whoever owns #2251, not made here.

### DIFF
--- a/.github/actions/ci-native-part/action.yml
+++ b/.github/actions/ci-native-part/action.yml
@@ -1,0 +1,30 @@
+name: emit CI lane native part
+description: Publish a completion-bound native evidence part for one registered producer
+
+inputs:
+  id:
+    description: Registry native part ID
+    required: true
+  seed:
+    description: Exact seed used by a seeded producer lane
+    required: false
+    default: ''
+
+runs:
+  using: composite
+  steps:
+    - name: emit the completion-bound native case
+      shell: bash
+      env:
+        CI_NATIVE_PART_ID: ${{ inputs.id }}
+        CI_NATIVE_PART_OUTPUT: .ci-native/${{ inputs.id }}/cases.json
+        CI_NATIVE_SEED: ${{ inputs.seed }}
+      run: node .github/scripts/ci-lane-native-part.mjs
+
+    - name: upload the attempt-qualified native part
+      uses: actions/upload-artifact@v7
+      with:
+        name: ci-native-part-${{ inputs.id }}-${{ github.run_attempt }}
+        path: .ci-native/${{ inputs.id }}/cases.json
+        if-no-files-found: error
+        retention-days: 14
--- a/.github/actions/prepare-release-promotion/action.yml
+++ b/.github/actions/prepare-release-promotion/action.yml
@@ -1,0 +1,78 @@
+name: prepare release promotion
+description: Download only the sealed candidate and successful qualification attestation, then verify both.
+
+inputs:
+  candidate_artifact:
+    description: Immutable candidate artifact name from this workflow run.
+    required: true
+  candidate_digest:
+    description: Exact sha256 digest of candidate-manifest.json.
+    required: true
+  attestation_artifact:
+    description: Immutable successful qualification-attestation artifact name.
+    required: true
+  attestation_digest:
+    description: Exact sha256 digest of the qualification attestation.
+    required: true
+  source_sha:
+    description: Exact source commit.
+    required: true
+  source_tree:
+    description: Exact source tree.
+    required: true
+  app_version:
+    description: Qualified application version.
+    required: true
+  app_tag:
+    description: Qualified application tag.
+    required: true
+  chart_version:
+    description: Qualified chart version.
+    required: true
+  correlation_nonce:
+    description: Fresh qualification nonce.
+    required: true
+  authorized:
+    description: Literal authorization-barrier result.
+    required: true
+  publish:
+    description: Literal per-artifact version-gate result.
+    required: true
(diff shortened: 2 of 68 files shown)


tsouza:
Closing — this backtested PR #2251's impact-routing approach, which is being abandoned in favor of a simpler fixed 2-tier CI model. The backtest's findings (documented in this PR) remain useful evidence for why the simpler approach was chosen, but the tool itself has no further purpose without #2251.

Issue #2350, opened 2026-08-17
update-golden.yml publish can race a PR merge, orphaning a regenerated golden diff


tsouza:
## What happened

PR #2347 dispatched `update-golden.yml -f shards=all -f branch=fix/histogram-mixed-setop-flatten-regression` while the PR was open. The PR was merged (and its head branch deleted, per the repo's normal `--delete-branch` convention) while that workflow run was still executing. All 10 `regenerate (*)` shard jobs finished successfully, but the final `publish golden updates` job's `Validate, commit, and push once` step failed:

    ##[error]target branch fix/histogram-mixed-setop-flatten-regression does not resolve to exactly one remote head

(run https://github.com/tsouza/cerberus/actions/runs/32078097124, job `publish golden updates`) — because `.github/scripts/manual-golden-update.mjs`'s `assertTargetUnmoved`/`remoteHead` check correctly refused to push to a branch that no longer exists, but by then the PR's code diff (which included a seed-data fix that changed what the golden `expected_rows` should be) was already on `main`, permanently orphaned from its own regenerated goldens. The result was two stale-golden fixtures silently breaking the `roundtrip (promql)` CI leg on `main` (https://github.com/tsouza/cerberus/actions/runs/32079793085/job/95540419656), fixed directly in #2349.

## The underlying gap

Nothing in the PR-merge path checks whether an `update-golden.yml` dispatch against that PR's branch is still in flight. A PR can look code-complete and green enough to merge while its own requested golden regen is still running, and merging deletes the branch out from under the workflow — the workflow fails loudly (good — it refuses to push somewhere wrong), but nothing stops the merge itself or re-surfaces the now-undelivered regen as an action item once the branch is gone. The regenerated diff (correct, already computed) is simply lost, and the stale golden reaches `main` looking like nothing is wrong until a later CI run (or a human) notices the drift.

## Suggested fix (not prescriptive — whoever picks this up should design it)

Some combination of: (a) a merge-time check/required-status-check that blocks merging while an `update-golden.yml` run targeting the PR's branch is queued/in-progress, and/or (b) having `manual-golden-update.mjs` retry against the merge commit / open a fresh PR with the regenerated diff instead of just failing closed when the original branch is gone, and/or (c) surfacing the failed `publish golden updates` job more visibly (e.g. a bot comment on the now-merged PR, or a follow-up issue auto-filed) so it doesn't require a human to notice a CI run failed on a branch nobody's watching anymore after merge.

[...]

PR #2414, opened 2026-08-20
test: disposable PR to verify update-golden-guard live

tsouza:
Disposable PR used only to live-verify `update-golden-guard.yml` against a
real in-flight `update-golden.yml` dispatch (issue #2350). Will be closed and
its branch deleted immediately after verification. Not meant to merge.


tsouza:
Disposable PR used only to live-verify update-golden-guard.yml against a real in-flight update-golden.yml dispatch (issue #2350). Verification complete — the guard correctly blocked while the dispatch was queued/in_progress and cleared once it completed. Closing without merging.

Issue #2114, opened 2026-08-13
promql: native ts_grid_range rate path is inert for DELTA-temporality schemas — needs a plan-level CUMULATIVE/DELTA split


tsouza:
## Context

Split out of #1963 (item 3) as its own tracked sub-issue — substantial enough to need its own design
decision, not a code tweak.

`internal/promql/lower.go`'s `nativeTSGridMatrixNode` refuses eligibility whenever
`rw.TemporalityColumn != ""` (a correctness guard: ClickHouse's `timeSeriesRateToGrid` has no
per-row runtime branch, so it cannot safely answer a DELTA-temporality window). Since the default
schema always declares `AggregationTemporalityColumn`, the opt-in `ts_grid_range` performance
feature is effectively inert for `rate()` over the default schema.

## Verified 2026-08-13, during #2113: ClickHouse has no temporality-aware native aggregate to switch to

Probed the chDB substrate directly — ClickHouse **26.5.1.1**, well above the 25.8 production floor
and the 25.9 `timeSeries*ToGrid` family floor. `system.functions` lists the whole family
(`timeSeriesRateToGrid`, `timeSeriesDeltaToGrid`, `timeSeriesInstantRateToGrid`,
`timeSeriesInstantDeltaToGrid`, `timeSeriesResetsToGrid`, `timeSeriesChangesToGrid`,
`timeSeriesDerivToGrid`, `timeSeriesPredictLinearToGrid`, `timeSeriesLastToGrid`,
`timeSeriesResampleToGridWithStaleness`) and **not one member takes a temporality argument**.
`timeSeriesRateToGrid` hard-rejects a third value argument:

```
Code: 42. DB::Exception: Aggregate function timeSeriesRateToGrid requires two arguments.
(NUMBER_OF_ARGUMENTS_DOESNT_MATCH)
```

So there is no per-row signal to hand the aggregate, and the "if ClickHouse 25.8+ exposes one" arm
of the original write-up is closed: it does not, as of 26.5.1.1.

## What remains: a plan-level split (route only CUMULATIVE-known series natively)

A real design decision, not a code change, needs to answer:

- **How is a series' temporality known before emit?** The value lives in the data, so either a
  pre-query catalogue read (an extra round trip, and a staleness question the no-caching invariant
  constrains) or a two-arm UNION whose native arm carries `WHERE AggregationTemporality != 1` and
  whose fan-out arm carries the complement.
[...]

PR #2465, opened 2026-08-22
perf(schema): index AggregationTemporality to prune the rate() split's redundant scan

### LINKED ISSUE #2458: perf(promql): NativeRateLowerer's temporality UnionAll split reads the base table twice
## What happens

`NativeRateLowerer.LowerRate` (`internal/promql/lower_strategy.go`) splits any rate()/increase()-eligible
range window whose selector carries `AggregationTemporality` into two `UnionAll` arms when the native
`ts_grid_range` path is active:

- a CUMULATIVE arm — `Filter(AggregationTemporality != DELTA)` feeding `chplan.RangeWindowGridNative`
  (the fast native `timeSeriesRateToGrid` aggregate, which has no DELTA semantics)
- a DELTA arm — `Filter(AggregationTemporality = DELTA)` feeding the unchanged fan-out `chplan.RangeWindow`

Both arms scan the **same underlying table** with the **same `MetricName`/`Attributes`/time-range
predicate**, differing only in the trailing `AggregationTemporality` conjunct — which is not part of
the table's `ORDER BY` (`MetricName, Attributes, TimeUnix`), so ClickHouse cannot skip granules on it.
Every row matching the selector is therefore read **twice**, once per arm, regardless of how many rows
actually carry the DELTA reading.

In real OTel deployments almost all Sum/Histogram data is CUMULATIVE-only (DELTA temporality for these
signal types is rare/opt-in), so the DELTA arm typically reads the entire matched row set and emits
**zero** rows — a pure, 100%-wasted second scan.

This is exactly the class of cost `FanoutRateLowerer`'s single-scan matrix emitter
(`emitWindowedArrayExtrapolatedMatrix`) already eliminated for the plain fan-out path — see the doc
comment on `TestFanoutRateLowererPreservesTemporalityColumn`
(`internal/promql/native_rate_temporality_test.go`): "chsql's single-scan matrix emitter ... now
branches on the temporality column per row inside one query, so the plan-level split ... is
unnecessary cost the lowerer no longer pays." `NativeRateLowerer` was written independently and never
received the equivalent fix.

## Real measurement (not a guess)

Using the real production-shaped sample already checked into `test/perf/nightly/testdata/samples/`
(`svc_http_requests_total.parquet`, `svc_http_request_duration_seconds.parquet` — 1,338,836 and
1,317,183 rows respectively) loaded into a real ClickHouse 25.8 (testcontainers), with
`ts_grid_range` forced on:

| query | plan | read_rows | table size | ratio |
|---|---|---|---|---|
| `rate(svc_http_requests_total[5m])` | native 2-arm | 2,677,672 | 1,338,836 | **2.00×** |
| `rate(svc_http_request_duration_seconds_count[5m])` | native 2-arm | 2,634,366 | 1,317,183 | **2.00×** |

Both queries reach `chplan.RangeWindowGridNative` via the companion-routing path
(`wrapHistogramCompanionProject` → `nativeTemporalityFilter` → the CUMULATIVE arm), confirming that
`rate(<histogram>_count[...])` / `rate(<histogram>_sum[...])` — an extremely common pairing alongside
`histogram_quantile` panels — is exactly the shape issue #2408 direction #1 asked about ("the ~55
queries that DO reach a native grid path over `otel_metrics_histogram`").

`EXPLAIN PIPELINE` on the corresponding CUMULATIVE-arm-only SQL text shows the ~1.3M-row `PREWHERE
MetricName IN (...) AND AggregationTemporality != 1` scan is a completely separate
`ReadFromMergeTree` from the DELTA arm's own `PREWHERE ... AggregationTemporality = 1` scan inside
the `UnionAll` — ClickHouse does not share the read between the two arms (no automatic
common-subexpression elimination across `UNION ALL` branches with only a trailing-predicate
difference, as expected).

At this sample's real cardinality, net query cost (duration/memory) was comparable to or slightly
better than the single-query fan-out alternative — the fan-out path pays its own ~2× read tax via the
`#2429` resource-bound pre-check subquery, a different, largely necessary mechanism. So this dead-arm
waste is real and precisely quantified (exactly 2× reads, confirmed on two independent real-data
queries), but is not by itself proven to be the dominant driver of #2408's originally-reported 269
CPU-s/query average — that figure comes from a 24h production corpus this investigation does not have
access to, and the same corpus shows native-rate-over-histogram reading ~26 GiB/query on average (18×
the bulk fan-out leg's per-query read), a gap this 2× finding only partially explains. Confirming the
rest needs the real corpus or the `#2370` large-scale perf lane once it can run this exact shape.

## Why this isn't a safe drop-in fix

`timeSeriesRateToGrid` is an opaque ClickHouse aggregate function with no DELTA semantics and no
`-If` combinator variant — you cannot feed it a temporality-mixed row stream and have it self-select.
Eliminating the duplicate *read* (as opposed to the duplicate *aggregation work*) requires either:

- ClickHouse materializing a shared filtered-row CTE once and feeding both arms from it — standard
  ClickHouse (25.8, no special settings) does not do this for a plain `WITH ... AS (SELECT ...)`
  (it's macro-inlined, not memoized), so this would need to be built and proven, not assumed; or
- Restructuring `NativeRateLowerer` to read the base selector once, use `timeSeriesRateToGridIf` (a
  combinator that does not exist today, requiring a chsql function-vocabulary addition — the
  `AggCombinator` sealed vocabulary already has the pattern via `CombIf`) for the CUMULATIVE
  reduction, and fuse the DELTA fan-out's own array/window machinery onto the *same* row stream in
  one query — a real rewrite of the native-rate emit path, not a small patch.

Both are legitimate engineering projects with real payoff (measured: exactly halves the read cost of
every native-eligible `rate()`/`increase()` query over a Sum/Histogram-table counter in the — very
common — case where the metric is provably single-temporality), but neither is safe to land as a
narrow patch without its own design pass and differential-parity coverage.

## Suggested next step

Design and implement a single-scan fusion for `NativeRateLowerer`'s temporality-bearing case,
eliminating the DELTA arm's redundant full scan when it contributes rows (ideally without a
data-dependent runtime pre-check, given cerberus's no-caching / single-round-trip architecture). A
`test/perf/nightly` sentinel pinning `rate(svc_http_requests_total[5m])`'s (or the `_count` companion's)
`read_rows` / CPU cost before and after is a natural first deliverable alongside the fix — see #2408
direction #4.

## Scope note

This is scoped narrowly to the CUMULATIVE/DELTA `UnionAll` split's redundant *read*, not the routing
split's *existence* (#2114/#2115/#1963 already fixed the correctness gap that split exists to close —
this issue is about its cost, not its correctness). It is also distinct from #2447 (RangeBucketFanout's
per-sample fan-out width / cardinality axis) — this is about `RangeWindowGridNative`'s own temporality
routing, a different node entirely.

## Related

- #2408 (parent investigation — direction #1). Direction #2 of #2408 (push `RangeBucketFanout`'s
  `sumForEach` collapse earlier) was investigated and answered negatively in the same session: see
  the comment on #2408. No follow-up needed for that direction.
- #2114 / #2115 / #1963 — added the temporality split this issue is about the cost of.
- #2429 — the fan-out path's own resource-bound pre-check subquery, the mechanism behind the ~2× read
  tax the fan-out alternative pays for a different reason (kept here only as the comparison point the
  measurement above relies on).

tsouza:
## What

Fixes the 2x redundant read #2458 measured for `NativeRateLowerer`'s temporality
split, by adding a `minmax` skip index on `AggregationTemporality` to the sum
and histogram metrics tables (`internal/schema/ddl`) — an additive
`ALTER TABLE ... ADD INDEX IF NOT EXISTS`, applied idempotently at boot via the
same mechanism the curated metadata-projection registry already uses.

## Why this shape, not a plan-level fix

`NativeRateLowerer.LowerRate` splits a temporality-bearing `rate()`/`increase()`
range window into a native CUMULATIVE arm (`chplan.RangeWindowGridNative`, fed
only non-DELTA rows) and a fan-out DELTA arm (`chplan.RangeWindow`, fed only
DELTA rows) — a `chplan.UnionAll{RangeWindowGridNative, RangeWindow}`. Both
arms scan the same base table with the same `MetricName`/`Attributes`
predicate, differing only in a trailing `AggregationTemporality` conjunct that
carries no `ORDER BY` benefit, so ClickHouse reads every matching row from
**both** arms — the issue's own real-data measurement: a confirmed 2.00x
`read_rows` ratio against table size on two independent `rate()` queries.

I investigated eliminating the split (the "obvious" fix — just delegate the
whole window to the fan-out `RateLowerer` whenever a temporality column is
present, matching how `FanoutRateLowerer` itself already handles mixed
temporality in one scan) and reverted it after finding it breaks load-bearing,
deliberately-built infrastructure:

- `internal/solver/native_union_route_test.go`'s
  `TestPlan_ProductionTemporalitySplitReanchorsBothArms` lowers the exact
  `rate(cerberus_queries_total[5m])` production shape and asserts the plan
  still contains **exactly one** `RangeWindowGridNative` arm and **one**
  `RangeWindow` arm — this is the precise shape #2114/#2117/#2120 taught the
  solver to slice and re-anchor for OOM-safe routing. Eliminating the split
  would silently disconnect that solver machinery from ever firing for
  `rate()` in production, and is explicitly out of scope per the issue's own
  "Scope note" ("not the routing split's *existence* ... this issue is about
  its *cost*").
- It would also have gutted the golden corpus's demonstrated coverage of the
  native aggregate for the (very common) plain-Sum-table-counter shape: 11 of
  the 12 `experimental_ts_grid_range` TXTAR fixtures currently reach this
  exact split, since the default schema declares `AggregationTemporality`
  broadly.

I then investigated a genuine single-scan SQL fusion — keep the plan shape,
make the two arms share one physical read — and confirmed empirically
(against chDB) that:

- a shared `WITH ... AS (subquery)` CTE does **not** save a physical read:
  `EXPLAIN PLAN` shows two independent `ReadFromMergeTree` nodes even for a
  byte-identical CTE referenced from both `UNION ALL` branches — ClickHouse
  macro-inlines it rather than memoizing it, confirming the issue's own
  caution ("this would need to be built and proven, not assumed");
- `timeSeriesRateToGridIf(...)(ts, value, cond)` — the `-If` combinator the
  issue names as the missing piece for a real fusion — **does** compose with
  the parametric native aggregate on this ClickHouse version (25.8), both as
  a plain aggregate and as a window function (`... OVER (PARTITION BY ...)`).

That second finding is genuinely new information beyond what the issue
established, and de-risks a future real fusion attempt, but building the rest
of it — restructuring the DELTA arm's array/window machinery to share one
row-pass with the native aggregate call, with its own differential-parity
coverage — is still the "real rewrite ... not a small patch" the issue's own
"Why this isn't a safe drop-in fix" section describes, not something safe to
attempt as a narrow patch here.

## The fix

A `minmax` skip index (metadata-only — unlike `ADD PROJECTION` it does not
store a second physical copy of the table) on `AggregationTemporality`, sized
`GRANULARITY 1` for the finest skip resolution. Real OTel deployments set
`AggregationTemporality` once per exporter configuration, so a given series'
samples land in temporality-homogeneous runs almost always — which means a
homogeneous granule's `[min, max]` mark is a single value, letting ClickHouse
prune it entirely for whichever arm's predicate doesn't match. This addresses
the redundant read with **zero** change to `chplan`/`chsql`/`promql`, so zero
risk to the solver routing tests, and it benefits every query that filters on
this column against these tables — including the parallel split
`NativeClassicHistogramWindowLowerer.LowerClassicHistogramWindow` builds for
the classic-histogram quantile window stage (same table, same predicate
shape), as a side effect with no code change there either.

### Verification

`internal/schema/ddl/temporality_index_skip_chdb_test.go` (chdb-tagged) proves
the mechanism against a live `EXPLAIN indexes=1` plan on the package's own
rendered production DDL — not a text-parse of the `ALTER` string:

- `TestTemporalityIndex_PrunesAllCumulativeGranules`: seeds 4 real granules
  (32,768 rows, the production table's actual `index_granularity=8192`) of an
  all-CUMULATIVE series. The DELTA-arm predicate (`AggregationTemporality = 1`)
  prunes **100%** of granules (`Granules: 0/N`); the CUMULATIVE-arm predicate
  correctly reads every granule (nothing to prune — every row matches it).
- `TestTemporalityIndex_DoesNotPruneAMixedGranule`: the required negative
  case — a granule carrying one DELTA row alongside CUMULATIVE ones must not
  be pruned for either predicate, and the query answer itself must be
  unaffected by the index (`count() == 1`, not silently dropped).

I did not additionally wire a `test/perf/nightly` read_rows sentinel against
the real captured sample data (the issue's own "Suggested next step" and the
methodology it used) — that harness is mid-rollout on its own PR-A/PR-B
schedule per its doc comments, and the chDB proof above already demonstrates
the causal mechanism directly. Worth adding as a follow-up once that harness's
baseline lands.

### Existing data

New parts get the index automatically once the `ALTER` runs (idempotent,
re-applied every boot). Existing parts need a one-time backfill —
`ALTER TABLE ... MATERIALIZE INDEX idx_agg_temporality` — documented in
`docs/operations.md` alongside the mechanism, mirroring the existing
`MATERIALIZE PROJECTION` runbook.

## Files

- `internal/chsql/ddl.go` (+test): new `AlterTableAddIndex` typed DDL builder
  (`ALTER TABLE ... ADD INDEX IF NOT EXISTS ... TYPE ... GRANULARITY ...`).
- `internal/schema/ddl/ddl.go` (+tests): wires the index onto the sum and
  histogram tables in the `Metrics` DDL signal.
- `internal/schema/ddl/temporality_index_skip_chdb_test.go`: the live
  granule-pruning proof described above.
- `docs/operations.md`: mechanism + backfill runbook, alongside the existing
  projection-registry section it mirrors.
- `test/e2e/migration/steps/schema.go` / `schema_live.go` (+tests): the new
  `ADD INDEX` ALTER tripped two migration-lane gates that only recognized
  `ADD PROJECTION` as a safe, idempotent non-CREATE statement — widened both
  to accept `ADD INDEX IF NOT EXISTS` too (same safety profile: metadata-only,
  never rewrites or destroys data), with the MIG-10 live-diff half getting the
  equivalent `renderedIndex`/`parseAddIndex` plumbing mirroring its existing
  projection handling.
- `.github/scripts/lib/golden-shards.mjs`: the `migration` shard's declared
  golden ownership only listed `test/e2e/migration/archetypes/*/expected`,
  not `test/e2e/migration/expected/schema` — a pre-existing gap (reachable by
  *any* rendered-DDL-shape change, not specific to this PR) that made
  `update-golden.yml` reject this PR's own schema-golden regeneration as
  "outside shard ownership." Declared the missing path.
- `test/e2e/migration/expected/schema/{default,metrics-ttl-override}.sql`:
  regenerated for the two new `ADD INDEX` statements. Done via
  `just migration-golden` directly rather than `update-golden.yml`: the
  workflow's controller scripts run from the WORKFLOW ref (`main`) by
  design — a deliberate trust boundary — so this PR's own fix to
  `golden-shards.mjs` above cannot take effect for the PR's own dispatch
  until it's on `main`. Confirmed by two separate `update-golden.yml`
  dispatches, both failing identically before and after that fix landed on
  this branch. The diff is exactly the two new ALTERs; `go test
  -tags=migration ./test/e2e/migration/tiers/tier0-offline/...` passes
  clean (8/8 scenarios).

## Testing

- `go test -race ./internal/schema/... ./internal/chsql/... ./internal/solver/...`
- `go test -tags chdb -run TestTemporalityIndex ./internal/schema/ddl/...`
  (both new cases pass, live against chDB)
- `go test -tags=migration ./test/e2e/migration/tiers/tier0-offline/...`
  (8/8 scenarios pass, including both `render the schema under a schema
  environment` cases)
- `go test ./test/e2e/migration/steps/...` and `go vet -tags=migration
  ./test/e2e/migration/...`
- `node --test .github/scripts/manual-golden-update.test.mjs` (11/11 pass)
- `node .github/scripts/forbid-sql-raw.mjs` — clean
- Confirmed `internal/solver`'s `TestPlan_ProductionTemporalitySplitReanchorsBothArms`,
  `TestPlan_NativeUnionSpineRoutes`, `TestPlan_NativeUnionShardsMoveBothArms`
  still pass unchanged (this PR touches no `chplan`/`chsql`/`promql` code)

Closes #2458

### DIFF
--- a/.github/scripts/lib/golden-shards.mjs
+++ b/.github/scripts/lib/golden-shards.mjs
@@ -215,7 +215,15 @@ export const SHARDS = {
   migration: {
     stage: STAGE_PRE,
     recipe: 'migration-golden',
-    goldens: ['test/e2e/migration/archetypes/*/expected'],
+    // migration-golden's MIGRATION_UPDATE_GOLDENS=1 run rewrites BOTH the
+    // per-archetype corpus goldens AND the MIG-10 rendered-schema goldens
+    // (test/e2e/migration/expected/schema/*.sql, one per schemaCases entry
+    // in test/e2e/migration/steps/schema.go) — a DDL-shape change (a new
+    // ADD PROJECTION / ADD INDEX ALTER, a retention/engine default change)
+    // moves the latter without touching a single archetype fixture, so both
+    // paths must be declared or the schema half fails ownership here even
+    // though the SAME recipe regenerated it.
+    goldens: ['test/e2e/migration/archetypes/*/expected', 'test/e2e/migration/expected/schema'],
     corpus: ['test/e2e/migration/archetypes'],
     // The Tier-0 harness does not emit the explain reports in-process: it
     // spawns the built `cerberus migrate explain` and captures its output
--- a/docs/operations.md
+++ b/docs/operations.md
@@ -1054,6 +1054,63 @@ costs — on a wide-`Attributes` table the read side dominates. Materialize one
 projection at a time and watch `system.mutations` before starting the next so
 a maintenance window isn't saturated by both at once.
 
+### AggregationTemporality skip index
+
+Auto-create also installs a small **data-skipping index** on the sum and
+histogram fact tables:
+
+```sql
+ALTER TABLE <db>.otel_metrics_sum       ADD INDEX IF NOT EXISTS idx_agg_temporality AggregationTemporality TYPE minmax GRANULARITY 1
+ALTER TABLE <db>.otel_metrics_histogram ADD INDEX IF NOT EXISTS idx_agg_temporality AggregationTemporality TYPE minmax GRANULARITY 1
+```
+
+Unlike the metadata projections above, this is a `minmax` **skip index**, not
+a stored second copy of the table — it costs a small amount of per-granule
+`[min, max]` metadata, not additional table storage.
+
+**Why it exists (issue #2458).** A range-mode `rate()`/`increase()` window
+over a temporality-bearing counter can split into a native
+`timeSeriesRateToGrid` arm (fed only non-DELTA rows) and a fan-out arm (fed
+only DELTA rows) — see `NativeRateLowerer.LowerRate` in
+`internal/promql/lower_strategy.go`. Both arms scan the SAME base table with
+the SAME `MetricName`/`Attributes` predicate, differing only in a trailing
+`AggregationTemporality` conjunct. That column is not part of the table's
+`ORDER BY` (`MetricName, Attributes, ServiceName, TimeUnix`), so without a
+skip index ClickHouse cannot prune a single granule on it and reads every
+matching row from BOTH arms — a confirmed, reproducible 2.00x `read_rows`
+ratio against table size, measured on real production-shaped data. Real OTel
+deployments set `AggregationTemporality` once per exporter configuration, so
+a given series' samples land in temporality-homogeneous runs almost always;
+the minmax index lets ClickHouse recognize a homogeneous granule and skip it
+entirely for whichever arm's predicate does not match, without requiring any
+change to the plan shape or the table's `ORDER BY`. The same index also
+prunes the ordinary single-arm case: any plain scan carrying an
+`AggregationTemporality` predicate (the fan-out emitter's own per-row branch,
+or a future consumer) benefits identically.
+
+`ADD INDEX IF NOT EXISTS` is metadata-only and idempotent, so the auto-create
+hook (re)applies it safely on every boot, covering both freshly-created and
+pre-existing tables. **New parts written after the ALTER carry the index
+automatically; existing parts are not back-filled by `ADD INDEX` alone.**
+
+#### One-time `MATERIALIZE INDEX` back-fill runbook
+
+To back-fill existing parts immediately on a deployment that predates the
+index, run the one-time materialize per table (a background mutation,
+non-blocking for reads):
+
+```sql
+ALTER TABLE <db>.otel_metrics_sum       MATERIALIZE INDEX idx_agg_temporality;
+ALTER TABLE <db>.otel_metrics_histogram MATERIALIZE INDEX idx_agg_temporality;
+```
+
+`MATERIALIZE` is intentionally **not** issued by the auto-create hook — it
+rewrites index metadata for every existing part and belongs in a deliberate
+maintenance window, not the boot path. Track progress in `system.mutations`
+(the `is_done` / `parts_to_do` columns). A `MATERIALIZE INDEX` mutation only
+reads and rewrites the indexed column's marks, not the whole part, so it is
+far cheaper than a `MATERIALIZE PROJECTION` over the same table.
+
 ### Startup requirements preflight
 
 `CERBERUS_REQUIREMENTS_CHECK` (**on by default**) runs a boot-time
--- a/internal/chsql/ddl.go
+++ b/internal/chsql/ddl.go
@@ -694,3 +694,82 @@ func (a *AddColumnBuilder) frag() Frag {
 func (a *AddColumnBuilder) SQL() string {
 	return RenderDDL(a.frag())
 }
+
+// --- ALTER TABLE ... ADD INDEX surface ---
+//
+// AddIndexBuilder renders
+// `ALTER TABLE [<db>.]<table> [ON CLUSTER x] ADD INDEX IF NOT EXISTS <name>
+// <expr> TYPE <type> GRANULARITY <n>`, the statement that installs a
+// ClickHouse data-skipping index on a column already present on a table
+// cerberus does not otherwise own the CREATE TABLE body of (the upstream
+// OTel exporter template — see the package doc comment). Adding a skip
+// index is metadata-only for NEW parts; it does not reorder or duplicate
+// existing data (unlike a PROJECTION, which stores a second physical copy),
+// but existing parts need a separate `ALTER TABLE ... MATERIALIZE INDEX`
+// backfill to benefit retroactively — see docs/operations.md.
+//
+// Like the other DDL builders it binds no positional `?` values, so SQL
+// renders through RenderDDL.
+
+// AddIndexBuilder builds an ALTER TABLE ADD INDEX statement.
+type AddIndexBuilder struct {
+	database    string // "" => unqualified table reference
+	table       string
+	name        string
+	cluster     string // "" => no ON CLUSTER clause
+	expr        Frag
+	indexType   string
+	granularity int
+}
+
+// AlterTableAddIndex starts an ADD INDEX builder installing a data-skipping
+// index named <name> over expr on [<database>.]<table>, of the given
+// ClickHouse index type ("minmax", "set(N)", "bloom_filter", …) and
+// GRANULARITY. An empty database emits no qualifier, so a table the
+// connection's own database owns is referenced bare.
+func AlterTableAddIndex(database, table, name string, expr Frag, indexType string, granularity int) *AddIndexBuilder {
+	return &AddIndexBuilder{database: database, table: table, name: name, expr: expr, indexType: indexType, granularity: granularity}
+}
+
+// OnCluster adds an `ON CLUSTER <name>` clause so the ALTER replicates the
+// same way the CREATE statements do under a classic ON CLUSTER deployment.
+// A Replicated database replicates the DDL itself and needs no clause.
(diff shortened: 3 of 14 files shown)


tsouza:
Filed the harder follow-up as #2466 (genuine single-scan native+DELTA SQL fusion) — captures the chDB findings from this PR's investigation (CTE sharing does not save a physical read; `timeSeriesRateToGridIf`'s -If combinator does compose) as prior art for whoever picks that up.

Issue #2346, opened 2026-08-17
promql: mixed float/histogram 'or' still rejected when composed under a wrapper (sum(a or b))


tsouza:
## Problem

Cerberus rejects a mixed float/histogram `or` when it is NOT the whole query — i.e. composed under a further PromQL wrapper such as `sum(...)` — while reference Prometheus accepts it the same way it accepts the root-level case #2330/#2335 already answers.

Representative query on current main:

```promql
sum(demo_latency_exp_hist or histogram_quantile(0.5, demo_latency_exp_hist))
```

## Evidence

- `internal/promql/histogram_native_mixed_or.go`'s `mixedExpHistogramSetOp` recognizer (added by #2335, closing #2330) is deliberately registered ROOT-ONLY in `lowerRoot` — see that file's own header doc: "Composing a Mixed result under a further PromQL wrapper is real follow-on work, not attempted here... `sum(a or b)` / `abs(a or b)` for a mixed `a or b` still fall through to \[cerberus's pre-existing rejection\], exactly as they did before this file existed."
- Concretely, a nested mixed `or` now reaches `internal/promql/binary.go:lowerVectorSetOp`'s own `leftHistogram != rightHistogram` guard and hits: `promql: 'or' between a float-valued and a histogram-valued operand is not supported; 'and'/'unless' support mixing them`.
- Surfaced while rebasing #2337 (cerberus issue #2325's fix) past #2335: the rejection-parity catalogue's `internal/promql/binary.go:lowerVectorSetOp#4924e0ba` divergence entry, previously tracked under #2330, had its root-level trigger query self-satisfied by #2335's own fix. Repointed to the nested trigger above, which needs its own tracking issue since #2330 is closed and scoped to the root-level case only.

## Acceptance

- `sum(demo_latency_exp_hist or histogram_quantile(0.5, demo_latency_exp_hist))` (and the general nested-composition case for a mixed `or`) either answers with reference-matching Mixed-result semantics, or the rejection-parity catalogue continues to track it explicitly as a divergence under this issue until it does.
- Add focused test coverage (unit + rejection-parity trigger) for the accepted behavior once implemented.

PR #2469, opened 2026-08-22
fix(promql): accept a windowed float side in a sum/avg-wrapped mixed or

### LINKED ISSUE #2333: promql: mixed float/histogram 'or' rejects a windowed or derived float side
## What's missing

`or` between a float-valued and a histogram-valued operand (cerberus issue #2330) now lowers and
executes correctly when the float side is a *plain, non-windowed* float shape — the canonical
`(MetricName, Attributes, TimeUnix, Value)` row a bare selector, a scalar binop, or
`histogram_quantile()` publishes. Reference Prometheus places no such restriction: any float-valued
vector, including one produced by a range-vector function (`rate(...)`), an `_over_time` reducer, or
a matrix-mode `query_range` window, can appear on either side of `or` against a histogram-valued
operand.

Cerberus rejects the windowed/derived case today with an explicit, clear error (not a silent
wrong-SQL bug):

```
rate(up[5m]) or latency_exp_hist
```

fails with `promql: 'or' between a float-valued and a histogram-valued operand only supports a plain
(non-windowed) float side today; got a grid-window-shaped float operand, which cerberus issue #2330
does not yet cover` — raised by `lowerMixedExpHistogramSetOp` in
`internal/promql/histogram_native_mixed_or.go`.

## Why #2330 didn't cover this

#2330's fix widens both arms of a `VectorSetOr` to a shared fourteen-column projection — the
canonical quartet, the nine `Histogram*Column` outputs, and a trailing per-row discriminator — via
`internal/chsql/vector_set_op.go`'s `emitMixedVectorSetOp` and
`vectorSetOpCanonicalQuartetFrags`/`mixedVectorSetOpArmFrag`. That canonicalisation already handles
three shapes for an arm feeding a *plain* (non-Mixed) `VectorSetOp` — canonical-shape input, a matrix
`RangeWindow`, and an instant derived-shape input (see `vectorSetOpCanonicalArmFrag`'s doc comment) —
but `lowerMixedExpHistogramSetOp` deliberately restricts the float side of a MIXED set op to
`chplan.SampleRowShape` only, rejecting the other two, to keep the initial change reviewable and
avoid combining two axes of complexity (histogram/float asymmetry AND windowed-shape reconciliation)
in one PR.

Closing this gap means teaching `mixedVectorSetOpArmFrag` (or a variant of it) the same
matrix/derived-shape canonicalisation `vectorSetOpCanonicalArmFrag` already has for the plain case,
and re-verifying the fourteen-column UNION ALL still type-unifies correctly when a matrix arm's
`anchor_ts`-sourced timestamp and a derived arm's synthesized timestamp are ALSO in play alongside
the Mixed-specific histogram-placeholder columns.

## Repro

```
rate(up[5m]) or latency_exp_hist
latency_exp_hist or rate(up[5m])
sum(rate(up[5m])) or latency_exp_hist
```

against `schema.DefaultOTelMetrics()`, `promql.LowerAtRange` — all three are rejected by
`lowerMixedExpHistogramSetOp`'s explicit "plain (non-windowed) float side" guard.

### LINKED ISSUE #2453: promql: sum()/avg()-wrapped mixed histogram/float 'or' still rejects a windowed float side
## What's missing

A mixed float/histogram `or` with a windowed/derived float side (cerberus issue #2333, closed) works at the query root. `sum`/`avg`-wrapped composition of a mixed `or` (cerberus issue #2346, closed) works for a plain float side. The INTERSECTION of the two — a windowed/derived float side, wrapped in `sum`/`avg` — is not yet supported by either fix:

```promql
sum(rate(up[5m]) or demo_latency_exp_hist)
```

Reference Prometheus places no restriction here; cerberus rejects it via `lowerMixedExpHistogramOperands`'s float-side shape guard, since `lowerSumOrAvgOverMixedExpHistogramSetOp` (the #2346 aggregate-wrapped recognizer) deliberately passes `allowWindowedFloatSide=false` — it feeds the float side into its own aggregation reduction (`lowerPlainAggOverMixedFloatArm`), which has not been taught the matrix/derived-shape canonicalisation the root-only leaf path uses.

## Why this is a separate issue from #2333 and #2346

Both #2333 and #2346 explicitly scoped this combination out in their own text and tests — #2333's own body: "If sum(rate(...)) or X still fails after your fix because of the SEPARATE root-only-registration restriction, that is expected and out of scope for #2333"; #2346's own test `TestLower_ExpHistogram_MixedSetOpOr_SumWrapped_WindowedFloatSideRejects` pins that this combination still rejects. Neither issue's acceptance criteria covered it, so closing both left this real, narrower gap untracked.

## What would close this

Teach `lowerPlainAggOverMixedFloatArm` (histogram_native_mixed_or_aggregate.go) the same matrix/derived-shape canonicalisation the root-only leaf path gets via `mixedVectorSetOpArmFrag`/`vectorSetOpCanonicalQuartetFrags`, then flip `lowerSumOrAvgOverMixedExpHistogramSetOp`'s call to `lowerMixedExpHistogramOperands(..., allowWindowedFloatSide: true)`.

## Repro

```
sum(rate(up[5m]) or demo_latency_exp_hist)
avg(demo_latency_exp_hist or rate(up[5m]))
```

against `schema.DefaultOTelMetrics()`, `promql.LowerAt` — both rejected by `lowerMixedExpHistogramOperands`'s float-side shape guard (`internal/promql/histogram_native_mixed_or.go`).

tsouza:
## What

`sum(rate(up[5m]) or demo_latency_exp_hist)` and `avg(demo_latency_exp_hist or rate(up[5m]))`
now lower successfully. This closes the gap #2333 and #2346 each explicitly scoped out of
their own fix: #2333 taught the root-only mixed histogram/float `or` recognizer to accept a
windowed/derived float side; #2346 taught the `sum()`/`avg()`-wrapped composition to accept a
mixed `or`, but only for a plain (non-windowed) float side. The intersection — windowed float
side **and** sum/avg-wrapped — was tracked as #2453.

## Why it was rejected

`lowerSumOrAvgOverMixedExpHistogramSetOp` (`internal/promql/histogram_native_mixed_or_aggregate.go`)
feeds the mixed `or`'s float-valued arm into its own aggregation reduction
(`lowerPlainAggOverMixedFloatArm`), which references the schema's `TimestampColumn` /
`AttributesColumn` by name. That only resolves when the arm is already canonical-shaped
(`chplan.SampleRowShape`); a raw windowed/derived arm (a matrix `RangeWindow` or an
instant-reduced one) doesn't expose those names directly. The root-only leaf recognizer
(`lowerMixedExpHistogramSetOp`, #2333) never hit this because `internal/chsql`'s
`vectorSetOpCanonicalArmFrag` / `vectorSetOpCanonicalQuartetFrags` canonicalise a `VectorSetOp`
arm at SQL-emission time regardless of shape — but that canonicalisation is chsql-internal and
wasn't available to the aggregate wrapper's own `chplan.Aggregate` construction.

Of the mixed `or`'s two operands, one already gets this canonicalisation for free: the side
that needs shadow-filtering (`mixedOrShadowUnless`'s own `VectorSetOp`) is canonicalised as a
side effect of that node's own SQL emission. The other side — the one kept unconditionally per
`or`'s own shadow rule (the source AST's LHS) — never passes through a `VectorSetOp` at all, so
nothing canonicalised it.

## The fix

Added `canonicalizeFloatArmForAgg` (`internal/promql/histogram_native_mixed_or_aggregate.go`),
a promql-level mirror of `vectorSetOpCanonicalQuartetFrags`: a no-op for an already-canonical
arm, a passthrough Timestamp reference for a matrix `RangeWindow` (which already exposes the
grid anchor under the schema's `TimestampColumn` name), and a synthesised instant anchor
(`chplan.NowNanoMinusStaleness`, byte-for-byte what `vectorSetOpSynthesizedAnchorFrag` renders)
for an instant-reduced shape. `lowerSumOrAvgOverMixedExpHistogramSetOp` applies it to whichever
arm skips `mixedOrShadowUnless`'s own canonicalisation.

With both arms now reliably canonical before reaching `lowerPlainAggOverMixedFloatArm`, the
`lowerMixedExpHistogramOperands` shape guard's `allowWindowedFloatSide` parameter is dropped —
both callers accept a windowed float side now, so the parameter was down to always `true`.

## Tests

- `TestLower_ExpHistogram_MixedSetOpOr_SumWrapped_WindowedFloatSide` replaces the old pinned
  `..._WindowedFloatSideRejects` test (which explicitly locked in the *old* rejecting
  behaviour) with cases covering both operand orderings, `sum` and `avg`, and a `by (...)`
  clause, in instant mode.
- `TestLower_ExpHistogram_MixedSetOpOr_SumWrapped_RangeMode` covers the same query in range
  mode, since the windowed arm's canonicalisation takes a different branch there (passthrough
  vs. synthesised anchor).
- Verified with `chsql.Emit` locally (not checked in) that both repro queries from the issue
  render to valid SQL text with no missing-column references.

## Rejection-parity catalogue

This PR's fix eliminates the `allowWindowedFloatSide` branch entirely, so the `divergence`
catalogue entry for `sum(rate(up[5m]) or demo_latency_exp_hist)` (tracking issue #2453,
`internal/promql/histogram_native_mixed_or.go:lowerMixedExpHistogramOperands#38b18b37`) no
longer has a rejection site to repoint to — the query now succeeds. Dispatching
`update-golden.yml` with the `parity` shard (auto-widened as needed) to regenerate the
catalogue, which should delete that entry per its own documented ratchet ("if someone fixes
the divergence, claim 1 fails ... and the entry must be deleted").

Closes #2453

### DIFF
--- a/internal/promql/histogram_native_mixed_or.go
+++ b/internal/promql/histogram_native_mixed_or.go
@@ -108,11 +108,7 @@ func lowerMixedExpHistogramSetOp(b *parser.BinaryExpr, s schema.Metrics, ctx low
 		return nil, fmt.Errorf("promql: 'bool' modifier is only allowed on comparison binary ops")
 	}
 
-	// The root-only leaf recognizer allows a windowed/derived float side
-	// (cerberus issue #2333) — see lowerMixedExpHistogramOperands's own
-	// doc comment for why this is NOT the same acceptance the sum/avg
-	// wrapper (histogram_native_mixed_or_aggregate.go, issue #2346) gets.
-	histNode, floatNode, histOnLeft, err := lowerMixedExpHistogramOperands(b, s, ctx, true)
+	histNode, floatNode, histOnLeft, err := lowerMixedExpHistogramOperands(b, s, ctx)
 	if err != nil {
 		return nil, err
 	}
@@ -147,23 +143,30 @@ func lowerMixedExpHistogramSetOp(b *parser.BinaryExpr, s schema.Metrics, ctx low
 // SAME two operand plans without duplicating the histogram-valued /
 // float-valued dispatch.
 //
-// allowWindowedFloatSide is per-caller, not a shared constant, because the
-// two callers' float-side acceptance genuinely differs:
+// Both callers accept a windowed/derived float side (matrix RangeWindow
+// or an instant-reduced shape), not just the plain canonical-shape float
+// side #2330 originally shipped — the shape guard below accepts all
+// three row shapes internal/chsql's vectorSetOpCanonicalArmFrag doc
+// comment names. The two callers earn that acceptance through two
+// DIFFERENT mechanisms, though, which is worth knowing when touching
+// either:
 //   - The root-only leaf recognizer ([lowerMixedExpHistogramSetOp], issue
-//     #2333) passes true: internal/chsql's mixedVectorSetOpArmFrag
-//     canonicalises a matrix RangeWindow or a reduced instant shape
-//     through the exact same vectorSetOpCanonicalQuartetFrags helper the
-//     plain (non-Mixed) VectorSetOp path already uses, so a windowed/
-//     derived float side is safe there.
+//     #2333) relies on internal/chsql's mixedVectorSetOpArmFrag, which
+//     canonicalises the float arm through the exact same
+//     vectorSetOpCanonicalQuartetFrags helper the plain (non-Mixed)
+//     VectorSetOp path already uses — entirely inside the chsql emitter,
+//     no chplan-level change needed.
 //   - The sum/avg-wrapped recognizer
-//     ([lowerSumOrAvgOverMixedExpHistogramSetOp], issue #2346) passes
-//     false: it feeds floatNode into its OWN aggregation reduction
-//     ([lowerPlainAggOverMixedFloatArm]), which has not been taught that
-//     canonicalisation, so a windowed float side there is a separate,
-//     unimplemented axis — pinned by
-//     TestLower_ExpHistogram_MixedSetOpOr_SumWrapped_WindowedFloatSideRejects.
+//     ([lowerSumOrAvgOverMixedExpHistogramSetOp], issue #2346 then #2453)
+//     feeds floatNode into its OWN aggregation reduction
+//     ([lowerPlainAggOverMixedFloatArm], histogram_native_mixed_or_aggregate.go),
+//     which cannot rely on the chsql emitter the same way: one of its two
+//     possible float arms never passes through a VectorSetOp at all (the
+//     "kept unconditionally" side of the `or` shadow rule), so that file's
+//     own [canonicalizeFloatArmForAgg] mirrors vectorSetOpCanonicalQuartetFrags
+//     at the chplan level instead.
 func lowerMixedExpHistogramOperands(
-	b *parser.BinaryExpr, s schema.Metrics, ctx lowerCtx, allowWindowedFloatSide bool,
+	b *parser.BinaryExpr, s schema.Metrics, ctx lowerCtx,
 ) (histNode, floatNode chplan.Node, histOnLeft bool, err error) {
 	histOnLeft = isExpHistogramValuedShape(b.LHS, s, ctx)
 	histExpr, floatExpr := b.LHS, b.RHS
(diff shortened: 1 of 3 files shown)


tsouza:
Duplicate of #2464 — same fix for the same issue (#2453), produced by an accidental re-run of the
same automated batch after an interrupted resume. #2464 is the more complete of the two (includes
the golden-shard regeneration, rejection-parity catalogue rotation, and docs/coverage.md update this
one is missing) and is the one being carried forward. Closing without merging; no unique content here
is being lost.

Issue #2509, opened 2026-08-23
promql: info() hard-rejects a bare exp-histogram base vector instead of joining/preserving it


tsouza:
`info(v)` hard-rejects a bare exponential-histogram selector via
`internal/promql/lower.go`'s `expHistogramSelectorRouting` catch-all:

```
promql: "demo_latency_exp_hist" is an exponential histogram metric; only a
bare "demo_latency_exp_hist" selector, sum()/avg()/count() over one,
rate()/increase()/delta()/irate()/idelta()/sum_over_time()/avg_over_time()/
resets()/changes() over one, histogram_quantile(), histogram_count(),
histogram_sum(), and the "demo_latency_exp_hist_count"/
"demo_latency_exp_hist_sum" companion selectors are supported
```

`lowerInfo` (`internal/promql/info_fn.go`) lowers its base-vector argument
through the generic `lower(c.Args[0], s, ctx)` with no
`lowerExpHistogramValuedShape` check, so any histogram-valued base falls
through to that rejection.

Reference Prometheus's `info()` implementation
(`promql/info.go::combineWithInfoVector` in `tsouza/prometheus`) does
**not** drop histogram samples in the base vector the way the
"process only float samples" family does (issues #2221, #2345, #2456,
#2498) — `evalInfo`'s `addToSeries` step copies both `sample.F` and
`sample.H` into the output series for the base vector unchanged, and the
float-only check (`"info sample should be float"`) applies to the *info*
metric's own samples (e.g. `target_info`), never to the base vector being
annotated. So `info(a_histogram_metric)` in reference Prometheus PRESERVES
each histogram sample and joins the matching `target_info` (or explicit
second-argument) labels onto it — it does not drop to an empty vector and
it does not error.

This makes the fix shape different from the drop-semantics pattern
#2221/#2345/#2456/#2498 already closed: `lowerInfo` needs to recognise a
histogram-valued base (`lowerExpHistogramValuedShape`) and still perform
the label join over it (`chplan.InfoJoin` or an equivalent), analogous to
how `last_over_time`/`first_over_time` preserve rather than drop a
histogram sample. That is a real design question (does `chplan.InfoJoin`
already tolerate a histogram-shaped `Input`, or does it need widening?),
not a mechanical copy of the drop-semantics guard, so it is filed
separately rather than folded into #2498.

Found while rotating the rejection-parity catalogue's
`expHistogramSelectorRouting` divergence entry
[...]

Issue #2474, opened 2026-08-22
promql: timestamp() over a raw exp-histogram selector is hard-rejected


tsouza:
## Problem

`timestamp(demo_latency_exp_hist)` is hard-rejected by cerberus, while reference Prometheus
answers it — `timestamp(v)` ignores a sample's value entirely and reports the sample's own
timestamp, so it is defined for a native-histogram-valued argument exactly as it is for a
float-valued one (`promql/engine.go`'s `funcTimestamp` reads `Point.T`, never `Point.H`/`Point.F`).
Confirmed directly against the vendored reference engine
(`github.com/tsouza/prometheus`'s `promql/promqltest`):
`timestamp(<native-histogram-series>)` answers the sample's timestamp unchanged, not an error and
not an empty vector.

`lowerDateFn` (`internal/promql/date_fns.go`), which implements `timestamp()` alongside the other
date-component functions, calls `lower(c.Args[0], s, argCtx)` unconditionally with no
`lowerExpHistogramValuedShape` check beforehand, so a bare exp-histogram selector argument falls
through the generic dispatch and hits `expHistogramSelectorRouting`'s catch-all rejection instead
of answering with the sample's timestamp.

Representative query on current main:

```text
timestamp(demo_latency_exp_hist)
```

## Evidence

- Rejected by `internal/promql/lower.go:expHistogramSelectorRouting` — `lowerDateFn`
  (`internal/promql/date_fns.go`) routes its instant-vector argument straight through `lower()`
  with no histogram exemption, unlike the instant-math functions (`instant_fns.go`), the clamp
  family (#2345), `sort`/`sort_desc` (#2456), and `sort_by_label`/`sort_by_label_desc` (#2462).
- Confirmed by probing `promql.LowerAt` directly against `timestamp(demo_latency_exp_hist)`: it
  returns the `expHistogramSelectorRouting` rejection.
- This is the same rejection site (`expHistogramSelectorRouting`) previously rotated through #1801,
  #2087, #2189, #2224, #2225, #2226, #2331, #2339, #2342, #2345, #2443, #2456, and #2462 as each
  triggering shape got fixed — `timestamp()` is the next reachable trigger once #2462 lands, tracked
  at `test/rejection-parity/catalogue/internal__promql__lower.go.json`.

## Acceptance

- `timestamp(demo_latency_exp_hist)` answers with reference-matching semantics — the sample's own
  timestamp, independent of its value — or the rejection-parity catalogue continues to track it
[...]

PR #2517, opened 2026-08-24
fix(promql): drop exp-histogram samples for the date-component functions

### LINKED ISSUE #2498: promql: date-component functions (year/month/day_of_week/...) over a raw exp-histogram selector hard-reject instead of dropping samples
## Problem

While closing cerberus issue #2482 (`ts_of_first_over_time` / `ts_of_last_over_time` /
`ts_of_max_over_time` / `ts_of_min_over_time` over a raw exp-histogram selector), the
rejection-parity catalogue's `expHistogramSelectorRouting` entry needed rotating to the next
reachable trigger query now that all four #2482 functions succeed. Confirmed directly against
`promql.LowerAt` on the branch closing #2482: every date-component function other than
`timestamp()` (fixed by #2474) still hits the exact same `expHistogramSelectorRouting` catch-all
rejection over a bare exp-histogram selector:

```text
year(demo_latency_exp_hist)
month(demo_latency_exp_hist)
day_of_month(demo_latency_exp_hist)
day_of_year(demo_latency_exp_hist)
day_of_week(demo_latency_exp_hist)
days_in_month(demo_latency_exp_hist)
hour(demo_latency_exp_hist)
minute(demo_latency_exp_hist)
```

Verified directly against `tsouza/prometheus`'s `promql/functions.go` before filing (not assumed
from #2474's own text): every one of these eight functions routes through the shared
`dateWrapper` helper, which explicitly DROPS a histogram sample rather than computing a value
from it:

```go
for _, el := range vectorVals[0] {
    if el.H != nil {
        // Ignore histogram sample.
        continue
    }
    ...
}
```

So reference Prometheus answers an EMPTY vector for `year(<all-histogram-window selector>)` —
never an error — exactly the same "process only float samples, drop the rest" pattern #2221,
#2477 and #2480 already closed for other function families. Cerberus currently hard-rejects the
whole query instead via `expHistogramSelectorRouting`'s catch-all, which is a wrong-rejection bug,
not a semantics gap: `timestamp()` is the ONLY date-component function whose value is independent
of the sample's payload (`dateFnExpr`'s `"timestamp"` case in `internal/promql/date_fns.go`), so
#2474 deliberately scoped itself to just that one function
(`internal/promql/histogram_native_timestamp_test.go`'s own
`TestLower_ExpHistogram_Timestamp_OtherDateFnsStillRejectHistograms` pins the CURRENT rejecting
behavior of the other eight as this issue's own scope boundary, not as intended-forever
semantics).

## Acceptance

- `year()` / `month()` / `day_of_month()` / `day_of_year()` / `day_of_week()` /
  `days_in_month()` / `hour()` / `minute()`, each applied to a bare exp-histogram selector, answer
  an empty vector over an all-histogram window, matching reference's drop semantics — OR the
  rejection-parity catalogue continues to track the site explicitly under this issue until they
  do.
- Focused test coverage (unit + rejection-parity trigger) for whichever subset is implemented.

tsouza:
## Summary

- Applies Prometheus's `dateWrapper` semantics (drop histogram-valued samples,
  answer only over floats) to the eight date-component functions —
  `year`/`month`/`day_of_month`/`day_of_year`/`day_of_week`/`days_in_month`/
  `hour`/`minute` — which previously fell through to
  `expHistogramSelectorRouting`'s catch-all rejection instead of Prometheus's
  drop-and-answer-empty behavior. `timestamp()` is left untouched since it
  already answers over histogram arguments (#2474/#2478).
- Rotates the rejection-parity catalogue's `expHistogramSelectorRouting` entry
  trigger off `year(...)` (now lowers successfully) onto
  `info(demo_latency_exp_hist)`, and files the `info()` join/preserve
  divergence separately as #2509 since it needs a different fix shape.

Closes #2498

## Test plan

- [x] New direct coverage for all eight date-component functions plus a
      range-mode case (`internal/promql`)
- [x] Rewrote `TestLower_ExpHistogram_Timestamp_OtherDateFnsStillRejectHistograms`
      to assert the drop instead of the stale rejection
- [x] `TestRejectionTriggersExerciseSites` passes with the rotated trigger query
- [x] `just test`

### DIFF
--- a/internal/promql/date_fns.go
+++ b/internal/promql/date_fns.go
@@ -75,6 +75,25 @@ func lowerDateFn(c *parser.Call, s schema.Metrics, ctx lowerCtx) (chplan.Node, e
 		if node, ok, err := lowerTimestampOverExpHistogram(c.Args[0], s, ctx); ok {
 			return node, err
 		}
+	} else {
+		// Reference Prometheus routes year/month/day_of_month/day_of_year/
+		// day_of_week/days_in_month/hour/minute through a shared dateWrapper
+		// helper (promql/functions.go) that explicitly skips every sample
+		// whose H field is set before computing the date component from F —
+		// the same "process only float samples, drop the rest" rule
+		// simpleFloatFunc applies for abs/ceil/floor/round/etc (issue
+		// #2221) and clamp/sort already mirror (issues #2345, #2456).
+		// These eight never routed their argument through
+		// lowerExpHistogramValuedShape, so a histogram-valued argument fell
+		// through to the generic lower() dispatch below and hit
+		// expHistogramSelectorRouting's catch-all rejection instead of
+		// Prom's drop-and-answer-empty semantics (cerberus issue #2498).
+		if hist, ok, err := lowerExpHistogramValuedShape(c.Args[0], s, ctx); ok {
+			if err != nil {
+				return nil, err
+			}
+			return dropExpHistogramSamples(hist, s), nil
+		}
 	}
 
 	// The argument is lowered under an ARGUMENT ctx rather than the caller's
--- a/internal/promql/histogram_native_date_fns_test.go
+++ b/internal/promql/histogram_native_date_fns_test.go
@@ -1,0 +1,98 @@
+package promql_test
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/prometheus/prometheus/promql/parser"
+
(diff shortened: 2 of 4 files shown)


tsouza:
Closing as a duplicate of #2508, which already merged this exact fix (drop-semantics for the eight date-component functions, closing #2498) — `mergeStateStatus: CONFLICTING` confirms main already has this change. No action needed here.

Issue #2534, opened 2026-08-24
promql: a drop-family binop nested as another binop's operand still hard-rejects


tsouza:
## Problem

While fixing issue #2528 (`lowerRoot`'s exp-histogram "drop" recognizers never
retrying on a nested sub-expression), rotating the rejection-parity
catalogue's `expHistogramSelectorRouting` entry forward required probing the
guard chain broadly again. #2528's own fix threads a new reusable
`lowerExpHistogramDroppingShape` recognizer into every EXISTING per-callsite
opt-in a wrapper already has for the "preserve" family (an aggregation of any
operator, `limitk`/`limit_ratio`, `sort`/`sort_desc`/`sort_by_label`, the
unary math functions, `clamp`/`round`, `scalar()`, `info()`,
`label_replace()`/`label_join()`, `timestamp()`, the eight date-component
functions) — so a drop-family binop (`demo_latency_exp_hist + 0`) now
resolves correctly as the direct argument of any of those.

The next reachable trigger is a different, narrower gap within the SAME
root cause family: a drop-family binop nested as an OPERAND of ANOTHER
binary expression, rather than as a wrapper's direct argument:

```text
(demo_latency_exp_hist + 0) * 2
(demo_latency_exp_hist + 0) + 5
sum(demo_latency_exp_hist + 0) + 1
(demo_latency_exp_hist + 0) or (demo_latency_exp_hist + 0)
```

all still hard-reject via `expHistogramSelectorRouting`'s catch-all, even
though the identical inner shape lowers cleanly on its own (as does
`abs(demo_latency_exp_hist + 0)` and every other wrapper #2528 fixed).

## Root cause

`internal/promql/lower.go`'s generic binary-expression lowering
(`lowerBinary` and its `lowerVectorScalar`/`lowerVectorVector` helpers in
`internal/promql/binary.go`) recurse into BOTH operands via the plain
`lower()` dispatcher, with no per-operand opt-in analogous to the ~12
callsites #2528 just patched. Every fix so far — both the "preserve" family
(`lowerExpHistogramValuedShape`) and now the "drop" family
(`lowerExpHistogramDroppingShape`) — has been threaded into a *wrapper's own
argument-lowering callsite* (a Call's single vector argument, an
AggregateExpr's `Expr`), never into the generic binary-expression recursion
that lowers `BinaryExpr.LHS`/`BinaryExpr.RHS` themselves. So an OUTER binop
whose operand happens to be a nested histogram-dropping (or
[...]

PR #2535, opened 2026-08-24
fix(promql): retry the exp-histogram drop family on a nested sub-expression

### LINKED ISSUE #2528: promql: lowerRoot's exp-histogram 'drop' recognizers never retry on a nested sub-expression
## Problem

While fixing #2518 (`limitk`/`limit_ratio` hard-rejecting a bare exp-histogram
argument), rotating the rejection-parity catalogue's `expHistogramSelectorRouting`
entry forward required probing the guard chain broadly. The next reachable trigger
is not another single missing per-function check — it is a structural gap that
spans essentially every aggregate and function in `internal/promql`:

```text
sum(demo_latency_exp_hist + 0)
topk(3, demo_latency_exp_hist + 0)
count(demo_latency_exp_hist + 0)
sort(demo_latency_exp_hist + 0)
limitk(3, demo_latency_exp_hist + 0)
```

all still hard-reject via `expHistogramSelectorRouting`'s catch-all:

```text
promql: "demo_latency_exp_hist" is an exponential histogram metric; only a bare
"demo_latency_exp_hist" selector, sum()/avg()/count() over one,
rate()/increase()/delta()/irate()/idelta()/sum_over_time()/avg_over_time()/resets()/
changes() over one, histogram_quantile(), histogram_count(), histogram_sum(), and the
"demo_latency_exp_hist_count"/"demo_latency_exp_hist_sum" companion selectors are
supported
```

even though the identical shape succeeds at the query root: `demo_latency_exp_hist
+ 0` on its own lowers cleanly (to an empty result — reference Prometheus's ADD
between a histogram-valued vector and a scalar is an incompatible-types combination
that drops the sample and emits an info annotation; cerberus reproduces that via
`expHistogramDroppingScalarBinop` / `lowerExpHistogramScalarBinop(..., drop=true)`).

## Root cause

`internal/promql/lower.go`'s `lowerRoot` is the one function that tries the full
histogram-recognizer sweep — both the "preserve" family
(`lowerExpHistogramValuedShape`, reused by many per-function fixes) and the
"drop" family (`expHistogramDroppingScalarBinop`, `expHistogramDroppingVectorBinop`,
`expHistogramDroppingHistogramBinop`, `droppingAggregationOverExpHistogram`) —
against an expression. It is called exactly twice in the whole codebase, both from
the two public entry points (`LowerAt`, `LowerAtRangeOpts`), and never recursively
from within itself or from any nested lowering.

Every fix so far that taught a specific function to accept a histogram-valued
*argument* (`info()` #2509, `sort_by_label()` #2462, `scalar()` #2515, `limitk`/
`limit_ratio` #2518, and others) did so by having that function call
`lowerExpHistogramValuedShape` directly on its own argument before falling back to
the plain `lower()` dispatcher — a manual, per-callsite opt-in to the "preserve"
half of `lowerRoot`'s sweep. None of them needed the "drop" half, because a scalar/
vector histogram-dropping binop was never their argument shape in the fixtures that
motivated those fixes.

The "drop" family has no equivalent per-callsite helper at all (nothing like
`lowerExpHistogramValuedShape` exists for it) and is wired into `lowerRoot` only.
So any wrapper — `sum(...)`, `topk(...)`, `count(...)`, `sort(...)`, `limitk(...)`,
an instant-math function, a range function, `label_replace`, etc. — whose own
lowering calls the plain `lower()` dispatcher on an inner expression that happens to
be a histogram-dropping binop (`histogram_metric + <scalar>`, `histogram_metric ==
bool <vector>`, `histogram_metric_a - histogram_metric_b`, …) never gets a second
chance at that recognition. The nested sub-expression falls straight through to
`lowerVectorSelector` → `expHistogramSelectorRouting`'s catch-all rejection instead
of answering reference Prometheus's actual "drop the sample, evaluate to empty"
semantics.

## Fix shape

This is one root cause, not N per-function leaves — the intended fix is a single
reusable recognizer/lowering pair for the "drop" family (mirroring
`lowerExpHistogramValuedShape`'s existing shape for the "preserve" family), threaded
into the same per-callsite opt-in points the "preserve" family already uses, so a
nested histogram-dropping binop composes correctly under any wrapper rather than
needing another one-off fix per consumer function. Confirm the exact scope by
probing the guard chain the same way #2509/#2462/#2515/#2518 each did before
committing to a design.

## Rejection-parity catalogue

`test/rejection-parity/catalogue/internal__promql__lower.go.json`'s
`internal/promql/lower.go:expHistogramSelectorRouting#bf746b59` divergence entry is
rotated (in the PR for #2518) to `trigger_query: "sum(demo_latency_exp_hist +
0)"` and `tracking_issue` pointing at this issue.

tsouza:
## Summary

`internal/promql/lower.go`'s `lowerRoot` was the only place that ran the exp-histogram
"drop" family's four recognizers (`droppingAggregationOverExpHistogram`,
`expHistogramDroppingScalarBinop`, `expHistogramDroppingHistogramBinop`,
`expHistogramDroppingVectorBinop`) — always against the whole query, never against a
nested sub-expression. Unlike the "preserve" family (`lowerExpHistogramValuedShape`,
already reused directly by `info()`, `sort_by_label()`, `scalar()`,
`limitk`/`limit_ratio`, and others at their own callsite), the drop family had no
per-callsite opt-in helper at all, so a histogram-dropping binop nested as a wrapper's
argument — `sum(demo_latency_exp_hist + 0)`, `topk(3, ...)`, `count(...)`, `sort(...)`,
`limitk(...)`, a math function, `label_replace(...)`, … — hard-rejected via
`expHistogramSelectorRouting`'s catch-all instead of composing to reference
Prometheus's real "drop the sample, evaluate to empty" answer.

## Design

Adds `lowerExpHistogramDroppingShape` (`internal/promql/histogram_native_dropping_shape.go`),
mirroring `lowerExpHistogramValuedShape`'s exact `(chplan.Node, bool matched, error)`
contract. Unlike that sibling, whose result stays in the thirteen-column histogram row
shape, this one's result is always reprojected to the canonical four-column float-Value
row shape (the same shape `dropExpHistogramSamples` already builds for the preserve
family): "drop" answers with an ordinary (currently empty) float vector under reference,
so any further wrapper — histogram-aware or not — sees an ordinary, currently-empty
float plan and runs its perfectly normal lowering over it. That is what lets an
arbitrarily deep stack of wrappers compose (`abs(sum(demo_latency_exp_hist + 0))`,
`sort(topk(3, demo_latency_exp_hist + 0))`, `sort(limitk(3, demo_latency_exp_hist + 0))`)
without a bespoke recognizer per wrapper.

Two new recognizer/lowering pairs live inside it:

- `aggregationOverExpHistogramDroppingShape` — ANY aggregation operator (not the
  restricted min/max/stddev/stdvar/quantile/topk/bottomk set `droppingAggregationOverExpHistogram`
  already covers for a histogram-**valued** argument) wrapping a drop-family argument.
  Every PromQL reducer answers empty over an empty input, so this is op-agnostic;
  `limitk`/`limit_ratio` are excluded since they route through their own,
  root-**and**-nested-reachable opt-in (`lowerLimitKInput`, below) instead.
- `labelCallOverExpHistogramDroppingShape` — `label_replace`/`label_join` wrapping a
  drop-family argument, the sibling of the existing preserve-family `labelCallOverExpHistogram`.

Threaded into every existing "try preserve, fall back to `lower()`" callsite:
`sort`/`sort_desc`/`sort_by_label`, the unary math functions, `clamp`, `round(v, n)`,
`scalar()`, `info()`, `timestamp()`, the eight date-component functions, and
`lowerLimitKInput`. Plus a new entry in `lowerRoot`'s own dispatch chain, appended
*after* the four existing raw drop-family dispatches so nothing already matching them
changes shape.

`histogram_native_float_fn.go`'s `dropExpHistogramSamples` is split into a shared
`floatShapedExpHistogramDrop` projection helper so the new code reprojects an
already-filtered plan without adding a second, redundant constant-false `Filter`.

## What remains

A drop-family binop nested as the operand of **another** binary expression — e.g.
`(demo_latency_exp_hist + 0) * 2`, `sum(demo_latency_exp_hist + 0) + 1` — still
hard-rejects. `lowerBinary`/`lowerVectorScalar`/`lowerVectorVector`
(`internal/promql/binary.go`) recurse into both operands via the plain `lower()`
dispatcher with no per-operand opt-in analogous to the callsites this PR threads —
that needs an opt-in inside the *core* recursive binary-expression lowering itself,
a materially different and larger change than another single-argument callsite, so
it's tracked separately: #2534.

The rejection-parity catalogue's `expHistogramSelectorRouting` entry
(`test/rejection-parity/catalogue/internal__promql__lower.go.json`) rotates its
`trigger_query` to `(demo_latency_exp_hist + 0) * 2` and `tracking_issue` to #2534,
matching that remaining gap.

## Testing

- `internal/promql/histogram_native_dropping_shape_test.go` — new unit tests covering
  every threaded wrapper (aggregations of every operator, `limitk`/`limit_ratio`,
  `sort`/`sort_desc`/`sort_by_label`, math functions, `clamp`, `round`, `scalar()`,
  `info()`, `label_replace`/`label_join`, `timestamp()`, a date-component function,
  and two-level composition), instant and range mode, aggregation-parameter validation
  still firing before the drop (`topk(NaN, ...)`, `limitk(NaN, ...)`,
  `count_values("", ...)`), and the exact plan shape.
- Two new TXTAR spec fixtures (`test/spec/promql/sum_of_dropping_binop_exp_histogram.txtar`,
  `sort_of_dropping_binop_exp_histogram.txtar`) pin the chDB roundtrip for a
  representative aggregation-based and Call-based wrapper; both need the CI golden
  regen to fill in their `sql`/`chplan`/`args` sections.
- `go build ./...` and `go test ./internal/promql/...` pass locally (the two new
  fixtures fail pending golden regen, as expected before that CI step runs).

Closes #2528


tsouza:
Re-triggering CI: GitHub Actions never created a check-suite for the latest push (34+ min with zero runs while other PRs' pushes triggered normally in the same window). Reopening to force a fresh pull_request synchronize event.

PR #2622, opened 2026-08-25
test(promql): cover the vector-vector operand path #2534's nested-binop fix added

tsouza:
## Summary
- An adversarial review of #2538 found `internal/promql/histogram_native_dropping_nested_binop_chdb_test.go`'s original four cases all routed through `lowerScalarBinopOperand` / `lowerVectorSetOpOperand`, leaving `lowerVectorVectorOperand` — the function #2534 added specifically for a drop-family binop nested as a real vector-vector op's operand — completely untested, and the file's own header comment falsely claimed one of the four already covered it.
- Adds two non-redundant cases (verified by tracing `lowerBinary`'s actual dispatch, and by neutering each opt-in in turn and confirming lowering regresses to a hard rejection): a comparison-shaped drop-family leaf nested under a real vector-vector `*` reaches `lowerVectorVectorOperand`'s own match, and a histogram-histogram compare nested under a scalar `+` reaches `lowerScalarBinopOperand`'s preserve-family half.
- Fixes the header comment to describe each case's real routing.

## Test plan
- [ ] CI (unit + spec + chdb-tagged lanes)

### DIFF
--- a/internal/promql/binary.go
+++ b/internal/promql/binary.go
@@ -141,11 +141,11 @@ func lowerVectorVector(b *parser.BinaryExpr, s schema.Metrics, op chplan.BinaryO
 		// emitter's "many" aggregation handles by construction).
 	}
 
-	left, err := lower(b.LHS, s, ctx)
+	left, err := lowerVectorVectorOperand(b.LHS, s, ctx)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


tsouza:
Closing — branch predates significant drift on main; lowerVectorVectorOperand no longer exists in its original shape, so this test coverage is stale. Scratched rather than rebased.

Issue #2495, opened 2026-08-23
classic-histogram fan-out merge leaks a literal non-finite ExplicitBounds value into the quantile clamp


tsouza-in-automode:
## Summary

The classic-histogram `histogram_quantile` merge path builds its per-`le`
canonical bound union straight from each row's own raw `ExplicitBounds`
values, with no `isFinite` filter. When a row's `ExplicitBounds` carries a
literal `+Inf` or `NaN` value — malformed but real OTLP data can produce
either — that value survives into the union that becomes the node's own
output `ExplicitBounds`, and `emitHistogramQuantile`'s "rank falls in the
overflow bucket" clamp
(`internal/chsql/histogram_quantile.go:191`, `highestBound := Subscript(w.bounds(), Call("length", w.bounds()))`)
can then return that garbage `+Inf`/`NaN` value as the answered quantile
instead of the highest genuine **finite** bound.

## Where

- `internal/promql/histogram_quantile.go:1412` —
  `classicBucketUnionBoundsExpr()`: `arraySort(arrayDistinct(arrayFlatten(hqAggBoundsListAlias)))`
  over the raw `groupArray(ExplicitBounds)` list, no `isFinite` filter.
  Feeds both the per-series window stage (`classicBucketWindowStage` /
  `classicBucketWindowReshape`, `internal/promql/histogram_quantile_window.go`)
  and the cross-series merge (`classicBucketMergeShaping.reshape`,
  `internal/promql/histogram_quantile.go`), so both stages of the ordinary
  fan-out `histogram_quantile(phi, <agg> by(le) (rate/sum_over_time(<bucket>[range])))`
  path are affected.
- `internal/promql/histogram_quantile.go:1462` —
  `classicBucketRowCumulativeExpr()`: pairs a row's raw bounds against its
  counts with no non-finite guard either, so a row's own literal `+Inf`
  entry gets folded in as an ordinary finite rung.
- `internal/chsql/histogram_quantile.go:191` — `histogramQuantileValueFrag`'s
  `highestBound` clamp reads `_cerb_hqc_bounds[length(_cerb_hqc_bounds)]`
  verbatim, so once a non-finite value has leaked into the coalesced
  `ExplicitBounds` array as its own last entry, the clamp branch (taken
  whenever the interpolation rank lands in the `+Inf` overflow bucket)
  returns it directly.

## Repro (verified against real chDB in a worktree, seed removed afterward)

[...]

PR #2624, opened 2026-08-25
fix(promql): pre-release audit fixes batch (findings 1-14, A-E)

tsouza:
## Summary

Batch of pre-release-audit fixes scoped to `internal/promql/`, dispatched as `audit-fix/promql-delta` for consolidation.

**HIGH**
- Finding 1: `absent()`/`absent_over_time()` over a regex/negated `__name__` matcher now correctly detects a live exponential-histogram metric's presence instead of silently reporting it ABSENT.
- Finding 2: `guardKeysOnTimestamp` now handles `*chplan.VectorSetOp`/`*chplan.NaryVectorSetOp`, fixing `label_replace`/`label_join` duplicate-labelset detection over an instant-mode `or`.
- Finding 3 (+4): group_left()/group_right() histogram binop merge now carries the same bucket-width budget guard the one-to-one path has; unrelated const-reuse cleanup bundled in.
- Finding A: nested-subquery outer range functions (`max_over_time(rate(m[1m])[5m:30s])[1h:5m]`-shaped) now guard against a histogram-valued inner instead of building an unemittable RangeWindow.
- Finding B/C: a further wrapper (math fn, `scalar()`, `timestamp()`, etc.) composed over `sort_by_label(<mixed or>, label)` no longer panics or silently mis-answers — narrows to float-only rows before forwarding.
- Finding D: float-domain `histogram_quantile` ladder now pairs `BucketCounts` with `ExplicitBounds` after filtering non-finite bounds (was desyncing by more than one element for a malformed multi-infinite-bound row).
- Finding E: bare classic `histogram_quantile(phi, m_bucket)` (no `le` matcher) now filters non-finite storage bounds the same way the aggregated path already did (#2495) — was answering `-Inf` verbatim for a malformed row.

**MEDIUM**
- Findings 5/6: named `mixedDiscriminatorFloat`/`mixedDiscriminatorHistogram` consts replace ~14 bare `0`/`1` literals across the mixed float/histogram `or` composer family.
- Finding 7: `(a or b) * (hist_x and other_selector)`-shaped plain operand now rejected at the same early, shared site every other unattempted histogram/mixed combination is, instead of a late recognizer-specific error.
- Finding 8: investigated thoroughly — **does not reproduce** on current `main` (already fixed by #2621). No production change; added a pinning regression test only.
- Finding 10: deduped the identical 3-line `ReturnBool` guard across three mixed-or composers into `shadowResolveFloatArmChecked`.
- Finding 11: documented the (harmless, dead-in-practice) `ZeroThreshold` guard invariant at all three sites rather than removing it.

**LOW**
- Findings 13/14: corrected two stale doc comments.
- Findings 9, 12, 15-18: skipped per the audit's own deprioritization (cosmetic/architectural, low value).

## Test plan

- [x] Every fix has a dedicated regression test, each verified to catch the reverted regression (stash the fix, confirm the test fails; restore, confirm it passes).
- [x] `go test ./internal/promql/...` — green except 8 pre-existing golden-fixture text mismatches (`histogram_quantile_classic_*`, `exp_histogram_binop_add_group_*`, `exp_histogram_set_op_or_mixed_label_*`) that need `just update-golden` — these are the *expected* SQL-shape changes from findings 2/3/D/E, hand-verified to be text-only (no value changes) for well-formed data.
- [x] New/changed chDB-tagged tests run individually — all pass, all verified to catch their regression.
- [ ] `just update-golden migration parity solver promql chsql codegen cardinality` — left for CI per coordinator direction.


tsouza:
While verifying the rejection-parity catalogue regen for finding 10 (audit-fix/promql-delta, commit f90eb0dcf), ran the full `internal/promql` package under `-tags chdb` (not per-fixture) and found two real, reproducing failures beyond the already-known golden-fixture text drift — filed as #2625. Not caused by the catalogue-only commit (touches zero `.go` files); appears related to findings D/E's finite-bounds-filter work, surfacing only when the whole package runs together rather than fixture-by-fixture. Worth resolving before this PR merges.

PR #2664, opened 2026-08-27
debug: per-attempt timing diagnostics for route-memo-activation (throwaway)

tsouza:
Throwaway diagnostic branch — not intended to merge. `route-memo-activation` has now timed out twice on main's push-triggered `chaos` job (once at ~123s against a 120s budget, once at ~182s against a widened 180s budget), both times with the very next corroboration check succeeding within 0.5s of the deadline. Adds per-attempt status/timing/consecutive-422-streak logging to the poll loop so a real chaos run reveals whether this is per-attempt variance interrupting the consecutive-failure streak, or a systematic pattern, before committing to any further timeout change.

Will close this PR once the investigation concludes, either with a root-caused fix on its own branch or a decision that the current 180s budget is acceptable.

### DIFF
--- a/.github/scripts/chaos-run.mjs
+++ b/.github/scripts/chaos-run.mjs
@@ -347,12 +347,10 @@ function kubectlOut(args) {
 }
 
 // dropClickHouseReadCaches clears the mark cache and uncompressed-block
-// cache server-side. See route-memo-activation's own use for why: a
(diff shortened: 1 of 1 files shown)


tsouza:
Throwaway diagnostic branch, purpose served — informed the real fix (real chDB-backed root-cause investigation, currently landing on branch fix/route-memo-activation-margin / PR #2669). Closing without merging.

Issue #2688, opened 2026-08-27
solver: an empty complementary DELTA arm vetoes routing for the whole classic-histogram union


tsouza:
Part of epic #2677. Found by tracing why a live-edge dashboard panel still
refuses to route after #2687's fanout fix.

## The veto

The classic-histogram range path lowers, on any schema declaring an
`AggregationTemporality` column, to a **complementary two-arm union**
(`internal/promql/histogram_quantile_classic_native.go`):

```
UnionAll{
  RangeBucketGridNative  (cumulative rows — the native ladder)
  RangeBucketFanout      (DELTA rows — the array fold)
}
```

`FanoutClassicHistogramWindowLowerer` sets `PeakIndependentOfGrid = true` on
**every** fan-out it builds, including this delta arm. The solver reads that
as `AnchorGridDivides() == false`, sets `sawIndivisibleAnchorGrid`, and
refuses the WHOLE plan with `ReasonAnchorGridIndivisible`.

That flag is well-founded on its own terms — its doc carries a real
measurement (slicing a fan-out spine 12 ways recovered 8.7% of peak while
costing 23× the ClickHouse work), and it is deliberately a policy declaration
rather than a threshold.

## Why it is wrong here

The two arms are gated on **complementary predicates**: the native arm takes
`AggregationTemporality != DELTA`, the fan-out arm takes `== DELTA`. They
partition the rows; they do not both process them.

On an all-cumulative deployment — the overwhelmingly common case, and the one
in the #2677 production incident — the delta arm matches **zero rows**. Its
peak is nil. Yet its policy declaration vetoes routing for the native arm,
whose peak genuinely does divide with the anchor grid.

So the plan is refused on the cost characteristics of an arm that does no
work.

## Observed

Real ClickHouse 26.6, 253 series, width 68, live-edge windows, after #2687's
fanout fix:

| window | before that fix | after |
|---|---|---|
| 60 min | `below-threshold` | `below-threshold` (correct — under the anchor floor) |
| 180 min | `below-threshold` | **`anchor-grid-indivisible`** |
| 360 min | `below-threshold` | **`anchor-grid-indivisible`** |

The cost misclassification is gone; this veto is what remains.

## Options, none obviously right

1. **Make the veto arm-local rather than plan-wide** for a union of
   complementary arms — the signal walk currently has no notion of "these arms
   partition the input", so this needs one.
2. **Do not set `PeakIndependentOfGrid` on an arm the native lowerer builds as
[...]

Issue #2677, opened 2026-08-27
promql/solver: classic-histogram quantile-of-rate memory scales with time-range width (epic)


tsouza:
## Summary (epic)

`histogram_quantile(q, <agg> by (le)(rate(<classic_bucket>[5m])))` over a wide
time range genuinely exhausts per-query memory on real ClickHouse — confirmed
live on a production stack right after the v1.17.0/#2666 deploy, and
reproduced on a faithful local replica (real CH 26.6, prod's real memory cap
and optimization set, synthetic data matching the measured production shape:
253 series, bucket width 68, ~84.5k rows/6h). Real cliff: between **90 and
105 anchors** at that shape under a 6 GiB per-query cap; extrapolated ~20+
GiB for a full 6h/360-anchor panel query.

Two distinct observed failure modes, same root:

| shape | window | outcome |
|---|---|---|
| `sum by (le)` (`agg=1;rbf;rbn`) | 6h | 422/502 — rejected by the axis2 density guard (~97.6% of the 400M cap by its own formula) |
| `sum by (le,uri)` (`agg=3;rbf;rbn`) | 6h | 422 — **passed** the guard, then real ClickHouse `MEMORY_LIMIT_EXCEEDED` downstream |

This issue is now the epic coordinating the fix; the deep-dive below
corrects the original root-cause analysis and maps every finding to a
sub-issue.

## Corrected root-cause analysis

The originally-traced `groupArray`/`_hq_window_ladder` SQL came from
**offline** `cerberus migrate explain`, which cannot resolve the live-CH
capability probe — so it showed the **fallback array-fold path**
(`internal/promql/histogram_quantile_window.go`), not what a capable
production server executes. On CH ≥ 25.9 with `ts_grid_histogram` resolved
(confirmed in the reporting stack's boot log), the shape lowers to
**`chplan.RangeBucketGridNative`** (`internal/chsql/range_bucket_grid_native.go`)
— the native ladder built for #2408/#2523, measured **28× faster / 26× less
memory / 129× less CPU** than the array fold it replaced.

The native path is enormously better, but its memory still scales with the
requested window, on two axes its own guard
(`internal/chsql/range_bucket_grid_native_bound.go`) models as
`costUnits = (groups × anchors) + (rawRows × width²)`:

1. **Level 1 grid states** — one `timeSeriesRateToGrid` +
   `timeSeriesResetsToGrid` state per (series, `le` rung), each holding
[...]

PR #2690, opened 2026-08-27
perf(promql): decide the DELTA arm's peak flag at construction, not by a walk

tsouza:
## Summary

#2689's fix for #2688 cleared `PeakIndependentOfGrid` on the DELTA complement
arm with a `chplan.Walk` over the returned subtree. That walk runs on **every**
classic-histogram lowering, and the cost is not theoretical:

```
Killed: 0, Lived: 0, Not covered: 16
Timed out: 295, Not viable: 0, Skipped: 0
Test efficacy: 0.00%
##[error]mutation phase executed zero mutants after full fallback
```

The `internal/promql` gremlins phase went from passing on main (3 consecutive
runs checked) to executing zero mutants — 295 timed out. The timeouts are
spread across the phase's other files (`lower_strategy.go` and friends), which
is the signature of the package getting uniformly slower rather than one
mutant hanging.

## Fix

Make the flag a **parameter of the fan-out construction** instead of building
the node with it and correcting afterwards. Same behaviour, no per-lowering
traversal, and the decision now reads at the one place it is made.

The `ClassifyHistogramWindowLowerer` interface cannot express the parameter,
so the direct path is taken only for a concrete
`FanoutClassicHistogramWindowLowerer`; any other implementation keeps its own
behaviour unchanged.

## Why this shipped in the first place

Worth recording, since the mechanism is generic: a post-hoc "build it then fix
it" pass reads as free at the call site, and its cost is invisible to
correctness tests — `go test ./internal/promql/` was green throughout. Only
mutation testing, which re-runs the suite once per mutant, amplified a small
per-lowering cost into a phase-wide timeout. The structural form (decide at
construction) is both faster and clearer, and was available from the start.

## Test plan

- [x] `go test ./internal/promql/ ./internal/solver/` green.
- [x] `chplan.Walk` no longer appears in the lowering path — the one remaining
      textual reference is the doc comment explaining why it was removed.
- [x] Behaviour unchanged: the complement arm is still built without
      `PeakIndependentOfGrid`, which is what #2688 needed.

Refs #2677

Refs #2688

### DIFF
--- a/internal/promql/histogram_quantile_classic_native.go
+++ b/internal/promql/histogram_quantile_classic_native.go
@@ -66,6 +66,23 @@ type FanoutClassicHistogramWindowLowerer struct{}
 // LowerClassicHistogramWindow builds the RangeBucketFanout + two-Project
 // reshape the classic-histogram range path has always emitted.
 func (FanoutClassicHistogramWindowLowerer) LowerClassicHistogramWindow(in classicHistogramWindowInput) chplan.Node {
+	return lowerClassicHistogramFanout(in, true)
+}
+
+// lowerClassicHistogramFanout is LowerClassicHistogramWindow with the
+// PeakIndependentOfGrid declaration made explicit rather than unconditional.
+//
+// peakIndependentOfGrid is true for the fan-out-ONLY lowering, where the
+// fan-out is the whole plan and its route-A cost was measured (see below), and
+// false for the DELTA complement the native lowerer builds, which is a
+// different construction site seeing a subset of rows and has never been
+// measured (#2688).
+//
+// Deciding it at construction rather than clearing it afterwards is not a
+// style choice: a post-hoc chplan.Walk over the returned subtree runs on every
+// classic-histogram lowering, and that cost is real enough to push the
+// internal/promql mutation phase past its per-mutant timeout.
+func lowerClassicHistogramFanout(in classicHistogramWindowInput, peakIndependentOfGrid bool) chplan.Node {
 	fanout := buildHistogramBucketFanout(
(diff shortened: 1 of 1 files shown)


tsouza:
## Correction: this did not fix the gremlins phase, and my diagnosis was wrong.

I attributed the `internal/promql` mutation-phase failure to the `chplan.Walk`
this PR removed. **That was wrong.** With the walk gone, the phase produces
byte-identical numbers:

```
Killed: 0, Lived: 0, Not covered: 16
Timed out: 295, Not viable: 0, Skipped: 0
Test efficacy: 0.00%
```

Same 295 timeouts, same 16 uncovered — before and after. The walk was not the
cause.

Where the timeouts actually are, which I should have checked first:

| file | timed-out mutants |
|---|---|
| `binary.go` | 70 |
| `info_fn.go` | 59 |
| `lower_strategy.go` | 43 |
| `histogram_native_mixed_or_math_fn.go` | 40 |
| `histogram_native_count_present_over_time.go` | 31 |
| `histogram_quantile_classic_native.go` | 23 |

Spread across files this PR never touched, with the file I changed a minor
contributor. That is the signature of the whole package's test suite getting
slower and every mutant exceeding its per-mutant budget — not of one hot path.

The phase passed on `main` for three consecutive runs before #2689 and has
failed since, so the regression is real and belongs to that merge; it is
tracked properly now rather than left attached to a fix that does not address
it.

**This PR still stands on its own merits** — deciding the flag at construction
is simpler and cheaper than building the node and correcting it, and removes a
per-lowering traversal — but it should not have been described as fixing the
mutation phase, and the merge should not be read as having fixed it.

Issue #2613, opened 2026-08-25
promql: absent()/limitk()/limit_ratio()/unary minus+plus still reject a mixed float/histogram or


tsouza-squid:
## Problem

`internal/promql/binary.go:lowerVectorSetOp#4924e0ba`'s catch-all ("'or' between a
float-valued and a histogram-valued operand is not supported") still rejects several more
functions wrapping a mixed float/histogram `or` (cerberus issue #2330: `<hist> or
<float>`), after cerberus issue #2611 composed `timestamp()`/`scalar()`/
`sort_by_label()`/`sort_by_label_desc()` over the same shape.

## Evidence

A systematic sweep of PromQL's function table against
`demo_latency_exp_hist or histogram_quantile(0.5, demo_latency_exp_hist)` found these still
rejecting with the catch-all above (verified live against `main` at the time of filing, via
`internal/promql.LowerAt`):

```promql
absent(demo_latency_exp_hist or histogram_quantile(0.5, demo_latency_exp_hist))
limitk(1, demo_latency_exp_hist or histogram_quantile(0.5, demo_latency_exp_hist))
limit_ratio(0.5, demo_latency_exp_hist or histogram_quantile(0.5, demo_latency_exp_hist))
-(demo_latency_exp_hist or histogram_quantile(0.5, demo_latency_exp_hist))
+(demo_latency_exp_hist or histogram_quantile(0.5, demo_latency_exp_hist))
```

All five share the identical root cause: none of `absent()`, `limitk()`/`limit_ratio()`, or
unary `-`/`+` has a mixed-`or` recognizer wired into the same
`histogram_native_mixed_or_*.go` composition family cerberus issues #2346/#2449/#2555/
#2571/#2577/#2578/#2587/#2595/#2600/#2605/#2609/#2611 already built for their siblings.
Each of these three functions already composes over the NON-mixed all-histogram shape
(`absent(demo_latency_exp_hist)` — cerberus issue #2443; `limitk`/`limit_ratio` — cerberus
issue #2585; unary minus/plus — cerberus issue #2593) — only the mixed-`or` argument shape
is the gap.

## Background

This entry has rotated through a long chain of composability fixes this session
(#2555→#2571→#2578→#2581→#2595→#2600→#2605→#2609→#2611, each closing the shape that most
recently triggered it and exposing the next). The PR closing #2611 rotates the trigger
forward to `absent(...)` above.

## Acceptance

[...]

PR #2721, opened 2026-08-29
fix(spec): give the 4 mixed-or parity fixtures a real chDB round-trip

tsouza:
## Summary

PR #2711 (issue #2613) enrolled `absent_mixed_or.txtar` / `limit_ratio_mixed_or.txtar` /
`unary_minus_mixed_or.txtar` / `unary_plus_mixed_or.txtar` in the PromQL parity corpus with a
`parity:` section, but none of them carried the `seed:`/`expected_rows:` opt-in `RunParity`
requires — the parity check reads the seeded rows back out of chDB rather than trusting a
hand-authored expectation, so a `parity:` section with no executable round-trip is a hard failure,
not a silent no-op.

Caught by the v1.18.1 release PR's `roundtrip-promql-shard`:

```
fixture absent_mixed_or carries a `parity:` section but no executable round-trip. The parity
check reads the seeded rows back out of chDB, so it needs the same `seed:` + `expected_rows:`
opt-in RunRoundTrip needs.
```

## Fix

Adds a self-contained `seed:` (one `otel_metrics_exponential_histogram` row) to each of the four
fixtures, matching the sibling `histogram_quantile_mixed_or.txtar` pattern, and regenerates
`expected_rows:` from real chDB execution via `GOLDEN_UPDATE=1` rather than a hand-derived guess.

`limitk_mixed_or.txtar` is unaffected — it correctly uses `parity_exempt` (nondeterministic
selection, same reason `limitk_by_job_truncates.txtar` already does), which `RunParity` never
requires a round-trip for.

## Verification

- `go test -tags chdb -run '^TestLower$/^(absent_mixed_or|limit_ratio_mixed_or|unary_minus_mixed_or|unary_plus_mixed_or|limitk_mixed_or)$' ./internal/promql/` — all 5 pass.
- `just update-golden parity solver promql cardinality` — clean, only the 4 fixture files plus
  their new cardinality-baseline entries changed.
- `forbid-sql-raw`, `forbid-skip` — clean.

Closes nothing on its own; unblocks the v1.18.1 release PR (#2719), which will pick this up once
merged.

### DIFF
--- a/test/perf/cardinality-baseline/promql/absent_mixed_or.json
+++ b/test/perf/cardinality-baseline/promql/absent_mixed_or.json
@@ -1,0 +1,11 @@
+{
+  "fixture": "promql/absent_mixed_or",
+  "fan_factor": 1,
+  "scan_rows": 2,
+  "peak_intermediate": 2,
+  "has_cross_join": false,
+  "has_array_join": false,
+  "has_recursive_cte": false,
+  "max_recursion_depth": 0,
+  "uncountable_levels": 0
+}
--- a/test/perf/cardinality-baseline/promql/limit_ratio_mixed_or.json
+++ b/test/perf/cardinality-baseline/promql/limit_ratio_mixed_or.json
@@ -1,0 +1,11 @@
+{
+  "fixture": "promql/limit_ratio_mixed_or",
+  "fan_factor": 1,
+  "scan_rows": 2,
+  "peak_intermediate": 2,
+  "has_cross_join": false,
+  "has_array_join": false,
+  "has_recursive_cte": false,
+  "max_recursion_depth": 0,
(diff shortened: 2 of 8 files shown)


tsouza:
Duplicate of #2720, which fixed the identical defect (the same four fixtures missing
`seed:`/`expected_rows:` for their `parity:` enrollment) and already merged to `main`
(commit history: c16ea7d06 → 8db6bbf78). Two agents working in parallel found the same
root cause independently — closing this one in favor of the already-merged fix rather
than reconciling two branches that do the same thing.

`roundtrip (promql)` is green on `main` post-#2720, and the v1.18.1 release PR (#2719)
picked up the fix via a branch update.

Issue #2797, opened 2026-08-31
Fixed-accumulator extrapolated rate/increase: temporality-bearing (DELTA/CUMULATIVE) counters


tsouza:
## Problem

Issue #2760's fixed-accumulator decomposition for the extrapolated window family (`rate()` / `increase()` / `delta()`, `chopt.FeatureFixedAccumulatorExtrapolated`) ships scoped to windows whose `RangeWindow.TemporalityColumn` is unset — i.e. Gauge-table input, non-OTel-temporality-tagged counter data, or the `Scan.UnionTables` cross-table read. A temporality-bearing counter (an OTel Sum-typed series carrying an explicit `AggregationTemporality`) keeps the unchanged `emitWindowedArrayExtrapolatedMatrix` array-fold emission for both its CUMULATIVE and DELTA arms.

This is the real-world-relevant population the fixed-accumulator fallback was partly motivated by: `nativeTSGridMatrixNode` already excludes any window with `TemporalityColumn != ""` from the native `timeSeriesRateToGrid` path unconditionally (no DELTA-temporality arm exists on that ClickHouse aggregate), so every temporality-bearing `rate()`/`increase()` query pays the full array-fold fan-out today, with or without #2760's fix.

## Why it was descoped from #2760

Closing the gap requires re-expressing two pieces #2760 intentionally left untouched:

1. **The CUMULATIVE/DELTA runtime branch.** `CounterOrDeltaSum`'s DELTA arm (`arraySum(values) - first_val`, i.e. no reset-correction walk at all) is a straightforward decomposition — `sum(value) - first_val` as ordinary fixed accumulators — and could probably land quickly.
2. **`deltaMatrixLevelSource` / `deltaMatrixLevelSourceAggregate`'s running-level reconstruction.** This machinery (already "anchor-grid level", not an array-fold itself) reconstructs the DELTA-temporality zero-clamp `first_val` override from a per-anchor running SUM over a separate `deltaPrefixPairsAlias` array (still built via `groupArrayIf`, a genuinely different array from the window's own `window_pairs`/its #2760 replacement). Wiring #2760's fixed-accumulator `first_val` into this reconstruction — including the `useAggregateDeltaPrefix` / issue #2389 aggregate-DELTA-prefix variant — needs its own careful re-plumbing and differential-parity proof.

## Proposal

[...]

Issue #2759, opened 2026-08-30
Adjacency pre-annotation via lagInFrame for changes/resets/irate/idelta fan-out paths


tsouza:
## Problem

Matrix mode fans each sample to ~range/step+1 anchors (`sampleAnchorFanoutFrag`), then GROUP BY (series, anchor_ts) rebuilds an O(window-rows) array per group: `window_pairs` for irate/idelta (which read only elements [n-1],[n]; emitters at `internal/chsql/range_window.go:3201/3222`), `window_vals` for changes/resets (:3128/:3165 — the arrayPopBack/arrayPopFront/arrayMap/arraySum adjacency fold re-runs once per covering anchor, so every adjacent pair is re-evaluated ~range/step times). Native coverage gaps: `ts_grid_changes` is permanently AutoSelect=false (upstream NaN overcount, #1721); irate/idelta have no native member yet. Note (verified): `ts_grid_resets` IS auto-enabled on servers >= 25.9, so the resets win here applies only to ineligible shapes, capability-forbidden profiles, or <25.9 servers.

## Proposal

[...]

Issue #2758, opened 2026-08-30
Epic: Reduce fan-out pressure via window functions, array primitives, and joins (chopt-gated)


tsouza:
## Theme

For range-function shapes with no native ToGrid family member (or where the experimental setting is forbidden by the deployment's settings profile), cerberus still pays the arrayJoin sample-per-anchor fan-out with O(window) groupArray state per (series, anchor). This epic converts those shapes to fixed-size accumulators, per-series slabs, and merge joins using stable, version-floor-free ClickHouse primitives: `lagInFrame`/`leadInFrame`, `argMin`/`argMax`/`uniqExact`, `arraySlice`/`arrayReduceInRanges`, ASOF JOIN, `-ForEach`, `argAndMax`.

## Why it matters

The fan-out's groupArray-class memory is what the production guards exist to contain: the 2.8M-row rate_window_fanout bound (vs 40M for argMax-class in `lwr_fanout_bound.go` — a 10x admission gap at the same 1 GiB cap), the four-incident guard recalibration history (#2470/#2522/#2651/#2665), and the 2.12 GiB rate() OOM class. These items either move shapes into the fixed-accumulator admission class, close unguarded OOM axes, or provide a Stable non-experimental path for locked-down profiles that today have no native escape at all.

## Children

- [x] #2759
- [x] #2760
- [x] #2761
- [x] #2762
- [x] #2763
- [x] #2764

## Compatibility invariant (applies to every child)

Nothing in this epic deletes or degrades an existing execution path. Every adoption ships as a boot-probed chopt registry feature (or an operator config flag where the capability is version-independent) with an explicit ClickHouse version floor: servers below the floor, capability-forbidden profiles, and deployments that disable the feature keep the current lowering byte-for-byte unchanged. "Moving traffic off the fan-out" means eligible queries on capable servers stop exercising the guarded shape — the fan-out (and every other current path) remains the permanent, fully-supported fallback for older ClickHouse versions.

Issue #2389, opened 2026-08-19
DELTA-temporality prefix reconstruction: replace bounded-lookback approximation with an exact, retention-independent mechanism


tsouza:
## Background

`internal/chsql`'s DELTA-temporality prefix reconstruction (`instantDeltaPrefixSource` and
`emitWindowedArrayExtrapolatedMatrix`'s DELTA branch, `range_window.go`) reconstructs "the
cumulative counter level immediately before a `rate()`/`increase()` window" for an OTel
DELTA-temporality Sum counter, needed only for Prometheus's counter-reset / extrapolation-boundary
correction (`duration_to_start`'s clamp) — never for the `rate()`/`increase()` value itself, which
always comes from in-window samples alone.

Correctly reconstructing that level requires summing the series' contributions since its own true
accumulation start, which has no natural stopping point short of the series' first-ever sample. A
prior PR (fixing an incident where live dashboards timed out because this scan read a
metric's entire retention unconditionally, Cumulative-temporality series included, on every
`rate()`/`increase()` query) introduced `CERBERUS_DELTA_PREFIX_LOOKBACK`, a configurable, bounded
lookback (default 24h) as an immediate, safe mitigation: it trades a bounded approximation of the
boundary-correction term for retention-independent scan cost. That PR is explicit that this is an
accepted, documented, real approximation — not a precision fix — and defers the fully precise
answer to this issue.

## The gap

`CERBERUS_DELTA_PREFIX_LOOKBACK` only approximates the correction term for a DELTA-temporality
series whose true prior contribution predates the configured lookback. A deployment that:

- genuinely has long-lived DELTA-temporality Sum counters (rare today — empirically well under 1%
  of rows in a real-world `otel_metrics_sum` table were DELTA-temporality at all), **and**
- needs Prometheus-exact boundary-correction precision for them at query time,

has no way to get both correctness AND a bounded per-query scan simultaneously — it must choose one
of: raise the lookback (widening the scan, reintroducing the original incident's cost as retention
grows), disable it (`0`, the pre-fix unbounded scan), or accept the documented approximation.

## The fix this issue tracks

A materialized / incrementally maintained running total for DELTA-temporality Sum counters, so
reconstructing "cumulative level immediately before window start" is an O(1) lookup rather than an
O(retention) scan on every query. Candidate shapes (not decided — needs its own spec per this
repo's non-trivial-change workflow):

- A ClickHouse `AggregatingMergeTree` / materialized view keeping a running
[...]

PR #2801, opened 2026-08-31
feat(chsql,promql,chopt): fixed-accumulator rate/increase/delta (#2760)

### LINKED ISSUE #2760: Fixed-accumulator extrapolated increase/delta/rate fallback (argMin/argMax/uniqExact decomposition)
## Problem

Every extrapolated shape the native family cannot serve runs `emitWindowedArrayExtrapolatedMatrix` (`internal/chsql/range_window.go:4377`, regroup at :4483; instant at :3813): arrayJoin fan-out then `groupArrayPairFrag` window_pairs per (series, anchor), with the extrapolatedRate arithmetic evaluated over arrays. That population is real even after the native features land: increase/delta today (no lowerer entries — verified), rate's DELTA-temporality rows (the DELTA/CUMULATIVE split fires when TemporalityColumn is set and the cumulative arm is eligible; when ineligible the whole window falls back — either way DELTA rows pay the array path), capability-forbidden settings profiles, and native-ineligible shapes. This is the shape behind the 2.12 GiB rate() OOM (run 27277793810) and the 2.8M-row `rate_window_fanout_bound` guard.

## Proposal

Decompose Prometheus `extrapolatedRate` into O(1) aggregates per (series, anchor): `argMin((ts,val), ts)` / `argMax((ts,val), ts)` for the window endpoints, `uniqExact(ts)` for the >=2-sample rule, and — for CUMULATIVE counters — the reset-correction term as `sumIf(prev_val, val < prev_val AND prev_ts > anchor - range)` over lagInFrame-annotated pairs. The extrapolation arithmetic (duration_to_start/end, 1.1x clamp, per-second divide) operates on four scalar outputs with the same test-pinned parenthesisation. Ship in two slices: (1) `delta()` — gauge, no reset correction, pure endpoint+count decomposition, NO annotation dependency; (2) increase/rate fallback — depends on the adjacency-annotation feature for the reset term, and must re-express the DELTA-temporality groupArrayIf window/prefix-pairs split (`deltaMatrixLevelSource`'s sum() OVER reconstruction is anchor-grid level and unaffected). Frame as a chopt feature; express the inter-feature dependency at the `nativeRangeLowerers` wiring site exactly as `ts_grid_recollapse` does (the registry has no dependency mechanism). Floor: inherited from the annotation feature only — the aggregates themselves are ancient. AutoSelect=false initially; promote after optcorpus A/B. Docs: https://clickhouse.com/docs/sql-reference/window-functions (lagInFrame) + aggregate-function references for argMin/argMax/uniqExact/sumIf.

## Expected win

Converts cerberus's single worst OOM shape from groupArray-class to argMax-class memory (the 10x admission gap in `lwr_fanout_bound.go`): memory tracks group COUNT (series x anchors) instead of group CONTENTS. The 2.8M bound can be raised or retired for these shapes; route-B pressure drops; the DELTA arm gets flat-per-group memory even when the CUMULATIVE arm goes native. Cheaper-partial note: increase-as-native-rate-x-range covers the happy path but cannot cover forbidden-profile/ineligible/DELTA fallback shapes — this item is the complement, not a duplicate.

## Risks / parity

Largest emitter surgery in the repo (range_window.go's extrapolated family is the 5,128-line file's core; counter-reset sandwich, DELTA prefix reconstruction, native-histogram HistogramProjection adjacencies all touch it). The reset-correction sum (and DELTA-arm raw sum) accumulates floats in CH's nondeterministic aggregation order where today's arrayFold walks time order — ULP divergence vs bit-exact goldens is possible, and Kahan fidelity for the correction term is lost; quantify against the parity corpus before enabling (precedent: timeSeriesRateToGrid's internal order was already accepted for the native path). Endpoint picks, counts, and delta()'s path are exactly order-independent. Guard constants encode the old accumulator class — recalibration required; >=2-sample and empty-window drop rules must be re-proven per slice.

## Effort

L

## Depends on

- Adjacency pre-annotation via lagInFrame for changes/resets/irate/idelta fan-out paths (slice 2 only; the delta() slice is independent)

tsouza:
## Summary

Fixed-accumulator fallback for the extrapolated window family — `rate()` / `increase()` / `delta()` — replacing the `groupArray` + `arraySort` + `arrayFilter` array-fold fan-out (`emitWindowedArrayExtrapolatedMatrix`) with per-`(series, anchor)` ClickHouse aggregates (`count`/`min`/`max`/`argMin`/`argMax`/`sumIf`, plus their `-If` forms for a temporality-bearing counter). No array is ever materialized, so peak memory tracks group **count** (series × anchors) instead of group **contents** — the argMax-class admission the parent epic (#2758) targets, versus the array-fold's groupArray-class state.

This is the direct sibling of #2759's `lagInFrame` adjacency shape for `changes()`/`resets()`/`irate()`/`idelta()`, extended to cover the boundary-extrapolation arithmetic those four don't need: the FIRST and LAST sample of each anchor's window (Prom's `extrapolatedRate` correction), not merely adjacent pairs.

Closes #2760.

## What's in scope

- **`delta()`** — full coverage (gauge, no counter-reset arithmetic).
- **`rate()` / `increase()`** — full coverage, including:
  - the counter-reset correction term (`(last_val - first_val) + reset_sum`, algebraically proven equal to the array-fold's sequential `if(c<p,c,c-p)` fold via a telescoping-sum identity — see the doc comment on `fixedAccumCounterDeltaFrag`), reusing #2759's own `lagInFrame` validity check and reset kernel directly;
  - the DELTA-vs-CUMULATIVE `AggregationTemporality` runtime branch (`CounterOrDeltaSum`'s own split), including the DELTA-temporality counter zero-clamp reconstruction (`deltaMatrixLevelSource`/`deltaFirstValFrag`, reused **unchanged** — this piece is anchor-grid-level, not the array-fold this issue targets).
- **Duplicate-timestamp handling** — the array-fold path this replaces *does* dedupe same-`(series,ts)` rows (keeping the max-value row per tied timestamp, an OTel/ClickHouse ingestion artifact) before computing first/last/count, unlike #2759's own targets. Reproduced relationally via one forward-looking `leadInFrame` pass (`fixedAccumDedupLayer`) rather than a signed-tuple `argMin`/`argMax` trick, to avoid any ClickHouse NaN-tuple-comparison ambiguity.

## What's out of scope (tracked)

- The EXACT, retention-independent DELTA-prefix aggregate mechanism (#2389 — `RangeWindow.DeltaPrefixAggregateInput` / `deltaMatrixLevelSourceAggregate`), a narrower opt-in-only population needing its own re-plumbing. Filed as #2797.
- The INSTANT (single-anchor, `OuterRange == 0`) shape — scoped to the query_range MATRIX shape only, mirroring #2759's own `LagAdjacency` field scoping (the instant shape's memory profile never motivated the OOM class either issue targets).

Every excluded shape keeps the unchanged array-fold emission — a strict narrowing, never a regression.

## Feature flag

`chopt.FeatureFixedAccumulatorExtrapolated` (`fixed_accumulator_extrapolated`) — `AlwaysAvailable` floor (argMin/argMax/uniqExact/count/lagInFrame/leadInFrame are all long-standing ClickHouse primitives, well below any version this repo targets), `AutoSelect: false` per the issue's own proposal (pending an optcorpus A/B pass — see Performance below for why). Reachable via `CERBERUS_CH_OPTIMIZATIONS=fixed_accumulator_extrapolated`. Layers beneath `rate`/`increase`'s existing native `ts_grid_range`/`ts_grid_increase` strategy exactly like `laginframe_adjacency` layers beneath `changes`/`resets`' native strategy; `delta()` has no native `timeSeries*ToGrid` competitor, so its strategy is the two-tier `FixedAccumulator{Fallback: Fanout{}}` directly.

## Correctness

`internal/chsql/range_window_fixed_accumulator_chdb_test.go` — dual-emit differential parity against the array-fold, on one ephemeral chDB session:

- `delta()`: **bit-identical** (`math.Float64bits` equality) across a normal walk, exactly-2-samples-at-the-window-boundary, a single-sample (absent) job, duplicate timestamps, all-NaN, and a NaN sandwiched between real samples — delta's raw result (`last_val - first_val`) is exactly order-independent, matching the issue's own claim.
- `rate()`/`increase()`: ULP-budget assertion (mirroring `range_window_grid_native_increase_chdb_test.go`'s own measured-not-assumed pattern) across a sawtooth-counter reset case, a monotonic no-reset case, a DELTA-temporality series with prefix history reaching before the query window (forcing the zero-clamp reconstruction to matter, not just the raw window's own first sample), duplicate timestamps, a single-sample job, and all-NaN. **Measured: 0 divergent cells at 0 ULP** on this fixture — both constants are pinned at the measured value (never loosened for headroom), with a documented path to bump them with a reviewed measurement if a genuinely parallel-aggregation ClickHouse substrate ever shows real drift.

Plus three TXTAR fixtures (`test/spec/promql/fixed_accumulator_{delta,rate,increase}_range_step.txtar`) pinning the SQL shape and chDB-executed `expected_rows`, and the `TestNativeStrategies_*` coverage ratchets (`internal/promql/native_strategy_coverage_test.go`) extended so the corpus can select this strategy standalone (unshadowed by the native ts_grid path).

## Performance (real ClickHouse, real production-shaped data)

Ran both emitters' SQL directly against a real ClickHouse 25.9 container (not chDB) loaded with the LFS smoke sample (`svc_http_requests_total.parquet`, ~18.6M rows / 31,073 real distinct series, 14-day span), for `sum by(job) (rate(http_requests_total[5m]))` at a 30s step:

| query_range window | array-fold (today) | fixed-accumulator |
|---|---|---|
| 30 min | 397 MiB, 3.6s ✅ | 1.13 GiB, 3.5s ✅ |
| 2 h | **rejected** — hits `rate_window_fanout_bound`'s 2.8M-row `throwIf` guard (0.8s, no data returned) | 2.79 GiB, 7.2s ✅ (693k rows) |
| 4 h | **rejected** — same guard | 3.23 GiB, 12.5s ✅ (1.1M rows) |

At any realistic dashboard-scale window (multi-hour, the common case), the array-fold's own OOM-prevention guard rejects the query outright today — the fixed-accumulator path serves it directly, with peak memory growing **sub-linearly** relative to result cardinality (≈5.4 KB/row at 30 min → ≈3.0 KB/row at 4 h), no evidence of runaway growth.

At the trivially-small 30-minute window, fixed-accumulator uses ~2.8× more memory than the array-fold — the two extra window-function passes (`leadInFrame` dedup, `lagInFrame` reset-detection) run over the full raw row stream *before* any fan-out/grouping, versus the array-fold's window-shaped computation running only after grouping collapses cardinality. Both figures are trivially small in absolute terms (sub-GB) and this scale never risks the guard either way — but it's a genuine, documented tradeoff, and exactly why this ships `AutoSelect: false` pending the optcorpus A/B the issue's own proposal calls for, rather than an unconditional win claim.

(The 18.6M-row full-table scan on every query in this benchmark is an artifact of the ad-hoc benchmark table's `ORDER BY (MetricName, Attributes, TimeUnix)` layout with a single constant `MetricName` — not a real production partitioning scheme — so absolute read-row counts here are not representative; the relative comparison between the two emitters, which both read from the identical table, is unaffected.)

## Follow-ups filed

- #2797 — temporality-bearing rate()/increase() under the EXACT DELTA-prefix aggregate mechanism (#2389).

## Test plan

- [x] `go build ./...`, `go vet ./...`
- [x] `go test ./internal/chsql/... ./internal/promql/... ./internal/chplan/... ./internal/chopt/... ./cmd/cerberus/...`
- [x] `go test -tags chdb -run TestFixedAccumulator ./internal/chsql/...` (dual-emit parity, 0 ULP divergence)
- [x] `just update-golden all` (TXTAR SQL/expected_rows, solver-decision baseline, parity-enrolment baseline)
- [x] `just gen-opt-docs` (generated feature table)
- [x] `node .github/scripts/forbid-sql-raw.mjs` / `forbid-verbatim-concat.mjs` / `forbid-chplan-fn-literal.mjs`
- [x] Real ClickHouse 25.9 (docker) + real production-shaped sample data (LFS smoke sample) — see Performance above
- [ ] CI (lint, strict-scan, compat, e2e — pushing now for real feedback)


tsouza:
Re-triggering CI: no GitHub Actions check-suite was created for the last push (50d96aa62) after 40+ minutes — closing and reopening to force a fresh pull_request event.

Issue #2804, opened 2026-08-31
Extend #2761's sorted-slab *_over_time shape: remaining functions + arraySlice index lookup


tsouza:
## Follow-up to #2761

#2761's PR ships the per-series sorted-slab shape for `sum_over_time()` /
`avg_over_time()` only, in query_range matrix mode. Two things were
deliberately scoped out of that PR and are tracked here:

### 1. Extend to the rest of the array-path `*_over_time` family

`first_over_time`, `last_over_time`, `stddev_over_time`, `stdvar_over_time`,
and `mad_over_time` still ride the array-fold fan-out
(`internal/chsql.emitWindowedArrayMatrix`) in matrix mode — the same
unguarded per-(series, anchor) regroup #2761 closes for sum/avg. Extending
the sorted-slab shape to them is mechanical (`overTimeArrayValueFrag`
already handles all five), but needs its own byte-identical-contract
argument per function: `first`/`last_over_time` pick a specific element by
window position (order-sensitive but not a reduction), and
`stddev`/`stdvar`/`mad_over_time` run a two-pass computation over the sliced
array — each deserves its own chDB differential coverage before landing.

### 2. arraySlice + real index lookup instead of arrayFilter

The #2761 PR's `sortedSlabWindowValsFrag` cuts each anchor's window out of
the per-series samples slab with `arrayFilter`, which is O(samples) per
anchor (proven memory-safe — no per-(series,anchor) materialization — but not
CPU-optimal: it re-scans the whole slab per anchor rather than a real O(log
n) or O(1)-amortized index lookup). The original issue's proposal was
`arraySlice` driven by binary-search index math over the sorted per-series
timestamp array. That needs a verified index-lookup ClickHouse array
function (e.g. a binary-search-style array function) confirmed against this
repo's pinned chDB substrate — not done in #2761 for lack of a verified
function name/version floor. `range_window_fused.go`'s own `coveredValuesFrag`
uses arithmetic grid-index math instead of a real binary search because BOTH
its anchor grids are step-aligned; #2761's anchors are grid-aligned but its
SAMPLES are not, so the arithmetic shortcut doesn't apply here and a genuine
lookup function is needed.

Both items are pure follow-on refinements — #2761 already closes the
memory/OOM-class problem the parent issue describes; this issue is about
extending coverage and a CPU refinement, not correctness.

Issue #2761, opened 2026-08-30
Replace matrix *_over_time fan-out with per-series sorted slabs (arraySlice/arrayReduceInRanges)


tsouza:
## Problem

Matrix-mode `sum`/`avg`/`first`/`last`/`stddev`/`stdvar`/`mad`/`quantile_over_time` ride `emitWindowedArray` (`internal/chsql/range_window.go`): arrayJoin fans each sample to range/step+1 anchors, then a blocking GROUP BY (series, anchor_ts) rebuilds per-window arrays — memory = samples x fanout. Precision (verified): min/max/count/present_over_time do NOT ride this path (`overTimeDirectAggFrag` gives them fixed-size direct aggregates), and `emitWindowedArrayMatrix` carries NO fanout guard — the 2.8M bound + second-scan truncation probe guard only the extrapolated (rate/increase/delta) sibling. So this shape is an UNGUARDED OOM axis of the same mechanism as the 2.12 GiB rate() incident, and this proposal closes it rather than deleting existing guards. The fused emitter (`range_window_fused.go`) already proves the O(samples-per-series) slab pattern, but only for min/max/count over rate|increase|delta subqueries (#1505).

## Proposal

Generalize the fused shape: one GROUP BY series builds a single sorted (ts, value) slab per series covering [start-range, end]; per-anchor (index, length) ranges derive from binary-search index math over the sorted ts array (same half-open membership); evaluate all anchors in one pass. Reducer forms: the `arraySlice` form reusing the exact fused helpers is MANDATORY for sum/avg/stddev/stdvar/mad under the byte-identical contract (`arrayReduceInRanges` segment-tree partial-state merging changes float association vs arraySum's linear order — sum/avg were deliberately kept off the direct-agg path for exactly this, range_window.go:2864); `arrayReduceInRanges` (20.4 — AlwaysAvailable, chopt id only for staged rollout/kill-switch) is fine for order-insensitive reducers. stddev/stdvar keep the two-pass variance frags computed over slices. Empty windows must yield NaN-drop, not zero-size aggregate output (filter zero-length ranges; CH #115008 concerns all-NULL states, which cerberus's non-nullable Float64 arrays likely never trigger — the filtering is for sum=0-vs-drop parity). Planner: model the new carrier as fanout=1 in `carrierGeometryOf`; teach the five slice-invariance sites. Doc: https://clickhouse.com/docs/sql-reference/functions/array-functions#arrayReduceInRanges

## Expected win

[...]

Issue #2429, opened 2026-08-21
engine: plain sum by()(rate(<counter>[5m])) hits the memory cap at real (3.7k-series) cardinality despite applySpillSettings


tsouza:
## What happened

The #2370 nightly measurement harness (in progress, `test/perf/nightly/`) loaded the real
14-day production sample (test/perf/smoke/testdata/samples/svc_http_requests_total.parquet,
trimmed to 2026-08-18: 1,338,836 rows, 3,741 distinct series, 5 distinct `method` values) into a
real ClickHouse 25.8-alpine and issued two plain PromQL `query_range` requests over the real data.
Both hit ClickHouse's `MEMORY_LIMIT_EXCEEDED` (code 241) against the default 1 GiB
`CERBERUS_CH_QUERY_MAX_MEMORY` cap:

1. `sum by (method) (rate(svc_http_requests_total[5m]))` over `[09:05:00, 13:25:00)` @ step=1m
   (260 steps) — **99.8% of the 1 GiB cap at termination** (1,071,845,102 bytes), aborted:
   `Query memory limit exceeded: would use 1.00 GiB ... While executing
   ScatterByPartitionTransform`.
2. `sum by (k8s_namespace_name) (rate(svc_http_requests_total{status_class="<value>"}[5m])) /
   sum by (k8s_namespace_name) (rate(svc_http_requests_total[5m]))` over the same window —
   **100.0% of cap** (1,073,682,175 bytes), aborted: `... While executing
   SourceFromNativeStream`.

Neither query is a histogram construct — this is a plain counter `rate()` + `sum by()`, the single
most common PromQL shape there is. 3,741 series and 260 steps is not an exotic cardinality:
`test/perf/smoke`'s own calibrated spill sentinel (`spill_high_cardinality_groupby`,
`applySpillSettings`) uses **10,000** series at a similar step count and measures only ~650 MiB
(61% of the same 1 GiB cap) — nearly 3x the series count at less than 2/3 the memory. Something
about this real data's shape (real, uneven scrape density — the sample averages ~358 samples per
series across the ~4.3h window, vs. the smoke tier's synthetic evenly-spaced 15s-step generator)
appears to defeat or bypass the always-on spill mechanism
(`applySpillSettings`/`internal/engine/spill.go`) that is supposed to convert exactly this class of
blowup into a slower disk spill rather than an outright OOM abort.

## Why this matters

[...]

PR #2805, opened 2026-08-31
feat(chsql,promql,chopt): sorted-slab sum_over_time/avg_over_time (#2761)

tsouza:
## Summary

- Replaces the `arrayJoin` sample-side fan-out + `GROUP BY (series, anchor)` regroup
  (`emitWindowedArrayMatrix`) with a per-series sorted-slab shape (one `groupArray` per series,
  sliced once per anchor via `arrayFilter`) for `sum_over_time()` / `avg_over_time()` in
  `query_range` matrix mode.
- That regroup is an unguarded fan-out axis today — unlike rate/increase/delta (#2429) it carries no
  size guard — so peak memory scales with samples × anchors per series. The sorted-slab shape bounds
  peak memory to samples-in-range per series, independent of anchor count, closing an unguarded OOM
  axis of the same shape as the 2.12 GiB `rate()` incident.
- Reuses `overTimeArrayValueFrag` unchanged over the sliced window, so `arraySum` / `arrayAvg` fold
  in the exact same left-to-right order the array-fold path already uses — the byte-identical
  reduction-order contract issue #2761 calls out as mandatory — without introducing
  `arrayReduceInRanges` (segment-tree partial-state merging would reorder the float summation).
- New `chopt.FeatureSortedSlabOverTime`: `AlwaysAvailable` (no version floor — every function used
  is an ancient ClickHouse primitive), `AutoSelect: false` pending an optcorpus A/B pass, reachable
  via `CERBERUS_CH_OPTIMIZATIONS=sorted_slab_over_time`.

## Scope

`sum_over_time` / `avg_over_time` only, matrix mode only. Deliberately narrower than #2761's full
proposal:

- The remaining array-path `*_over_time` members (`first_over_time`, `last_over_time`,
  `stddev_over_time`, `stdvar_over_time`, `mad_over_time`, `quantile_over_time`) and a genuine
  binary-search-style index lookup into the sorted slab (vs. this PR's `arrayFilter` scan) are
  tracked as a follow-up: #2804.
- `min_over_time` / `max_over_time` / `count_over_time` / `present_over_time` are already off the
  array-fold entirely (`overTimeDirectAggFrag`'s direct CH group aggregate) — out of scope here.
- The instant (single-anchor) shape is unchanged — the sorted-slab decomposition is matrix-mode only.

## Notes for reviewers

- No dedup: the array-fold path this narrows (`windowValsFrag`) does not collapse duplicate-timestamp
  samples before reducing — every sample counts individually — and this emitter reproduces that
  exactly (no `dedupWindowPairsByTsFrag`, unlike the extrapolated/fused matrix emitters, which need it
  for a different reason — boundary-sample tie-breaking for rate/increase/delta's extrapolation, not
  a `*_over_time` concern).
- `carrierGeometryOf` (`internal/solver/planner.go`) and `chplan/reanchor.go` are intentionally
  untouched: `*chplan.RangeWindow`'s existing entry already covers every sub-mode uniformly
  (`LagAdjacency`, `FixedAccumulatorExtrapolated`, and now `SortedSlabOverTime`), and #2760 shipped
  the same kind of emitter-only flag without touching either file — confirmed against
  `git show --stat df2478394`, which only regenerated one solver-decision baseline row.
- `internal/chplan/sliceinvariant.go` gets a documentation-only addition (`RangeWindow` is already
  blanket-registered as slice-invariant): `SortedSlabOverTime` needs no lagInFrame-style argument at
  all, since it reads no window function and no scan-order-dependent seed.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [ ] chDB dual-emit differential test (`SortedSlabOverTime=false` vs `=true`) — in progress, pushing
      as a follow-up commit to this PR
- [ ] `test/spec/promql` TXTAR fixture(s) — in progress
- [ ] `just lint` clean
- [ ] `just test` passes (race + spec)
- [ ] CI green

Opening as a draft while the differential test / fixtures land in a follow-up commit — the core
lowering + emitter implementation is done and reviewable now.


tsouza:
Verification pass from a sibling agent that was independently working #2761 (see the message thread for context on the collision — leaving this here rather than duplicating the PR).

Findings:
- Design is sound: single `GROUP BY series` slab, no dedup (correctly matches the un-deduped `windowValsFrag` semantics of the array-fold path this replaces — verified by reading `range_window.go` directly), reuses `overTimeArrayValueFrag` unchanged for byte-identical reduction order. Confirmed `arrayLastIndex` is a real function already used in this codebase (`internal/chsql/histogram_quantile_native.go:612`) in case the arrayFilter-vs-arraySlice choice gets revisited later.
- Gap: the doc comments in `internal/chsql/range_window_sorted_slab.go` and `internal/chplan/sliceinvariant.go` both cite `range_window_sorted_slab_chdb_test.go` as the source of "verified by dual-emit parity" claims, but that file does not exist in this PR's diff — no test was actually added. That's very likely why `forbid-skip`, `mutation`, and `mutation-select` are currently failing (no coverage on the new emitter to kill mutants against). Recommend adding the differential chDB test (SortedSlabOverTime=false vs =true, across duplicate-timestamp, single-sample, and empty-window cases per the doc's own claim) before marking this ready for review — the assertion of verification shouldn't outlive the test that backs it.

Issue #2370, opened 2026-08-18
Design: large-scale, realistic-scale performance/resource-cost test lane


tsouza:
# Design: a large-scale, realistic-scale performance/resource-cost lane

## Problem statement

A recent change disabled ClickHouse's query analyzer for a specific plan shape, in order to fix a
correctness/lowering concern. The change broke a performance assumption the SQL emitter relied on:
the analyzer's common-subexpression-elimination (CSE) fold had been keeping two array-scan
expressions cheap. The change was validated against a request-deadline timeout on CI-fixture-scale
data (small row counts, low cardinality) and passed cleanly. On realistic-scale, high-cardinality
data, the same plan shape produced `MEMORY_LIMIT_EXCEEDED` failures instead of a timeout — a
different failure mode entirely, on a resource axis (memory) nothing in CI measures, discovered only
after the change reached a live deployment.

This is not a one-off gap. Cerberus already has a structurally identical precedent:
`applyNativeHistogramAnalyzerFix` in `internal/engine/query_settings_rules.go` disables the CH
analyzer for native/exponential-histogram merge shapes because the analyzer is superlinear on deeply
nested `arrayFold`/`arrayMap` trees. That fix's own doc comment cites the validation method: a
manual, non-repeatable timing comparison (~8.4s vs ~1.4s on CH 24.8.14, 8 runs, one engineer, one
sitting) — time-only, never validated against memory pressure at realistic cardinality, and never
re-validated by anything that runs again. That is exactly the kind of one-off validation that let the
incident above go uncaught in CI: cerberus tunes analyzer/execution settings per plan shape based on
manual investigation, then has no durable, repeatable harness that re-checks the assumption on every
subsequent PR that touches an analyzer-affecting code path.

**The general, durable gap**: nothing in cerberus's test suite measures ClickHouse-side resource cost
(peak memory, CPU, wall-clock) against data shaped and scaled realistically, for any of the three
signal types (metrics, logs, traces). This issue is a design for closing that gap — not an
implementation. It is written for whoever picks this up next.

## What exists today (and why it doesn't cover this)

Cerberus's existing performance-testing story (`docs/performance.md`, `docs/test-strategy.md` Layers
11–12) is built entirely around **compute fan-out** — intermediate row-set explosion relative to leaf
scan rows — not resource cost at scale:

- **Static fan-out lint** (`internal/perf/fanout`) — pre-execution pattern matching on plan shape, no
  data at all.
[...]

Issue #1971, opened 2026-08-08
test(chdb): -race + chdb SIGSEGVs in runtime.racefini() during process exit


tsouza-squid:
## What happens

Reproduced while investigating #1917: the chdb-tagged test substrate cannot run under `-race` at
all. A package's own tests all `PASS`, but the process then dies with `SIGSEGV` inside
`runtime.racefini()` during exit — i.e. after everything asserted has already succeeded, during
the race detector's own teardown while libchdb's CGO side is being unwound. The exit status is a
crash, so the lane reports failure on a run in which nothing actually failed.

Reproduced locally (3 clean runs of `./internal/api/loki/`, but a crash on the first run of
`./internal/api/...`):

```
$ go test -race -tags chdb -timeout 300s -count=1 ./internal/api/...
...
ok  	github.com/tsouza/cerberus/internal/api/prom	...
PASS
SIGSEGV: segmentation violation
PC=0x7e18f6d82a1a m=8 sigcode=1 addr=0x20

goroutine 1 gp=0xc0000081e0 m=8 mp=0xc000189008 [syscall]:
runtime.racefini()
	runtime/race.go:507 +0x1b
os.runtime_beforeExit(0x0)
	runtime/proc.go:342 +0x2a
os.Exit(0x0)
	os/proc.go:75 +0x3a
main.main()
	_testmain.go:528 +0x16a
...
FAIL	github.com/tsouza/cerberus/internal/api/tempo	3.922s
```

The crashing package varied between local attempts (`internal/api/tempo` here; the original report
that surfaced this, a comment on #1917, saw it on `internal/api/loki`), consistent with a
teardown-ordering race rather than a deterministic per-package bug.

## Root cause (shared with #1917, different symptom)

Same root as #1917's Parquet page-index corruption: libchdb is a process-global session
(`chdb-io/chdb-go`'s `globalSession`, see `chdb/session.go`) with no lifecycle discipline on the Go
side, so its failure modes arrive as panics and segfaults indistinguishable from cerberus defects.
#1917 forges a `500` from a handler that is correct; this forges a crash from a suite that is
green.

## Blast radius

CI does not see this because `just test-chdb` (the `chdb` CI job) omits `-race` — which is the part
worth naming: the race detector is a required gate for `just test-unit` over the untagged tree, and
the chdb-tagged code paths (`internal/chclienttest`, and every handler test file gated
`//go:build chdb`) are the only ones it structurally cannot cover today. That's a coverage hole with
[...]

PR #2897, opened 2026-09-01
fix(perf): read nightly sample tuple fields by name, not prefix-collapsing dot access

### LINKED ISSUE #2875: nightly perf-nightly run did not reach a clean pass
The nightly `perf-nightly` schedule run did not reach a clean pass.

Run: https://github.com/tsouza/cerberus/actions/runs/33495961118 (id 33495961118)

Non-success jobs (anything but `success` — a real failure, a `cancelled`
kill, or an unexpected `skipped`):
- `perf-nightly: failure`

Filed/updated automatically by this lane's notify script — see tsouza/cerberus#2370 for why this mechanism exists: a nightly that reports red into a place nobody looks is as silent as one that never reports at all. This issue closes itself on the next clean nightly.

tsouza:
## The break

`perf-nightly` is a `RELEASE_REQUIRED_CHECKS` entry (`.github/workflows/release.yml:267`) and had
concluded `failure` on every recent push to `main`, including the tip. One step failed: **Run the
ts_grid_instant real-data memory measurement (real ClickHouse)**, on its 26.6-alpine container.

```
insert into otel_metrics_sum from svc_http_requests_total.parquet: code: 10,
message: Not found column or subcolumn ResourceAttributes.deployment.environment.name in block.
There are only columns: ... ResourceAttributes.deployment.environment, ...
```

## Root cause

`test/perf/nightly/loader.go:70` (pre-fix) rendered each Parquet tuple field as backtick-quoted **dot
access**:

```go
valueList += fmt.Sprintf("%s.`%s`", column, k)
```

`sampleResourceAttributeKeys` (`loader.go:44-50`) lists both `deployment.environment` and
`deployment.environment.name` — the OTel semantic-convention name the captured production workload
emitted and its successor, emitted side by side during that convention's migration window.

### The fixture really does carry both keys

The error's `There are only columns: ...` list is the **pruned requested column set for that query**,
not the file's schema. The same repro with only two projections reports
`There are only columns: ResourceAttributes.deployment.environment` — a one-entry list. So that list
is not evidence the fixture lacks the key.

The fixture's own schema, read by **26.6 itself**:

```
$ clickhouse-local (26.6-alpine)
  "SELECT field FROM (SELECT arrayJoin(tupleNames(ResourceAttributes)) AS field
                      FROM file('svc_http_requests_total.parquet', Parquet) LIMIT 24)
   WHERE field LIKE 'deployment%'"
deployment.environment
deployment.environment.name
```

`DESCRIBE TABLE file(...)` agrees and reports the identical 23-field tuple under 25.9 and 26.6 — the
two engines read the same schema out of the same bytes.

Both fields carry real, non-null data, and the longer one carries strictly more of it, so dropping it
would be a genuine reduction in what the lane loads and measures:

```
$ clickhouse-local (26.6-alpine), over all rows of the fixture
  uniqExact(deployment.environment)            = 1
  uniqExact(deployment.environment.name)       = 3
  countIf(deployment.environment.name IS NULL) = 0
  count()                                      = 1338836
```

### Why dot access breaks on 26.6 and not 25.9

Reading through the `file()` table function, ClickHouse pushes the requested column paths down into
the format reader, and from 26.x that pushdown splits the requested path at the longest matching
**column** prefix: it asks Parquet for `ResourceAttributes.deployment.environment` (a `String`) and
then cannot find the `.name` remainder inside it. Backticks do not disambiguate it — the split
happens after identifier parsing.

Reproduced directly, outside Go, against both pinned engines on the real fixture:

```
$ clickhouse-local (25.9-alpine) "SELECT ResourceAttributes.`deployment.environment`,
                                         ResourceAttributes.`deployment.environment.name`
                                  FROM file('svc_http_requests_total.parquet', Parquet) LIMIT 1"
scrubbed-26f195d4a6b3   scrubbed-cb08d24f7f62

$ clickhouse-local (26.6-alpine) — same query
Code: 10. DB::Exception: Not found column or subcolumn
ResourceAttributes.deployment.environment.name in block. (NOT_FOUND_COLUMN_IN_BLOCK)
```

The same query against an **in-memory** tuple works on both versions, which places the change in the
table-function column pushdown, not in general expression resolution.

That version split is why this went unnoticed: the `#2370` nightly harness pins `perfNightlyCHImage`
= 25.9-alpine and still loads all three samples fine. Only the ts_grid_instant measurement, whose
feature floor sits above 25.9 and which therefore boots its own 26.6 container, went red. And the
lane's `merge_posture` is `never` (`.github/ci-lanes.json`), so no PR ran it — it merged broken and
reddened only on the schedule.

## The fix

Render `tupleElement(col, 'key')` instead. The field name is a string **literal**, so no path
splitting happens at all and every field — prefix-colliding or not — resolves to its own column.
Verified to work on both 25.9 and 26.6 over the real fixture.

Both keys stay in `sampleResourceAttributeKeys`.

The second commit adds the `TestMain` calling `chdbsession.CloseForExit` that
`TestChDBTaggedPackagesCloseSessionOnExit` requires of any package with a chdb-tagged test file
(#1971) — `test/perf/nightly` gained its first one here.

## Validation

Both perf-nightly steps were run end to end against real ClickHouse containers, not inferred from
green neighbours:

- **`perf-nightly` itself, dispatched on this branch** (`workflow_dispatch`, since the lane never
  runs on a PR): green on this exact tip (`5744f11`) —
  https://github.com/tsouza/cerberus/actions/runs/33560938675. Both steps measured for real:
  `loaded counter sample: 1338836 rows` / `PASS: TestPerfNightlyRealCH (137.76s)` on 25.9, and
  `lookback=1d: fan-out peak memory_usage = 167239118 bytes; native peak memory_usage = 114727538 bytes (native is 68.6% of fan-out)`
  on 26.6. (An earlier dispatch on the pre-rebase commit was green the same way:
  https://github.com/tsouza/cerberus/actions/runs/33559235273.)
- **Locally, 26.6 leg** (`just ts-grid-instant-memory-integration`, the step that was failing) — PASS
  in 80s, 1,338,836 real rows loaded, `fan-out 209352204 bytes / native 177736449 bytes (84.9%)`.
- **Locally, 25.9 leg** (`just perf-nightly-integration`, already passing) — PASS in 697s, all five
  sentinels, identical row counts (1317183 / 1338836 / 1676160 / 1317183 derived) and in-band memory
  readings. No regression from the changed rendering.

## Regression pins

`test/perf/nightly/loader_prefix_collision_chdb_test.go` (new, `chdb`-tagged, enrolled via the
`chdb.perf-guards` lane's `test/perf/**` globs) writes a Parquet whose `ResourceAttributes` tuple
carries `sampleResourceAttributeKeys` verbatim, then reads every key back through the SELECT list
`mapArraysExpr` itself renders, requiring each to return its own distinct seeded value. chDB is 26.5
and reproduces the collapse exactly.

Non-vacuity was checked, not assumed: with dot access temporarily restored in `mapArraysExpr`, the
new test fails with the identical `NOT_FOUND_COLUMN_IN_BLOCK` message seen in CI.

Two unit pins back it up, covering the two ways this could silently come back:

- `TestMapArraysExpr_NeverUsesDotAccess` — no rendered field access may be dot access, checked over
  the loader's real key sets.
- `TestSampleResourceAttributeKeys_KeepPrefixCollision` — fails if the sample key set ever stops
  carrying a dotted-prefix pair, which is the shape "fix it by deleting a key" would take.

Closes #2875

https://claude.ai/code/session_01E6EwMr2g22x8oT3kk7eTtR

### DIFF
--- a/test/perf/nightly/loader.go
+++ b/test/perf/nightly/loader.go
@@ -10,11 +10,10 @@
 // ClickHouse exporter's own templates, see internal/schema/ddl) declares
 // both as Map(LowCardinality(String), String) — so loading is not a bare
 // `INSERT ... SELECT * FROM file(...)`; each named tuple field is read back
-// via dot access (backtick-quoted for the dotted ResourceAttributes keys)
-// and re-packed with mapFromArrays. This file is test-only data-loading
-// code, not the chsql emitter, so invariant 10 (typed Frags only) does not
-// apply here — the same carve-out scale_wall_pin_chdb_test.go's own seeding
-// already uses.
+// by name (see mapArraysExpr) and re-packed with mapFromArrays. This file
+// is test-only data-loading code, not the chsql emitter, so invariant 10
+// (typed Frags only) does not apply here — the same carve-out
+// scale_wall_pin_chdb_test.go's own seeding already uses.
 package nightly
 
 import (
@@ -41,6 +40,15 @@ var sampleAttributeKeys = map[string][]string{
 // sampleResourceAttributeKeys names the ResourceAttributes struct fields
 // every sample metric carries identically (the extraction pipeline flattens
 // the same resource-level key set regardless of metric type).
+//
+// Two of these are a PREFIX PAIR: "deployment.environment" (the OTel
+// semantic-convention name the captured production workload emitted) and
+// "deployment.environment.name" (its successor, emitted alongside it during
+// the convention's own migration window). Both are real, both carry
+// distinct values in the captured sample, and both stay listed — dropping
+// either would silently narrow what the nightly lane loads and measures.
+// Their coexistence is exactly what mapArraysExpr's field access has to
+// survive; see that function's comment.
 var sampleResourceAttributeKeys = []string{
 	"container.image.name", "container.image.tag", "deployment.environment", "deployment.environment.name",
 	"host.name", "k8s.cluster.name", "k8s.cluster.uid", "k8s.container.name", "k8s.deployment.name",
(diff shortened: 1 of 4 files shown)


tsouza:
Dispatched `perf-nightly` on this branch (the lane's `merge_posture` is `never`, so it does not run on the PR itself):

https://github.com/tsouza/cerberus/actions/runs/33559235273

Issue #2910, opened 2026-09-01
ci(mutation): split gremlins' per-mutant leash into an overall deadline and a run-only one


tsouza:
Spun out of #2903, which fixes the two halves that live in this repository —
`gremlins-threshold.mjs` no longer scores a timeout as a detection, and
`mutation-run.mjs` now measures the per-mutant budget rather than declaring a
flat one. Both are compensations. The cause they compensate for is in the
`tsouza/gremlins` fork, and it is a two-line shape.

## The mechanism

`internal/engine/executor.go` at `v0.6.0-cerberus-timeout-max-consume` leashes
each mutant twice, from the same number:

```go
// runTests
ctx, cancel := context.WithTimeout(m.runCtx, m.testExecutionTime)
cmd := m.execContext(ctx, "go", m.getTestArgs(pkg)...)
...
if errors.Is(ctx.Err(), context.DeadlineExceeded) {
    return mutator.TimedOut
}

// getTestArgs
// Here we add some seconds to the timeout to be sure it's gremlins that catches the test
// timeout and not the test itself. ...
args = append(args, "-timeout", (2*time.Second + m.testExecutionTime).String())
```

The two leashes bound different things:

- the **context deadline** wraps the whole `go test` child — resolve, compile,
  link, and only then run;
- **Go's own `-timeout`** bounds only the test binary's EXECUTION. The compile
  is already finished by the time it starts counting.

`getTestArgs` deliberately sets the second one 2s ABOVE the first, with the
stated intent that "it's gremlins that catches the test timeout". So the
run-only leash is present, correct, and unreachable: the context deadline always
fires first, and every overrun — whatever caused it — is recorded as
`TIMED OUT`.

## Why that matters

The per-mutant cycle is dominated by the part the run-only leash does not
cover. Measured per mutant on an idle-ish 8-core machine:

| scope | compile + link | test run |
| ----------------- | -------------- | -------- |
| `internal/chsql`  | 12.7-14.4s     | 0.31s    |
| `internal/promql` | 8.4-8.7s       | 2.07s    |

So `TIMED OUT` is, in practice, a statement about the compiler. #2903 has the
full evidence: two runs of `phase4-promql-lower` with the same 486 kills where
the slower runner scored three points higher, and a green run where all four
`internal/chsql` legs reported 100.00% with `lived: 0` over 1266 mutants of
which 1230 timed out.

## The fix

Let the two leashes carry the two meanings they already have:

[...]

Issue #2891, opened 2026-09-01
Epic: CI gate integrity — enforcement at the root, and lanes that can actually fail


tsouza:
## Why this epic exists

The seven `ch-capability-audit` epics each own a **product mechanism**. None owns the machinery that decides whether a release can ship, or whether any mechanism's evidence is trustworthy. That machinery is currently in poor shape, and its defects share a single root cause.

**The root cause: gates in this repo are scoped by file membership, not by the property they enforce.**

- `.github/scripts/perf-sentinel-obligation.mjs:69-72` sets `TRIGGER_FILES` to the literal pair `['internal/engine/query_settings_rules.go', 'internal/engine/spill.go']`, and `:106-108` fires on *any* edit to either. It cannot tell a memory-bounding setting from a workload tag or a lazy-materialization flag. That is what manufactured three content-free waiver receipts (#2832, #2833, #2849) instead of one scoping fix.
- `.github/scripts/mutation-phases.mjs` is half-converted to the `include_files` form. Three `./internal/chsql` legs use it; roughly twenty promql/logql/traceql legs still hand-maintain `exclude_files` regexes. `phase4-logql-parser` and `phase4-logql-lsyntax` exclude each other's files with **no catch-all**, so a new `foo.go` in that directory matches neither exclusion, is claimed by both legs, and trips `mutation-matrix.mjs`'s "claimed by N mutation phases" check. The `internal/traceql/ast` pair is structurally identical.
- `test/perf` never populates engine `SettingsRules` and both corpora pin `25.9-alpine`, below `join_spill`'s 26.4 floor — so all seven shipped settings mechanisms, including an `AutoSelect:true` memory guardrail descended from the #2364 production OOM, have structurally zero coverage. (Tracked as #2820 under epic #2778, where its lineage belongs, but it is the same class and is scheduled alongside these.)

Under the repo's own **"neutralized === missing"** and **first-green-after-adding-a-gate-is-vacuous** rules, every one of these is a real defect, not a testing nicety.

## Standing obligations

1. **Scope every gate by mechanism class, never by file membership.** A gate that fires on "someone edited this file" is a gate that will be waived.
2. **Prove activation, never observe green.** A new sentinel must be demonstrated to FAIL when the thing it guards is removed.
3. **One catch-all per mutation scope survives.** `mutation-matrix.mjs:363` fails the opposite way — "claimed by no mutation phase" — if every leg in a scope becomes include-shaped.
[...]

Issue #2883, opened 2026-09-01
ci: phase4-promql-lower mutation efficacy is below 95% on main, unrelated to any single PR


tsouza:
## Summary

`gremlins phase4-promql-lower (./internal/promql @ 95%)` and its `mutation`
aggregator fail on push-to-main repeatedly and independently of any specific
PR — this is pre-existing technical debt in `internal/promql/lower.go`, not a
regression from one change.

## Evidence

- Recent `mutation` workflow runs on `push` to `main` (before and unrelated to
  PR #2881): 33509695248, 33505244653, 33500024214, 33487778938 (schedule),
  33484461197, 33477213934, 33471969601 all show `gremlins phase4-promql-lower`
  + `mutation` as `failure`.
- On PR #2881 (`feat/ts-tag-groups`, cerberus issue #2750's grouping-keys
  slice), the same leg failed at `test_efficacy = 93.99%` (485 killed / 516
  tested, threshold 95%). Every surviving mutant's line falls **outside**
  that PR's actual diff to `lower.go` (which only adds a `TagGroups bool`
  struct field + one field-copy at lines 128-157) — cross-checked against
  `git diff` for the PR's `lower.go` hunk.
- The PR's mutation leg only ran a full-file scan (rather than the usual
  diff-scoped incremental one) because it also updated an existing test,
  `gremlins_kill_binary_lower_mixed_test.go` (a call-site signature update,
  zero added production lines) — per `mutation-matrix.mjs`'s documented
  "test/harness changes... fall back to the full selected phase" rule. That
  correctly exposed this pre-existing gap; it did not cause it.

## Why this matters

`mutation` is informational on an ordinary PR (see the header comment in
`.github/workflows/mutation.yml`) and is not in `required_status_checks`, so
it does not block merges today. But it IS in `RELEASE_INFORMATIONAL_CHECKS`
and is meant to be genuinely green by release time — and it is currently
red on `main` itself, repeatedly, which means any release cut right now
inherits this gap silently.

## Survivors (from a full-file run, phase4-promql-lower, 2026-09-01)

31 LIVED + 48 NOT COVERED = 79 total gaps against the 95% floor. Full list:

```
NOT COVERED CONDITIONALS_NEGATION at lower.go:197:9
LIVED INVERT_LOGICAL at lower.go:737:54
NOT COVERED CONDITIONALS_NEGATION at lower.go:871:22
LIVED INVERT_LOGICAL at lower.go:797:25
[...]

Issue #2695, opened 2026-08-28
ci(mutation): gate the timed-out share, which is excluded from both efficacy and mutants_total


tsouza:
Spun out of the diagnosis behind #2692, which fixes the budget root. This is a reporting root: even with a correct budget, a leg can leave most of its mutants unadjudicated and still report a high efficacy.

## The mechanism

gremlins excludes `TIMED_OUT` mutants from both sides of the verdict (`internal/report/report.go` in the `tsouza/gremlins` fork):

- line 95: `rep.tEfficacy = float64(rep.killed) / float64(rep.killed+rep.lived) * 100`
- line 179: `MutantsTotal: r.lived + r.killed + r.notViable`

So a timed-out mutant is in neither the ratio nor `mutants_total`. `.github/scripts/gremlins-threshold.mjs` gates on the reported efficacy, and `.github/scripts/mutation-run.mjs` gates on `mutants_total > 0`. Neither can see a mutant that timed out. Only the fully degenerate all-timeout case trips anything, and it trips `mutation-run.mjs`'s zero-mutant check rather than the threshold.

## This is live today, not hypothetical

CI run 33126692323 — every leg GREEN:

- `phase2-other`: `Killed: 77, Lived: 1, Not covered: 109 / Timed out: 262`. Efficacy 77/78 = **98.72%**, comfortably over the 95% bar, while **262 of 340 mutants received no verdict**.
- `phase2-builder`, same run: `Killed: 159, Lived: 2, Timed out: 158`.
- Run 33124139929, `phase2-builder`: `Killed: 159 / Timed out: 158`.

A leg reporting 98.72% on 23% of its mutants is the hollow-green shape `docs/test-strategy.md` forbids.

## Shape of the work

The report JSON has no top-level timed-out count, but `files[].mutations[].status` carries `TIMED OUT`. `gremlins-threshold.mjs` should count it and fail the leg when the timed-out share exceeds a named bound, so a budget that is too small can never again be indistinguishable from a pass.

## Why it needs its own risk budget

Adding that gate flips several currently-green `internal/chsql` legs red on its first run — `phase2-other` (262/340) and `phase2-builder` (158/319) at minimum. That is the honest reading of what those legs actually adjudicated, but it converts a lane that passes today into one demanding real work, and that work overlaps the ceiling question. Sequencing matters: the timeout-share gate and the ceiling/memory-cap work should land together, or the gate is red for however long the ceiling work takes.

tsouza:
[...]

PR #2911, opened 2026-09-01
fix(ci): stop scoring a starved mutation runner as a strong test suite

### LINKED ISSUE #2903: ci(mutation): timed-out mutants on phase4-promql-lower are budget starvation, and are scored as kills
Found while fixing #2883. Not a blocker for that issue's own fix (PR #2901
raises `phase4-promql-lower` by killing real survivors), but it is the reason
the leg OSCILLATES rather than being steadily red, and it makes the leg's green
runs mean less than they appear to.

## The claim

`gremlins-threshold.mjs` counts a timed-out mutant as a DETECTION, on the
premise — stated at the top of the file — that a mutant which exhausts the 15s
budget "did not do so because the machine was slow. It did so because the
mutation broke termination." That premise was checked against
`internal/chsql`'s recursive renderer in #2695, and it holds there.

It does not hold on `phase4-promql-lower`. On that leg the timeouts are
ordinary, terminating mutants whose `go test` child lost a race with the
compiler, and counting them as detections converts real survivors into fake
kills.

## Two runs, same 486 kills, three points of efficacy apart

| | red run [33542904091](https://github.com/tsouza/cerberus/actions/runs/33542904091) | green run [33551271099](https://github.com/tsouza/cerberus/actions/runs/33551271099) |
| --- | ---: | ---: |
| KILLED | 486 | 486 |
| LIVED | 31 | 15 |
| TIMED OUT | 4 | 36 |
| NOT COVERED | 48 | 48 |
| reported `test_efficacy` | 94.00% | 97.01% |
| verdict | FAIL | PASS |

Identical kill count. The entire three-point swing is 16 mutants moving from
LIVED into TIMED OUT.

Matching the two runs' per-mutant records by (mutator, source line text) rather
than by line number — the file grew 190 lines between the two revisions —
**18 of the red run's 31 LIVED mutants appear as TIMED OUT in the green run**.
Among them, the exact guard clauses PR #2901 adds assertions for:

```
INVERT_LOGICAL         | if rw.Identity || rw.Step <= 0 || rw.Start.IsZero() || rw.End.IsZero() {
INVERT_LOGICAL         | if companionValueColumn == "" || companionSuffixed == "" || companionBare == "" {
INVERT_LOGICAL         | if t == "" || t == s.HistogramTable {
INVERT_LOGICAL         | if len(rw.GroupBy) != 1 || !isIdentityColumnRef(rw.GroupBy[0], s.AttributesColumn) {
INVERT_LOGICAL         | if rw.DeltaPrefixAggregateInput != nil || len(rw.Variants) != 0 ||
CONDITIONALS_BOUNDARY  | return len(literalCompanionValueTables(s)) > 0
ARITHMETIC_BASE        | args := make([]chplan.Expr, 0, len(a.Grouping)*2)
```

## The green runner was the SLOWER one

| | red | green |
| --- | ---: | ---: |
| `Gathering coverage... done in` | 36.8s | 53.2s |
| `Mutation testing completed in` | 11m38s | 16m43s |

44% slower on comparable work, and it scored three points higher. That is the
signal inverted: **a more contended runner produces a greener leg**, because
every mutant it fails to adjudicate is booked as a detection.

## Why these mutants cannot be "broken termination"

They are short-circuit boolean guards in non-recursive classifier functions.
There is no loop or recursion for a negated conditional to unbound. Measured
directly on `main`'s `lower.go`, with the `lower.go:3574` `||` -> `&&` mutation
applied and the leg's own command run:

```
FAIL  github.com/tsouza/cerberus/internal/promql  1.996s
wall=22.07s
```

The test binary **runs to a verdict in 2.0 seconds**. The 22s of wall clock is
the recompile and link, on a machine under load. And gremlins' budget wraps the
whole `go test` child — `mutation.yml` says so itself: "The budget bounds the
whole `go test` child", against a measured "~6.6s per-mutant recompile+link+run"
for `internal/promql`.

So the 15s ceiling sits about 2.3x over a cost that is dominated by
COMPILATION, not by test execution. Any runner materially slower than the
machine that measurement came from pushes the slowest mutants over, and each
one that goes over is scored as a kill.

## Why the existing gates do not catch it

- The threshold gate reads `report.test_efficacy` = `killed / (killed + lived)`,
  which drops timeouts from both sides. Moving a survivor into TIMED OUT
  removes it from the denominator.
- `detectedEfficacy` recomputes with timeouts as detections and is strictly MORE
  permissive, and is emitted as a `::notice::` rather than gated on — correctly,
  since it can never fail where the reported ratio passed.
- `minCompletedMutants` only catches the fully degenerate case (#2692's budget
  collapse). 36 timeouts out of 585 is nowhere near it.

Neither formula can distinguish "the suite caught this by hanging on it" from
"the compiler was slow today", and on this leg it is the second one.

## Relationship to the closed issues

- #2695 built the timeouts-are-detections reading and shipped the gate. Its
  evidence was `internal/chsql`'s `phase2-builder` / `phase2-other`, where the
  reading is defensible. This is a counter-example from a different leg, not a
  claim that #2695 was wrong where it looked.
- #2694 proposed deriving `MUTANT_TIMEOUT_MAX` from measurement behind a memory
  cap, and was closed NOT_PLANNED on the argument that "those mutants do not
  terminate at any finite budget, so raising the ceiling would spend real CI
  minutes to reach the same verdict". For `phase4-promql-lower` that argument is
  false as measured above: these mutants terminate in ~2s of test time and would
  reach a real verdict at a larger budget. The closing rationale deserves
  re-examination on this evidence rather than a silent reopen.

## Shape of the work

Whatever the fix, the property to restore is that a leg's score cannot improve
because the runner got slower. Some directions, not a decision:

1. **Separate the build from the run.** The budget is meant to leash a runaway
   mutant's EXECUTION; it currently also leashes the compiler. Pre-building the
   test binary per mutant and timing only its execution would make the ceiling
   mean what its comment says it means, and would shrink it rather than grow it.
2. **Stop scoring what was not adjudicated.** Treat TIMED OUT as UNKNOWN and
   gate on the timed-out SHARE against a named bound per leg, instead of folding
   it into the kill side. #2695's own title — "gate the timed-out share" — is
   this; the implementation landed the detections reading instead.
3. **Distinguish the two causes.** A mutant that hangs and one that compiles
   slowly are separable by re-running the timed-out set once, serially, at a
   larger budget. Only the ones that time out again are non-terminating.

Option 1 is the root fix; 2 is the cheap guard that makes the leg honest in the
meantime.

## Acceptance

- A `phase4-promql-lower` run on a slow runner cannot report a higher efficacy
  than the same commit on a fast one, beyond genuine mutant-level nondeterminism.
- The 18 mutants tabulated above receive a real verdict (killed or lived) rather
  than a timeout, on both fast and slow runners.

tsouza:
## The defect

The mutation lane's per-mutant budget wraps the whole `go test` child — a
recompile, a link, and only then a run. On this repo's legs the compile is most
of that. Measured per mutant on an 8-core machine, against the flat 15s budget
the lane declared:

| scope             | compile + link | test run |
| ----------------- | -------------- | -------- |
| `internal/chsql`  | 12.7-14.4s     | 0.31s    |
| `internal/promql` | 8.4-8.7s       | 2.07s    |

So `TIMED OUT` recorded "the compiler was slow today", not "this mutation broke
termination" — and `gremlins-threshold.mjs` scored it as a detection. Because
gremlins drops a timeout from BOTH sides of its own ratio, starving the runner
*removed survivors from the denominator*, and the score went UP as the machine
got busier.

## Independent verification

Re-derived from the two runs' own `gremlins.json` artifacts rather than from the
issue's tables.

**The denominator arithmetic.** Same leg, same 486 kills:

| | red [33542904091](https://github.com/tsouza/cerberus/actions/runs/33542904091) | green [33551271099](https://github.com/tsouza/cerberus/actions/runs/33551271099) |
| --- | ---: | ---: |
| KILLED | 486 | 486 |
| LIVED | 31 | 15 |
| TIMED OUT | 4 | 36 |
| `test_efficacy` (gated on) | 94.00% | 97.01% |
| verdict | FAIL | **PASS** |
| with timeouts in the denominator | 93.28% | **90.50%** |

The whole three-point swing is 16 mutants moving `LIVED` -> `TIMED OUT`, on the
runner that gathered coverage in 53.2s where the other took 36.8s.

**The 18 survivors.** Matching the two revisions' per-mutant records by
`(mutator, column, trimmed source line)` — `lower.go` grew 190 lines between
them, so line numbers do not match — the red run's 31 survivors resolve as 18
`TIMED OUT` / 11 `LIVED` / 1 `KILLED` on the green run. Independently
reproduced; the 18 are short-circuit boolean guards and `make([]T, 0, n*2)`
capacity arithmetic, none of which can unbound a loop.

**The compile-vs-execution split.** With the `lower.go:3574` `||` -> `&&`
mutation applied and the leg's own cycle run:

```
MUTANT_COMPILE wall=15.77       # exceeds the entire 15s budget on its own
native_rate_temporality_test.go:125: instant native rate shape = ...
FAIL
MUTANT_RUN wall=1.62            # a KILL verdict, in 1.6s
```

The mutant is killed in 1.6s of test execution. The budget was spent by the
compiler before the test binary started. Repeated with a content change (not
`touch` — Go's build cache is content-keyed, so `touch` measures nothing):
`internal/promql` 8.4/8.6/8.7s, `internal/chsql` 12.7/14.4/12.9s.

**A second, stronger signature the issue does not name.** gremlins runs each
mutant with `-failfast`, so a KILLED mutant exits at its first failing test
while a survivor must run the suite to the end. Starvation therefore converts
*survivors* preferentially — the bias is monotone, not noise. Its endpoint, on
the green run [33522074818](https://github.com/tsouza/cerberus/actions/runs/33522074818):

| leg | killed | lived | timed out | reported | timeouts as non-kills |
| --- | ---: | ---: | ---: | ---: | ---: |
| `phase2-other` | 37 | 0 | 339 | **100.00%** | 9.84% |
| `phase2-builder` | 22 | 0 | 323 | **100.00%** | 6.38% |
| `phase2-compare` | 36 | 0 | 284 | **100.00%** | 11.25% |
| `phase2-range` | 16 | 0 | 284 | **100.00%** | 5.33% |

All four `internal/chsql` legs reported a perfect score with **zero survivors**,
over 1266 mutants of which 1230 were never adjudicated. `lived: 0` across four
independent legs is not a strong test suite; it is a runner that never ran long
enough to see a survivor.

## The direction chosen, and why

Two changes, one commit each. Both are needed: the second is the correctness
fix, the first is what keeps an honest leg from paying for it.

**1. Measure the budget instead of declaring it** (`mutation-run.mjs`). The
script times one mutant's worth of work before gremlins starts — appends a byte
to the largest production file in the leg's scope, runs the scope's tests
exactly as gremlins runs them for a mutant (`-count=1 -failfast`), restores the
file — and scales the result by the concurrent-mutant count and a named
headroom (2, against the 44% runner-to-runner swing measured above). The edit is
load-bearing: Go's build cache is keyed on content, so timing an unedited
package times a cache hit.

`MUTANT_TIMEOUT_MIN` (15s, the value the lane already ran on) is the floor a
measurement may raise but never lower, so an unmeasurable cycle leaves the lane
exactly where it was. `MUTANT_TIMEOUT_MAX` (120s) is the ceiling, so one
pathological measurement cannot spend the job's whole wall clock on hung
mutants.

**2. Score a timeout as unadjudicated** (`gremlins-threshold.mjs`). Efficacy
becomes `killed / (killed + lived + timedOut)` — the kill rate over every mutant
the runner *attempted* to adjudicate. The rule it encodes: **the score must be
monotone non-increasing in the timeout count.** Moving a mutant into `TIMED OUT`
can never raise the number the gate reads.

### The failure direction, stated plainly

**This change biases toward RED under starvation.** A runner too slow to
adjudicate its mutants now fails the leg instead of scoring it higher. That is
the safe direction because the two errors are not symmetric: a false red costs
one investigation and names its own cause in the error message, while the false
green it replaces was *indistinguishable from a strong test suite* and had
already let `phase2-*` report 100.00% on 3% of its mutants for weeks. It is also
what makes measurement-derived budgets safe to get wrong in the tight direction
— an undersized budget is now visible.

## Verification

**The gate fails where it used to pass, on the real artifacts:**

```
red/gremlins.json     FAIL  efficacy 93.28% < 95% (killed 486 of 521, lived 31, timed out 4)
green/gremlins.json   FAIL  efficacy 90.50% < 95% (killed 486 of 537, lived 15, timed out 36)
phase2-builder        FAIL  efficacy  6.38% < 95% (killed 22 of 345, lived 0, timed out 323)
```

The starved run now scores **below** the run it beat, and the leg that reported
100.00% reports 6.38%.

**It still passes a genuinely-killing run:** a leg with 990 killed / 32 lived /
0 timed out passes, and emits no timeout notice at all.

**Both directions are pinned by tests that fail if the behaviour regresses**, and
each was proven to fail rather than assumed to:

- reverting the formula to timeouts-as-detections: **5 of 13** gate tests fail,
  including `the starved run of a leg scores BELOW the run it was compared
  against` and the `moving any mutant into TIMED OUT never raises the score`
  property;
- making the probe stop editing the source (so it times a cache hit): `the probe
  edits the largest production file, and restores it exactly` fails;
- dropping the concurrent-mutant factor: `the budget scales with the measured
  cycle and the concurrent-mutant count` fails.

The `mutation-run` fixture fakes `go` as well as `gremlins`, because a probe
that silently measured nothing would fall back to the floor and every budget
assertion would still pass against a script that never probed.

**End-to-end against real packages**, with a stub gremlins recording its argv:

```
::notice::measured per-mutant cycle 1.3s for ./internal/spansscan; x8 x2 = 22s -> budget 22s
::notice::measured per-mutant cycle 18.3s for ./internal/chsql; x8 x2 = 294s, clamped -> budget 120s
```

The working tree is byte-identical afterwards.

## Does #2694 / #2695's rejected rationale still hold?

**#2694 (`NOT_PLANNED`) does not.** It was closed on the measurement
"`internal/chsql` recompiles, links and runs in 1.04s at eight cores", giving a
6x-15x margin under 15s. That figure does not reproduce: forcing the recompile a
mutant forces, `internal/chsql` costs 12.7-14.4s of compile+link and 0.31s of
test run. The closing argument — "those mutants do not terminate at any finite
budget" — is therefore false on the very leg it was measured on. Its *other*
concerns survive and are honoured here: worst-case wall clock is why
`MUTANT_TIMEOUT_MAX` exists, and this PR does not raise the budget blindly, it
measures it.

**#2695 is half-vindicated.** Its title — "gate the timed-out share" — is what
this PR implements. Its shipped implementation went the other way, and the
detections reading is what #2903 refutes. Its sequencing warning is the sharpest
thing in it: "the timeout-share gate and the ceiling work should land together,
or the gate is red for however long the ceiling work takes." That is exactly why
both halves are in this PR.

The memory cap #2694 also proposed remains unimplemented and unclaimed here; the
budget still carries the OOM bound alone, now bounded by `MUTANT_TIMEOUT_MAX`
rather than by 15s.

## What this makes visible

Legs whose reported efficacy was carried by timeouts will now report what they
actually adjudicated. On the last full green run that is every `phase2-*` leg
and several `phase4-promql-*` legs. The measured budget is what those legs get
in exchange; where a leg still lands under its bar with a healthy timeout count,
the remedy is the one `docs/test-strategy.md` already prescribes — kill the
survivors — never a wider `exclude_files` and never a lower bar. `mutation` is
informational on PRs and in `RELEASE_INFORMATIONAL_CHECKS`, so nothing is
blocked while that work happens.

#2883's exit criterion is two consecutive green push-to-main `mutation` runs
precisely because one green proves nothing today. This is the change that makes
a single green mean mutants were killed.

The remaining root — gremlins collapsing an overall deadline and a run-only one
into a single number, so `TIMED OUT` still cannot distinguish a hang from a slow
compile — is #2910, filed under epic #2891 with the same milestone. That work is
in the `tsouza/gremlins` fork and needs its own authorization.

Closes #2903

### DIFF
--- a/.github/scripts/README.md
+++ b/.github/scripts/README.md
@@ -915,7 +915,12 @@ what actually runs.
     `run_heavy=true`, logged via `::notice::`, never a hard failure); `1` on an
     unrecognised `MODE`.
 - **`gremlins-threshold.mjs`** — `mutation.yml`, the
-  `enforce efficacy threshold` step.
+  `enforce efficacy threshold` step. Gates on `killed / (killed + lived +
+  timedOut)`, the kill rate over every mutant the runner ATTEMPTED to
+  adjudicate, rather than on gremlins' own `killed / (killed + lived)`. A
+  timed-out mutant is unadjudicated, not detected, so moving one into
+  `TIMED OUT` can never raise the score and a starved runner reads as red
+  rather than as a flattering green (#2903).
   - Env: `REPORT` (default `gremlins.json`), `THRESHOLD` (a number).
   - Exit: `0` when efficacy is `>=` threshold, `1` when below.
 - **`mutation-phases.mjs`** — data, imported by `mutation-matrix.mjs`. The single
@@ -961,9 +966,20 @@ what actually runs.
   phase. A non-empty `DIFF_REF` first invokes gremlins' native merge-base
   `--diff` filter; if that report executed zero mutants (a comment-only or
   otherwise mutation-free edit), it deletes the report and reruns the phase in
-  full rather than letting a hollow incremental run report green.
-  - Env: `SCOPE`, `REPORT`, `MUTANT_TIMEOUT_MAX` (required); `WORKERS`,
-    `EXCLUDE_FILES`, `DIFF_REF` (optional).
+  full rather than letting a hollow incremental run report green. Also derives
+  the per-mutant budget by MEASURING this leg's own recompile+link+run cycle on
+  this runner — it edits a byte of the scope's largest production file, times
+  the scope's tests as gremlins runs them for a mutant, restores the file, and
+  scales the result by the concurrent-mutant count and a named headroom. The
+  flat budget it replaced bounded a `go test` child that is 80-98%
+  COMPILATION, so a contended runner starved ordinary mutants into `TIMED OUT`
+  (#2903).
+  - Env: `SCOPE`, `REPORT`, `MUTANT_TIMEOUT_MIN`, `MUTANT_TIMEOUT_MAX`
+    (required); `WORKERS`, `EXCLUDE_FILES`, `DIFF_REF` (optional).
+    `MUTANT_TIMEOUT_MIN` is the floor the measurement may raise but never
+    lower, so an unmeasurable cycle leaves the lane where it was;
+    `MUTANT_TIMEOUT_MAX` is the ceiling, so one pathological measurement
+    cannot spend the job's whole wall clock on hung mutants.
   - Exit: `0` when the final report executed at least one mutant; `1` on a bad
     env value, a pre-existing report at `REPORT`, a gremlins invocation
     failure, or zero mutants surviving the full fallback.
--- a/.github/scripts/gremlins-threshold.mjs
+++ b/.github/scripts/gremlins-threshold.mjs
@@ -12,42 +12,72 @@
 // no measured-vs-threshold numbers. Gating after the report is written keeps
 // both. Do not "simplify" this by passing --threshold-efficacy.
 //
-// Reproduces the original bash exactly:
+// A TIMED-OUT MUTANT IS AN UNADJUDICATED MUTANT, AND COUNTS AGAINST THE SCORE.
 //
-//   measured=$(jq -r '.test_efficacy' gremlins.json)
-//   if awk 'BEGIN { exit (m < t) ? 1 : 0 }'; then  # m < t  -> fail
-//     ::notice:: measured >= threshold
-//   else
-//     ::error::  measured <  threshold ; exit 1
-//
-// The comparison is `measured < threshold` -> FAIL (strict less-than),
-// i.e. `measured >= threshold` passes. Matches the awk `m < t` semantics.
-//
-// A TIMED-OUT MUTANT IS A DETECTED MUTANT, and gremlins counts it as neither
-// killed nor lived.
+// This is the property the whole gate rests on, so it is worth stating as a
+// rule rather than as a formula: THE SCORE MUST BE MONOTONE NON-INCREASING IN
+// THE TIMEOUT COUNT. Moving one mutant from any other status into TIMED OUT may
+// never raise the number this gate reads. A gate that violates that rule pays a
+// slower runner in efficacy points, and a green run then proves that the
+// machine was busy rather than that the tests are strong.
 //
 // gremlins excludes TIMED_OUT from BOTH sides of its own verdict
 // (internal/report/report.go: `tEfficacy = killed / (killed + lived)` and
-// `MutantsTotal = lived + killed + notViable`), so a leg where most mutants time
-// out reports a confident-looking ratio computed over the few that did not. Run
-// 33156989462's `phase2-builder` reported 97.50% over 40 of 399 mutants; 279
-// timed out and appear in no number the gate could previously read.
+// `MutantsTotal = lived + killed + notViable`), so its ratio VIOLATES that rule
+// in the worst way — a timeout leaves the denominator entirely, so starving the
+// runner raises the score. This gate therefore recomputes efficacy over every
+// mutant the runner ATTEMPTED to adjudicate, timeouts included:
 //
-// The orthodox reading, and the one the measurements support, is that those 279
-// were KILLED. `internal/chsql` recompiles, links and runs in 1.04s at eight
-// cores and 2.42s at one, against a 15s budget — a 6x-15x margin — so a mutant
-// that exhausts it did not do so because the machine was slow. It did so because
-// the mutation broke termination, which is exactly what negating a conditional
-// in a RECURSIVE renderer does: the mutator statistics put the timeouts in
-// CONDITIONALS_NEGATION (75%) and CONDITIONALS_BOUNDARY (93%), not in
-// INVERT_LOOPCTRL (4 mutants in the whole leg). The suite caught each one by
-// hanging on it.
+//   efficacy = killed / (killed + lived + timedOut)
 //
-// Efficacy is therefore recomputed here with timeouts counted as detections.
-// That is MORE permissive than gremlins' own ratio, which is why it is paired
-// with minCompletedMutants: the one thing this must never wave through is
-// #2692's budget collapse, where nothing completed and the timeouts prove
-// nothing about the tests.
+// which is monotone by construction, and is never above the ratio gremlins
+// reported.
+//
+// Why the opposite reading was wrong (cerberus issue #2903)
+// --------------------------------------------------------
+// This gate previously counted a timeout as a DETECTION, on the premise that a
+// mutant which exhausts the per-mutant budget "did not do so because the
+// machine was slow — it did so because the mutation broke termination". The
+// premise was checked against `internal/chsql`'s recursive renderer and looked
+// sound there. It is false, and it is false on that very leg.
+//
+// The budget wraps the whole `go test` child, which is a RECOMPILE, a LINK and
+// then a run. Measured per mutant, on an 8-core machine, against the 15s budget
+// those legs ran under:
+//
+//   internal/chsql   compile+link 12.7-14.4s   test run 0.31s
+//   internal/promql  compile+link  8.4-8.7s    test run 2.07s
+//
+// The compiler is 80-98% of the budget and the mutant's own execution is a
+// rounding error, so TIMED OUT records "the compiler was slow today", not
+// "this mutation stopped the suite terminating".
+//
+// Two signatures in the real reports confirm it:
+//
+//   - Runs 33542904091 (red) and 33551271099 (green) on `phase4-promql-lower`
+//     recorded the SAME 486 kills. The whole 94.00% -> 97.01% swing was 16
+//     mutants moving LIVED -> TIMED OUT, and the green runner was the SLOWER
+//     one (coverage 53.2s vs 36.8s). Matching mutants across the two revisions
+//     by source text, 18 of the red run's 31 survivors — short-circuit boolean
+//     guards and slice-capacity arithmetic, none of which can unbound a loop —
+//     were booked as detections on the green run.
+//
+//   - gremlins runs each mutant with `-failfast`, so a KILLED mutant exits at
+//     its first failing test while a LIVED mutant must run the suite to the
+//     end. Budget starvation therefore converts SURVIVORS preferentially. Run
+//     33522074818 shows the endpoint of that bias: all four `internal/chsql`
+//     legs reported 100.00% with `lived: 0`, over 1266 mutants of which 1230
+//     timed out.
+//
+// Counting timeouts as detections scored that last case — a leg that
+// adjudicated 3% of its mutants and found no survivor because it never ran long
+// enough to see one — as a perfect leg.
+//
+// The paired defence is the per-mutant budget itself: .github/scripts/
+// mutation-run.mjs now MEASURES the leg's own recompile+link+run cycle and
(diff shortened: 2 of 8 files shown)


tsouza:
## Full-matrix result

This PR edits `mutation-phases.mjs`' `HARNESS_PATHS`, so it swept all 30 legs
([run 33572151680](https://github.com/tsouza/cerberus/actions/runs/33572151680)).
That makes it the first honest reading of the whole lane, and it settles both
halves of the change against real runners rather than fixtures.

**Timeouts across the matrix: 1483 -> 64.** The measured budget did the work the
flat 15s could not.

### The hollow greens became earned greens

All four `internal/chsql` legs previously reported 100.00% while adjudicating
6-11% of their mutants. With a budget sized to their ~14s recompile they
adjudicate everything, and they clear the bar on merit:

| leg | before | after |
| --- | --- | --- |
| `phase2-builder` | 100.00% over 22 of 345 | **98.01%**, 0 timed out |
| `phase2-compare` | 100.00% over 36 of 320 | **99.06%**, 0 timed out |
| `phase2-range` | 100.00% over 16 of 300 | **99.33%**, 0 timed out |
| `phase2-other` | 100.00% over 37 of 376 | **96.54%**, 1 timed out |

This is the outcome the change was aimed at and it is worth stating plainly:
those legs were not weak, they were unmeasured.

### The leg the issue was filed about

`phase4-promql-lower`, which oscillated 94.00% / 97.01% on the same 486 kills
depending on how busy the runner was:

| | red 33542904091 | green 33551271099 | this PR |
| --- | ---: | ---: | ---: |
| KILLED | 486 | 486 | 504 |
| LIVED | 31 | 15 | 33 |
| TIMED OUT | 4 | **36** | **0** |
| efficacy | 94.00% | 97.01% | **93.85%** |
| verdict | FAIL | PASS | FAIL |

Zero timeouts, so gremlins' own ratio and the gated ratio are now the *same
number* — there is nothing left unadjudicated for the two to disagree about. The
18 mutants that were being booked as detections have real verdicts. The leg
fails because it has 33 genuine survivors, which is #2883's actual work.

### Seven legs now report a real shortfall

| leg | killed | lived | timed out | efficacy | reported before |
| --- | ---: | ---: | ---: | ---: | ---: |
| `phase4-promql-lower` | 504 | 33 | 0 | 93.85% | 96.30% |
| `phase4-promql-g` | 389 | 21 | 0 | 94.88% | 95.26% |
| `phase4-logql-lsyntax` | 449 | 13 | 25 | 92.20% | 97.19% |
| `phase5-qlcommon` | 299 | 14 | 17 | 90.61% | 95.53% |
| `phase4-traceql-ast` | 265 | 12 | 8 | 92.98% | 95.67% |
| `phase4-logql-other-b` | 496 | 23 | 6 | 94.48% | 95.57% |
| `phase4-logql-parser` | 74 | 2 | 2 | 94.87% | 97.37% |

Filed as #2913, under epic #2891 in the same milestone. `mutation` is
informational on a PR and in `RELEASE_INFORMATIONAL_CHECKS`, so nothing is
blocked while that work happens — and every one of these reds names its own
cause in the error message, which the greens they replace did not.

Four of the five legs still carrying timeouts are the `workers: 1` parser legs,
where a loop-advance inversion in a hand-written scanner genuinely does not
terminate — the case #2695's reasoning actually fits. Asserting that from the
status alone is exactly what this PR argues nobody can do, so #2910 (splitting
gremlins' overall deadline from its run-only one) is what settles it.

The remaining 23 legs pass, and the 9 that already had zero timeouts report
byte-identical numbers to their last green run — the change is inert where there
was nothing to fix.

PR #2939, opened 2026-09-02
ci: repoint the release-gate drift lane at the ruleset API and make it able to run

### LINKED ISSUE #2937: ci: the release-gate drift lane reads a branch-protection endpoint that no longer exists
`main`'s legacy branch protection has been replaced by ruleset **#22113683**, and the legacy object
is deleted. `GET /repos/tsouza/cerberus/branches/main/protection` now returns
`404 Branch not protected` (verified 2026-09-02). Everything in this repository that reads or names
that endpoint is now pointing at nothing.

## What is actually broken

`.github/scripts/release-gate-drift.mjs:270` reads the dead endpoint:

```js
const protection = await apiJson(
  `${apiBase}/repos/${repo}/branches/${encodeURIComponent(branch)}/protection`,
  ...
```

That call feeds `protectionContexts()`, which feeds **both** of the drift lane's live-comparison
directions:

- **Direction A** (`protectionDrift`) — a branch-protection-required context that appears in neither
  `RELEASE_REQUIRED_CHECKS` nor `RELEASE_INFORMATIONAL_CHECKS`, i.e. a lane certified by absence on
  the maintenance publish path.
- **Direction C** (`pinnedProtectionDrift`) — set EQUALITY between the live contexts and
  `branchProtectionContexts` in `test/regression/release_required_checks_test.go`. That pin is the
  thing every in-tree gate reasons off, because reading the live setting needs repo-admin rights no
  test token carries. Direction C is the only mechanism in the repository that can tell the pin from
  the reality.

Neither can run now. The lane fails closed rather than passing vacuously, which is the right failure
mode, but the pin is unverifiable until the read is repointed.

## Second, older half: the lane has never once executed that comparison

`release-gate-drift` has been red on every scheduled run back to at least 2026-08-04, and not for the
ruleset reason. `BRANCH_PROTECTION_TOKEN` is unset, so the script aborts before the API call:

```
##[error]BRANCH_PROTECTION_TOKEN is unset — the live required-context comparison cannot run
```

(Run 33501351829, 2026-09-01; same failure on 32818568978, 32108095937, 31467717705, 30892700334.)

So the drift lane is a gate that has never produced a verdict. That is what let `update-golden-guard`
become a required context on 2026-08-24 and stay absent from `branchProtectionContexts` until
`ci/update-golden-guard-merge-group` pinned it by hand: Direction C would have caught it on the very
next Monday, had it been able to run. Repointing the URL without also provisioning the credential
leaves the lane exactly as silent as it is today.

## Third half: seven sites name the dead command as the source of truth

- `CLAUDE.md:86` — invariant 1, which names it as *the* source of truth for the required-check set
- `docs/agent-workflow.md:74`
- `docs/compatibility.md:385`
- `.github/workflows/codeql.yml:25`
- `.github/scripts/release-gate-drift.mjs:173` (the remediation hint it prints on drift)
- `.github/scripts/generated-baseline-structural-guard.mjs:57`
- `.github/scripts/post-merge-golden-drift.mjs:14`
- `test/regression/release_required_checks_test.go:44`

Each tells the next reader to run a command that 404s.

## Fix shape

1. Repoint the live read at `GET /repos/{owner}/{repo}/rules/branches/{branch}`, extracting
   `[.[] | select(.type=="required_status_checks") | .parameters.required_status_checks[].context]`.
   Note the shape difference: the rules endpoint returns an array of rule objects, not a single
   protection object, and a context there is `{context, integration_id}` rather than a bare string —
   `protectionContexts()` has to change, not just the URL.
2. Provision `BRANCH_PROTECTION_TOKEN` (repository administration:read), or the repoint changes
   nothing observable.
3. Update the eight prose/comment sites to the rules command.
4. Pin the shape: the drift script's parser needs a unit test over a recorded rules-endpoint
   response, so a future endpoint change fails a test rather than a Monday cron nobody reads.

## Not a regression of the merge-queue work

`ci/update-golden-guard-merge-group` verified its `branchProtectionContexts` change by hand against
the live ruleset (17 contexts, set-equal). This issue is about restoring the automated check, which
was already inert before the ruleset migration.

tsouza:
## What broke

`main`'s legacy branch protection was replaced by repository ruleset **#22113683** and the legacy
object deleted, so

```console
$ gh api repos/tsouza/cerberus/branches/main/protection
gh: Branch not protected (HTTP 404)
```

`release-gate-drift.mjs` read exactly that endpoint, which killed the two directions that need a
live set: **A** (a required context in neither `RELEASE_REQUIRED_CHECKS` nor
`RELEASE_INFORMATIONAL_CHECKS`) and **C** (set equality between the live set and
`branchProtectionContexts`).

## The worse half: the comparison had never once run

The protection endpoint needed repository `administration:read`, which the workflow token cannot
receive, so the step carried a `BRANCH_PROTECTION_TOKEN` secret. That secret was never provisioned.
Every scheduled run from 2026-08-04 to 2026-09-01 aborted on `BRANCH_PROTECTION_TOKEN is unset`
before reaching the API, so the only automated check that the repo's pin matches reality had never
produced a verdict — which is how `update-golden-guard` became a required context on 2026-08-24 and
stayed absent from `branchProtectionContexts` for a week.

Repointing the URL alone would have left the lane exactly as silent.

## What this changes

**Repointed at the ruleset API.** The live read is now
`GET /repos/{owner}/{repo}/rules/branches/{branch}`. The shape is different in three ways and all
three are handled rather than assumed away:

- the response is a flat **list of rules**, not one protection object, so rules are selected by
  `type === 'required_status_checks'` — `deletion`, `non_fast_forward`, `pull_request` and the
  `merge_queue` rule about to appear alongside them are ignored **by type**, never by position;
- a context is `{context, integration_id}`, not a bare string;
- **more than one ruleset can govern one branch.** This repo already runs a second ruleset over
  `refs/heads/release/*.x`, so the endpoint can return two `required_status_checks` rules with
  different `ruleset_id`s. The governing set is the **union** across all of them, deduplicated;
  reading only the first match would under-report the live set and land as a bogus direction-C
  "pinned but not required" on a context that *is* required. The contributing ids are carried
  through into the drift messages and the step summary, so a remediation hint names the ruleset to
  edit — in the plural when two are in force.

Every degenerate payload throws rather than degrading to an empty set: a non-list body, a branch
with no `required_status_checks` rule at all, malformed `parameters`, a non-string or empty context,
and a rule requiring zero contexts. An empty live set makes directions A and C vacuously green,
which is the failure this lane exists to prevent.

**No token, because none is needed.** `rules/branches/{branch}` carries no `administration:read`
requirement — it is served to any credential that can read the repository, and on this public repo
to no credential at all:

```console
$ curl -s -o /dev/null -w '%{http_code}\n' \
    https://api.github.com/repos/tsouza/cerberus/rules/branches/main
200
```

So the secret is **deleted, not re-provisioned**, and the lane now runs on `${{ github.token }}`.
There is no absent-secret path left to fail or skip, which is the strongest available answer to
"how does it fail loudly when the credential is missing": there is no credential to be missing. The
regression pin enforces that structurally — `TestReleaseGateDriftDetectorRunsLiveAndNotSelfTestOnly`
now fails on **any** `secrets.` reference in `release-gate-drift.yml` and on **any** `if:` guard on
any step, because `if:` on a secret is how "no token, skip" is spelled in workflow YAML.

**A second real bug the repointing exposed.** With the read fixed, direction B immediately accused
`forbid-deferral` — a healthy required lane with a green push-to-main run — of being dead. The scan
read only the first page of each commit's check-runs. `2733b38c7` carries 146 check-runs across 26
check suites, and `forbid-deferral` sits on page 2. The detector's first real verdict in a month
would have been a false accusation against a working gate, which is how a rot detector gets
dismissed as noise a second time. `apiPaged()` now walks check-runs and combined statuses to
exhaustion, bounded, throwing rather than silently returning a prefix.

**Shape pinned.** The self-test carries a **recorded** `rules/branches/main` response (captured
2026-09-02, trimmed to the fields the parser reads) and asserts it yields the real 17 contexts from
ruleset #22113683 with the matrix name intact. A further endpoint change now fails a test on a PR,
not a Monday cron nobody reads. `TestReleaseGateDriftDetectorRunsLiveAndNotSelfTestOnly` also pins
the endpoint itself, judged on the URLs the script *builds* (every request is interpolated off
`${apiBase}`) so the header comment may go on explaining which endpoint died.

**Eight stale sites repointed**, `CLAUDE.md` invariant 1 included — it named the 404ing command as
*the* source of truth for the required-check set. The replacement everywhere is:

```bash
gh api repos/tsouza/cerberus/rules/branches/main \
  --jq '[.[] | select(.type == "required_status_checks")
             | .parameters.required_status_checks[].context] | unique[]'
```

`CLAUDE.md`, `docs/agent-workflow.md`, `docs/compatibility.md`, `.github/workflows/codeql.yml`,
`.github/scripts/release-gate-drift.mjs` (the remediation hint, which pointed at the
protection-contexts `POST` — under a ruleset the edit is a whole-ruleset `PUT`),
`.github/scripts/generated-baseline-structural-guard.mjs` and
`.github/scripts/post-merge-golden-drift.mjs` (both cite `strict: false`, now
`strict_required_status_checks_policy: false`), and
`test/regression/release_required_checks_test.go`. The `.github/scripts/README.md` entry is rewritten
to match.

`branchProtectionContexts`, `protectionDrift` and `pinnedProtectionDrift` keep their names: what
they denote — the set of contexts a merge into `main` must pass — did not change, only the endpoint
that reports it. Each doc comment now says so explicitly.

## Evidence

**It detects a real drift, unprompted.** The lane's first working run finds the drift #2938 is
independently fixing, from both directions and nothing else:

```console
$ GITHUB_TOKEN=… GITHUB_REPOSITORY=tsouza/cerberus node .github/scripts/release-gate-drift.mjs
- required contexts in force via ruleset #22113683: **17** (pinned: **16**)
- commits scanned for posted lane names: **20**
- ❌ update-golden-guard: ruleset-REQUIRED context in neither RELEASE_REQUIRED_CHECKS nor
  RELEASE_INFORMATIONAL_CHECKS — the release publishes without waiting for it. …
- ❌ update-golden-guard: required by ruleset #22113683 on `main` but absent from
  branchProtectionContexts. …
```

Direction B is clean in the same run — the pagination fix is what makes that true. Once #2938 lands
(it adds `update-golden-guard` to both `RELEASE_REQUIRED_CHECKS` and the pin) all three directions
go clean; this PR deliberately does not touch the pin, so the drift stays visible and belongs to the
PR that caused it.

**Each new assertion fails when violated**, checked by mutating the tree and re-running
`TestReleaseGateDriftDetectorRunsLiveAndNotSelfTestOnly`:

| mutation | result |
| --- | --- |
| re-add `BRANCH_PROTECTION_TOKEN: ${{ secrets.… }}` to the step | FAIL — "references a repository secret again" |
| add `if: env.SOME_TOKEN != ''` to the live step | FAIL — "conditions a step on …" |
| point the read back at `…/branches/${…}/protection` | FAIL — "builds a branch-protection request again" |
| none | pass |

**It runs in the workflow, not just in a shell with a human's `gh` token.** Verified by dispatching
`release-gate-drift.yml` on this branch — see the run linked in the comment below, which reaches the
API on `${{ github.token }}` and reports the drift above.

**Local, targeted:** `node .github/scripts/release-gate-drift.mjs --self-test` (green),
`go test ./test/regression/ -count=1` (green), `just lint-actions`, `markdownlint-cli2` on the four
touched docs, `CHECK=all node .github/scripts/forbid-skip.mjs`,
`node .github/scripts/assert-gate-suites-wired.mjs`, `node .github/scripts/ci-lane-contract.mjs`.

Closes #2937

### DIFF
--- a/.github/scripts/README.md
+++ b/.github/scripts/README.md
@@ -1116,35 +1116,57 @@ what actually runs.
   `RELEASE_REQUIRED_CHECKS` + `RELEASE_INFORMATIONAL_CHECKS` out of release.yml
   itself (one copy of the data, one parser; an empty parse throws rather than
   comparing against nothing) and checks two directions the preflight structurally
-  cannot see from the inside. PROTECTION DRIFT: a live branch-protection required
-  context in neither list is a lane every PR must pass and the release does not
-  wait for — the dangerous direction, and invisible to an allow-list of names to
-  wait for. LANE DRIFT: a required name that posted no check-run anywhere in the
-  scanned commit window no longer matches a lane, so the next release waits out
-  its full window and aborts mid-publish. The window spans many commits because a
-  single commit's check-runs are a subset of the lane inventory (a docs-only push
-  skips `check`); absent from every commit means dead, not skipped. PIN DRIFT:
-  set equality between the live contexts and `branchProtectionContexts`, the
-  repo's own pin of them in `test/regression/release_required_checks_test.go`.
-  Every in-tree gate reasons off that pin because the endpoint needs repo-admin
-  rights no test token has, so a pinned-but-unenforced context is a lane that
-  posts green on every PR while contributing nothing to the merge decision — and
-  nothing else in the tree can tell. Pure exported `parseBlockScalar` /
-  `parseCheckLists` / `parsePinnedContexts` / `protectionDrift` / `laneDrift` /
-  `pinnedProtectionDrift` / `readBranchProtection` plus a `--self-test` the job
-  runs first. The self-test rejects missing credentials, HTTP 403, empty, and
-  malformed protection data rather than treating any as a clean empty set.
-  - Env: `GITHUB_TOKEN` (least-privilege commit/check/status reads),
-    `BRANCH_PROTECTION_TOKEN` (dedicated repository-administration read
-    credential, confined to the protection request), `GITHUB_REPOSITORY`,
-    `GITHUB_API_URL` (default
+  cannot see from the inside. PROTECTION DRIFT: a live required context in
+  neither list is a lane every PR must pass and the release does not wait for —
+  the dangerous direction, and invisible to an allow-list of names to wait for.
+  LANE DRIFT: a required name that posted no check-run anywhere in the scanned
+  commit window no longer matches a lane, so the next release waits out its full
+  window and aborts mid-publish. The window spans many commits because a single
+  commit's check-runs are a subset of the lane inventory (a docs-only push skips
+  `check`); absent from every commit means dead, not skipped. It also spans every
+  PAGE of each commit's check-runs — a busy `main` commit carries ~150 across
+  ~26 check suites, and a first-page-only read reported the healthy
+  `forbid-deferral` lane as dead. PIN DRIFT: set equality between the live
+  contexts and `branchProtectionContexts`, the repo's own pin of them in
+  `test/regression/release_required_checks_test.go`. Every in-tree gate reasons
+  off that pin because a Go test does not reach the network, so a
+  pinned-but-unenforced context is a lane that posts green on every PR while
+  contributing nothing to the merge decision — and nothing else in the tree can
+  tell.
+  The live set is read from `GET /repos/{owner}/{repo}/rules/branches/{branch}`,
+  the RULESET API: this repository's legacy branch-protection object was replaced
+  by a ruleset and deleted, so `branches/{branch}/protection` answers `404`. That
+  endpoint returns the flattened rules in force, each tagged with the
+  `ruleset_id` that contributed it, and more than one ruleset can match a branch
+  (a second one governs `refs/heads/release/*.x`) — so the governing set is the
+  UNION across every `required_status_checks` rule, and rule types the script
+  does not model (`deletion`, `non_fast_forward`, `pull_request`, `merge_queue`)
+  are ignored by type rather than by position. It needs no elevated credential:
+  the read is served to `${{ github.token }}`, so the lane depends on NO secret.
+  It used to depend on a `BRANCH_PROTECTION_TOKEN` that was never provisioned,
+  which left the pin comparison without a single verdict from 2026-08-04 to
+  2026-09-01; `TestReleaseGateDriftDetectorRunsLiveAndNotSelfTestOnly` now
+  rejects any secret reference or `if:` guard in the workflow for that reason.
+  Pure exported `parseBlockScalar` / `parseCheckLists` / `parsePinnedContexts` /
+  `protectionDrift` / `laneDrift` / `pinnedProtectionDrift` / `describeRulesets`
+  / `rulesetContexts` / `apiPaged` / `readRequiredContexts` plus a `--self-test`
+  the job runs first, and `ci.yml` runs on every PR. The self-test pins a
+  RECORDED rules-endpoint response so a further shape change fails a test rather
+  than a Monday cron nobody reads, and rejects a missing token, HTTP 403, a
+  non-list payload, a branch with no `required_status_checks` rule, malformed
+  parameters, a non-string context and a zero-context rule rather than treating
+  any of them as a clean empty set.
+  - Env: `GITHUB_TOKEN` (the only credential — least-privilege workflow token,
+    used for the ruleset read as well as the commit/check/status reads),
+    `GITHUB_REPOSITORY`, `GITHUB_API_URL` (default
     `https://api.github.com`), `DRIFT_BRANCH` (default `main`), `DRIFT_HISTORY`
     (default 20 commits), `RELEASE_WORKFLOW` (default
     `.github/workflows/release.yml`), `PROTECTION_PIN_FILE` (default
     `test/regression/release_required_checks_test.go`).
   - Exit: `0` when all three directions are clean; `1` on any drift, an API read
-    failure, an unparseable release.yml, or protection reporting zero contexts
-    (fails closed — unreadable protection is not a clean bill of health).
+    failure, an unparseable release.yml, or the rules endpoint reporting no
+    `required_status_checks` rule or zero contexts (fails closed — an unreadable
+    live set is not a clean bill of health).
 - **`release-preflight.mjs`** — `release.yml`, the `preflight` job. The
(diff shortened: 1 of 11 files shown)


tsouza:
## Proof it runs in the workflow, and that it detects a real drift

`release-gate-drift.yml` dispatched on this branch:
**[run 33641391510](https://github.com/tsouza/cerberus/actions/runs/33641391510)**

The step env is the whole point — there is no second credential, and the job still reached the API:

```text
Run node .github/scripts/release-gate-drift.mjs
env:
  GITHUB_TOKEN: ***
  GITHUB_REPOSITORY: tsouza/cerberus

##[error]update-golden-guard: ruleset-REQUIRED context in neither RELEASE_REQUIRED_CHECKS nor
RELEASE_INFORMATIONAL_CHECKS — the release publishes without waiting for it. …
##[error]update-golden-guard: required by ruleset #22113683 on `main` but absent from
branchProtectionContexts. …
Process completed with exit code 1.
```

Four things that log establishes, none of which was true yesterday:

1. **It runs.** No `BRANCH_PROTECTION_TOKEN`, no abort before the API call. Every scheduled run from
   2026-08-04 to 2026-09-01 died on `BRANCH_PROTECTION_TOKEN is unset`; this one reached the
   endpoint on the default workflow token.
2. **It reached the ruleset API specifically.** `ruleset #22113683` is not a value in the tree — it
   comes back in the live `rules/branches/main` response and is carried into the message so the
   remediation names the ruleset to edit.
3. **It detects a real drift, and names the offender.** Both directions fire on
   `update-golden-guard`, which #2938 is independently fixing. This is the first verdict this
   detector has ever produced, and it is correct: the context is required on `main` and is in
   neither `branchProtectionContexts` nor `RELEASE_REQUIRED_CHECKS` on `main` today.
4. **Direction B is clean**, which is only true because of the pagination fix in this PR. Before it,
   the same run also accused `forbid-deferral` — a healthy lane whose check-run sits on page 2 of a
   146-run commit — of being dead.

This PR deliberately leaves the pin alone, so the drift stays visible and is fixed by the PR that
caused it. Once #2938 lands, all three directions go clean.

Issue #2930, opened 2026-09-02
ci(mutation): every leg is credited a KILL for mutants that never compiled


tsouza:
Every mutation leg has been credited with a detection for mutants that never compiled. Measured, not
inferred: PR #2929 taught the gremlins fork to classify a build failure as `NOT VIABLE`, ran it, and
the kills each leg lost map **exactly** onto the mutants newly recorded `NOT VIABLE`.

## The mechanism

`go test` collapses a failing test, a package that does not build and a test that ran past its
`-timeout` into its own exit status **1**. Only the test *binary* exits 2, and gremlins spawns `go`.
So `internal/engine/executor.go`'s

```go
func getTestFailedStatus(exitCode int) mutator.Status {
	case 1: return mutator.Killed
	case 2: return mutator.NotViable   // "the mutant does not compile"
```

never reaches `case 2` through `go test`, and a mutant that does not compile is booked as a KILL —
the suite credited for a detection against source that a compiler rejected.

This is the "separate latent bug on the same function" noted on #2910.

## What is actually being mutated

gremlins' `INVERT_BITWISE` mutates `&` without distinguishing the binary bitwise-AND operator from
the **unary address-of** operator. On this codebase that means every `&chplan.Foo{...}` composite
literal becomes `|chplan.Foo{...}`:

```
internal/logql/detected_level.go:491:9: syntax error: unexpected |, expected expression
```

`internal/logql`'s plan-building code is almost entirely address-of composite literals, which is why
the count is so large there.

## Measured impact

Same legs, same corpus, main (`57383f96`, run 33605711543) against PR #2929 (run 33611413378):

| leg | main | with build failures classified |
| ------------------------- | ------------------------------------------------- | ------------------------------------------------------ |
| `phase4-logql-other-a` | killed 97, lived 4, timed out 0, not viable **0** — 96.04% PASS | killed 45, lived 4, timed out 0, not viable **52** — 91.84% FAIL |
| `phase4-traceql-ast` | killed 265, lived 12, timed out 8, not viable **0** — 92.98% | killed 208, lived 12, timed out 8, not viable **57** — 91.23% |
| `phase4-logql-lsyntax` | killed 449, lived 13, timed out 25, not viable **0** — 92.20% | killed 361, lived 13, timed out 25, not viable **88** — 90.48% |

`97 - 45 = 52`, `265 - 208 = 57`, `449 - 361 = 88`. Every lost kill is a mutant that never compiled.
[...]

Issue #2919, opened 2026-09-02
ci(mutation): a longer per-mutant budget lets a runaway mutant reap the runner on logql-other-b


tsouza:
`gremlins phase4-logql-other-b` now dies with no verdict at all: the runner
agent is reaped mid-run and the job reports

```
##[error]The runner has received a shutdown signal. This can happen when the
runner service is stopped, or a manually started runner is canceled.
```

That is the exact failure mode `.github/workflows/mutation.yml` already names
in the comment above the budget derivation — "a mutant that inverts a
scanner's loop-advance never terminates and allocates per iteration, so an
unbounded leash lets it exhaust the runner's memory. The OOM killer then reaps
the runner agent and the job dies with no verdict at all, which reads as flake
rather than as a missing bound."

## The causal chain, from two runs of the same leg

The mutant is `REMOVE_SELF_ASSIGNMENTS at internal/logql/logpattern/pattern.go:129:5`.

| run | measured cycle | derived budget | that mutant's verdict | leg outcome |
| --- | ---: | ---: | --- | --- |
| [33572151680](https://github.com/tsouza/cerberus/actions/runs/33572151680) (main, #2911's own) | 25.7s | **52s** | `TIMED OUT` | completed, 525 adjudicated |
| [33589651650](https://github.com/tsouza/cerberus/actions/runs/33589651650) attempt 1 (PR #2918) | 30.0s | **61s** | never reached | runner reaped at ~11.5 min |
| 33589651650 attempt 2 (a plain re-run of the same job) | 31.2s | **63s** | never reached | runner reaped at ~11.5 min |

Both PR attempts stop at exactly the same place — the four mutants on
`pattern.go:123:7` / `:124:5` / `:129:5` are adjudicated identically to the
baseline right up to `:129:5`, at which point gremlins prints
`Shutting down gracefully...`, marks every remaining mutant `NOT COVERED`, and
the runner disappears. At a 52s leash that mutant is killed by the leash
before it exhausts memory; at 61–63s it is not.

## Why this is #2911's derivation, not the mutant

The runaway mutant is not new — it timed out on `main` too, and #2913 records
the honest reading that a `TIMED OUT` mutant is unadjudicated. What changed is
the LEASH. Before #2911 the per-mutant bound was a flat 15s enforced through
the fork's `--timeout-max`, which is what kept an allocating non-terminating
mutant inside the runner's memory ceiling. #2911 replaced that with a budget
[...]

Issue #2917, opened 2026-09-02
ci(mutation): three legs still below the 95% floor — traceql-ast survivors, and two legs the timeouts cap


tsouza:
Spun out of #2913, which killed the survivors on four of the seven legs that
reported a real shortfall once #2911 stopped a timed-out mutant scoring as a
detection. Three legs remain below the 95% bar, and they split into two
different problems.

## 1. `phase4-traceql-ast` — a survivor-kill leg with no owner

Standing on the honest full-matrix run
[33572151680](https://github.com/tsouza/cerberus/actions/runs/33572151680):
265 killed, 12 lived, 8 timed out over a 285 denominator — 92.98%. Six kills
clear the bar (271/285 = 95.09%); killing all twelve reaches 97.19%.

The twelve survivors:

```
CONDITIONALS_BOUNDARY  lexer.go:517:15      sb.Len() > longestScopePrefix
CONDITIONALS_NEGATION  expr.go:56:33        BinaryOperation.impliedType, t != TypeAttribute
CONDITIONALS_NEGATION  expr.go:100:33       ScalarOperation.impliedType, t != TypeAttribute
CONDITIONALS_NEGATION  pipeline.go:190:10   Aggregate.impliedType, a.op == AggregateCount
CONDITIONALS_NEGATION  rewrite.go:164:8     staticTypeAllowed, t == a
INVERT_LOGICAL         lexer.go:543:23      tryScanDuration, r == scanner.EOF || IsSpace(r)
INVERT_LOGICAL         parse.go:191:17      ParseIdentifier, expr == nil || len(...) == 0
INVERT_LOGICAL         rewrite.go:115:52    foldArrayComparison's restrict clause
INVERT_LOGICAL         static.go:216:23     Static.Equals, s.Type == TypeNil || o.Type == TypeNil
INVERT_LOGICAL         validate.go:422:9    validateRegexPattern, !ok || Type != TypeString
INVERT_LOOPCTRL        parse.go:99:4        replaceIncompleteMatchers
INVERT_LOOPCTRL        rewrite.go:112:4     foldArrayComparison's attr-mismatch skip
```

Four are unambiguously killable by direct unit tests on pure functions — the
two `impliedType` arms, `Aggregate.impliedType`'s count case, and
`staticTypeAllowed` (which a `!=` mutation makes return true for the FIRST
non-matching entry, so `staticTypeAllowed(TypeInt, {TypeString, TypeStringArray})`
distinguishes it).

[...]

PR #2940, opened 2026-09-02
fix(ci): stop generating the mutants no compiler accepts, and stop vet judging the rest

### LINKED ISSUE #2933: ci(mutation): 234 mutants a leg still generates cannot compile, from three mechanisms with one root
Found while measuring #2930. The fix there removes the dominant source of mutants that cannot
compile — `INVERT_BITWISE` applied to a prefix `&` or `^` — but it removes one shape of a wider
problem, and the rest is measurable.

## Measured

Every `NOT VIABLE` record across all 30 legs of the honest full-matrix run
[33611413378](https://github.com/tsouza/cerberus/actions/runs/33611413378), read out of the uploaded
`gremlins.json` reports:

| mutator | count | share |
| ------------------- | ----: | ----: |
| `INVERT_BITWISE` | 3511 | 93.8% |
| `ARITHMETIC_BASE` | 150 | 4.0% |
| `INVERT_LOGICAL` | 78 | 2.1% |
| `INVERT_ASSIGNMENTS` | 5 | 0.1% |
| `INVERT_LOOPCTRL` | 1 | 0.0% |
| **total** | **3745** | |

\#2930 removes the 3511. The remaining 234 stay, and they are not one accident — they are three
mechanisms with one root.

## The root

gremlins rewrites a token in the AST and writes the file back. It never asks whether the result is
legal Go in that context: it has no type information, and it does not model what `go test` itself
rejects. #2930's prefix-versus-infix case was the positional instance of that. These are the rest.

**1. The operand's type forbids the mutated operator** — 155 records
(`ARITHMETIC_BASE` + `INVERT_ASSIGNMENTS`). `+` on strings is concatenation, and `-` is not defined
on them at all. `internal/traceql/ast/pipeline.go:45` is representative:

```go
parts[i] = hint.Name + "=" + hint.Value.String()   // ARITHMETIC_BASE ADD -> SUB
// invalid operation: operator - not defined on name (variable of type string)
```

The same root also reaches constant arithmetic. `internal/chsql/ddl.go:90` holds
`int64(d / (24 * time.Hour))`; `MUL -> QUO` makes the divisor `24 / time.Hour`, which is the untyped
constant `0`:

```
invalid operation: division by zero
```

**2. `go test`'s own vet pass rejects the mutant** — 78 records (`INVERT_LOGICAL`). These COMPILE.
`go test` runs a subset of `go vet` before it runs anything, and the `bools` analyzer reports a
conjunction of equalities against distinct constants. `internal/logql/detected_level.go:89` is the
one that survives #2930 on `phase4-logql-other-a`:

```go
return name == detectedLevelLabel || name == levelLabelAlias   // INVERT_LOGICAL LOR -> LAND
```

```
./p.go:5:44: suspect and: name == a && name == b
FAIL    pkg [build failed]
```

`go build` is silent on it, so a viability check that shells out to the compiler alone would miss
this one; the check has to be whatever `go test` will actually accept.

**3. The statement is not legal where it sits** — 1 record (`INVERT_LOOPCTRL`). `break` inside a
`switch` becomes `continue`:

```
continue is not in a loop
```

Each of the three was reproduced against the real toolchain rather than read off the report.

## What it costs, honestly

Not a single efficacy point. Since #2929 a `NOT VIABLE` mutant leaves both sides of
`killed / (killed + lived + timedOut)`, so these 234 neither credit nor penalise any leg. This issue
is not about a wrong number.

What it costs is the meaning of a leg's mutant set, which is the same argument #2930 makes at 15x
the scale: a package whose report lists 234 entries no compiler accepts is describing the compiler,
not its tests. It also costs runner time — each one is a full recompile-and-link cycle spent
producing a verdict that says nothing.

## Where a fix goes

`tsouza/gremlins`, alongside #2930's `MutantTypesFor`. The fork already has the gate that would
catch a regression: `TestEveryGeneratedMutantCompiles` (added by #2930) builds a package exercising
every operator the table maps, applies each generated mutant and hands it to the real compiler. It
passes today only because its fixture holds no string concatenation, no constant divisor and no
`break` inside a `switch`; widening that fixture is what turns this issue red, and is the first step.

The mechanism itself is a design choice with a real trade-off, and it should be taken deliberately
rather than assumed:

- **Type-check before emitting.** `go/types` on the mutated file answers cases 1 and 3 exactly. It
  costs a type-check per candidate mutant, on a tool whose whole cost model is already dominated by
  recompiles.
- **Restrict by operand shape without types.** `ADD` on an operand that is a string literal, or on a
  `+` chain containing one, covers most of case 1 for nothing. It is a heuristic and will not be
  complete.
- **Case 2 is not a compile question at all.** It needs the mutant run through `go vet`'s test
  subset, or the `bools` analyzer's rule reimplemented at the mutation site.

## Acceptance

- `TestEveryGeneratedMutantCompiles`'s fixture covers all three shapes above, and passes.
- A full-matrix run reports `NOT VIABLE` at or near zero across all 30 legs, with any residue named
  and explained rather than tolerated.
- No leg's efficacy floor moves, and no file is excluded — the floors are unaffected by this, which
  is exactly why it must not be used to justify touching one.

tsouza:
The tail of #2891's three-part cleanup. #2903/#2911 stopped scoring a starved runner as a
detection, #2919/#2921 gave a memory-reaped runner a verdict, #2930/#2932 stopped crediting
uncompilable mutants as kills. This is the 234 that #2932 left — the control in its own
measurement, unchanged before and after, so a bounded set rather than an estimate.

## What the 234 were

| mutator | records | mechanism |
| --- | ---: | --- |
| `ARITHMETIC_BASE` | 150 | the operand's type, or a constant's value, forbids the mutated operator |
| `INVERT_ASSIGNMENTS` | 5 | the same, through `s += x` on a string |
| `INVERT_LOOPCTRL` | 1 | the statement is not legal where it sits |
| `INVERT_LOGICAL` | 78 | `go test`'s own vet pass rejects it — **not** a compile question |

One root under the first three: gremlins' table says what a rewrite *means* for a token read in
isolation, which is all a table can say. Whether the result is a program depends on the operand
types, on the constant values around it, and on what statements the enclosing function admits —
none of which is in the token. #2930's prefix-versus-infix case was the positional instance of the
same thing.

## Mechanism 1 and 3: type-check the candidate's whole package before emitting it

Both are decided exactly by `go/types`, and the same check decides both, so there is one mechanism
rather than two rules. Reproduced against the real toolchain, each of them:

```
./zoo.go:30:41: invalid operation: operator - not defined on a (variable of type string)
./zoo.go:39:44: invalid operation: division by zero
./zoo.go:54:1:  missing return
./zoo.go:61:3:  continue is not in a loop
```

The unit is the **package**, not the expression, and that is not incidental. `const week = 7 * 24 *
time.Hour` mutated `MUL -> QUO` is a perfectly legal constant expression on its own line; it is the
`d / week` three lines down that becomes a division by zero. Anything narrower cannot see it. The
issue offered "restrict by operand shape without types" as the cheap alternative, and the corpus
rules it out on its own evidence: `internal/traceql/ast/attribute.go:96` is `return prefix + name`
and `internal/promql/histogram_native_range_fn.go:613` is `return hqWindowAllArrayAliasPrefix +
alias`. No string literal anywhere in either `+` chain. A shape heuristic misses both, and it would
have missed the constant case entirely.

**Cost.** One type-check per candidate mutant, measured at ~100ms for `internal/promql` (101 files,
93k lines) against a ~10s per-mutant recompile-link-run. It *removes* cycles rather than adding
them, and generation runs on its own goroutine behind the executor pool, so it is hidden entirely
behind execution. Package load is ~3.5s per package directory, once. Peak RSS of the loaded graph
for `internal/promql` is 366 MB.

**Degradation.** A package that cannot be loaded and type-checked *as it stands* is not used as an
oracle at all: its mutants are generated and left to the compiler exactly as before, and a log line
names the package. That falls out of a single baseline check rather than an enumeration of failure
modes — wrong build tags, an unresolved dependency, a language version `go/types` cannot express all
land there. The direction matters: silently dropping a mutant nobody proved illegal would shrink the
set a score is measured against.

One bug worth naming, because it made the first version of this a silent no-op: gremlins'
`GoModule.Root` is `"."` whenever the scope is given as a relative path, which is how every
documented invocation gives it, while `go list` reports absolute paths. The index never matched and
the checker suppressed nothing while looking like it worked. It is fixed, and a file the checker
cannot answer for now says so out loud rather than answering "viable".

## Mechanism 2: `-vet=off`, not a refusal to generate

This one is argued rather than assumed, because it is the one place where "do not generate it" is
the *wrong* answer.

`go test` runs a subset of `go vet` before it builds anything and reports a finding as
`FAIL pkg [build failed]` — from the outside, indistinguishable from source that does not compile.
Its `bools` analyzer rejects exactly what `INVERT_LOGICAL` produces from `name == a || name == b`: a
conjunction of equalities against two distinct **constants**, "suspect and". (It ignores the same
shape over variables, whose values it cannot compare — worth knowing, because a fixture built from
parameters does not reproduce the bug.)

Every one of those 78 is legal Go and a real change of behaviour. `go build` accepts them without a
word; the mutated predicate is unsatisfiable where the original was not; a suite exercising either
operand kills it. Refusing to generate them would drop from the corpus mutants the suite *ought to
be asked about* — the one direction a mutation tool must never shrink in, and the same argument this
repo makes against tolerance files. What is wrong is not the mutant; it is that a style analyzer
gets to decide whether a mutant may be adjudicated at all. So the analyzer leaves the loop. It also
takes a vet pass off every mutant's build, on a tool whose cost is dominated by builds.

**This is the one half of this change that can move a leg's number.** 78 mutants that used to leave
both sides of `killed / (killed + lived + timedOut)` now get a real verdict. If all are killed a leg
rises slightly; if one lives, that is a gap in the suite that was never being asked about, and it
gets fixed at the test rather than absorbed or floored away. The full matrix on this PR is what
measures it.

## The gates, and that they can fail

Four tests in the fork, all behavioural:

- **`TestEveryGeneratedMutantCompiles`** — builds a package exercising every operator the table maps
  plus every shape above, applies each generated mutant, hands it to the real compiler. The fixture
  is widened, and the engine now runs over a real directory rather than an `fstest.MapFS`, because a
  map in memory is not something a compiler can be asked about.
- **`TestTheViabilityCheckIsWhatKeepsTheZooCompiling`** — the break-proof. Identical generation with
  the check removed, and the compiler must reject **exactly** the sites the fixture marks
  `// illegal:`. Neither a fixture that quietly stopped exercising a shape (missing rejection) nor a
  check that suppresses more than the compiler would (extra rejection) leaves the first test
  passing. Verified in both directions by deliberately mis-marking the fixture.
- **`TestViabilityKeepsTheLegalMutationsAtTheSameSites`** — the over-suppression guard, since
  removing mutants is precisely how a score gets inflated. Every mutation marked `// legal:` at
  those same sites must survive.
- **`TestMutantsAreRunWithVetDisabled`** — runs the suspect conjunction through `go test` twice,
  with the flag and without, and requires the second to fail on `suspect and`. Both halves, because
  only the pair is evidence: one shows the mutant now builds and runs, the other shows it is the
  flag that makes it. Verified to fail when `-vet=off` is removed from `getTestArgs`.

Break-proof for the compile gate, with the check sabotaged — all four diagnostics, at their sources:

```
--- FAIL: TestEveryGeneratedMutantCompiles
    ./zoo.go:30:41: invalid operation: operator - not defined on a (variable of type string)
    ./zoo.go:30:41: invalid operation: operator - not defined on a + "-" (value of type string)
    ./zoo.go:31:37: invalid operation: operator - not defined on a (variable of type string)
    ./zoo.go:39:44: invalid operation: division by zero
    ./zoo.go:54:1:  missing return
    ./zoo.go:61:3:  continue is not in a loop
```

## Before / after on this tree

`--dry-run` with the shipped binary against one built with the check disabled, over every scope the
lane's 30 legs use. The unchecked column is the control:

| scope | candidates before | after | removed |
| --- | ---: | ---: | ---: |
| `internal/chplan` | 1034 | 1022 | 12 |
| `internal/chsql` | 1687 | 1607 | 80 |
| `internal/logql` | 1152 | 1117 | 35 |
| `internal/optimizer` | 119 | 110 | 9 |
| `internal/promql` | 1948 | 1913 | 35 |
| `internal/qlcommon` | 328 | 319 | 9 |
| `internal/spansscan` | 112 | 107 | 5 |
| `internal/traceql` | 688 | 632 | 56 |
| **total** | **7068** | **6827** | **241** |

**Added: 0, everywhere.** By mutator: 230 `ARITHMETIC_BASE`, 6 `INVERT_ASSIGNMENTS`, 4
`INVERT_LOOPCTRL`, 1 `INVERT_BITWISE`. Every removal was read back against its source line and is
string concatenation, a constant that becomes zero and is then divided by, or a loop-control
statement moved somewhere it is not legal.

241 exceeds the issue's 234 because a dry run enumerates every candidate including the `NOT COVERED`
ones, which never reached a verdict there was a `NOT VIABLE` to report; and because the 78
`INVERT_LOGICAL` are not in this column at all — they are still generated, and are now run.

Two things the corpus said that the issue did not predict:

- **The single `INVERT_LOOPCTRL` record is the mirror image of the one #2933 describes.** The issue
  predicted `break` inside a `switch` becoming `continue is not in a loop`. The actual record,
  `internal/logql/jsonpath.go:168`, is a `continue` — and it is what keeps `next()`'s `for {}` from
  terminating, so `break` leaves the function `missing return`. Both shapes are in the fixture; only
  the second was ever generated here. The other three the sweep found are the predicted shape, in
  `case` arms of type switches that no loop encloses.
- **A fifth shape nobody predicted.** `internal/optimizer/constant_fold.go:367` is
  `func foldNumeric[T int64 | float64]`, and the `|` of a type-parameter constraint is an
  `*ast.BinaryExpr` like any other, so `INVERT_BITWISE` still reached it after #2930:
  `int64 & float64 is not a type`. Nothing about it resembles the four shapes that were catalogued.
  That is the argument for a type-check over a rule per shape, made by the corpus rather than by me.

## Fork

New tags on both branches, fetched and installed from the module proxy before this PR was opened:

- `v0.6.0-cerberus-viable-mutants` (`cerberus-sigterm-fix`, upstream module path, reviewable diff)
- `v0.6.0-cerberus-viable-mutants-consume` (`cerberus-sigterm-fix-consume`, the one the lane
  installs)

The two branches differ only by module path — verified by normalising the import prefix out of the
inter-branch diff and confirming every remaining line pairs. `test/regression/mutation_timeout_max_test.go`
pins the tag whole, with the family history extended; `docs/toolchain.md`'s fork section gains both
fixes. The fork's own suite is green and `golangci-lint` is at its pre-change baseline of 39
findings, none in the new files.

## Not verified here

- **Efficacy movement.** Nothing local can measure it; the full matrix runs on this PR because
  `.github/workflows/mutation.yml` is a `HARNESS_PATHS` entry. Any leg that moves gets explained
  against these numbers, not absorbed, and no floor moves either way.
- **The dry runs above ran against this branch's tree**, which has drifted from the SHA the issue's
  artifacts were read from (`2c347cf`). Per-site correspondence was checked by source text rather
  than by line number, and every site the issue named is in the removed set.

Closes #2933

### DIFF
--- a/.github/workflows/mutation.yml
+++ b/.github/workflows/mutation.yml
@@ -164,7 +164,7 @@ jobs:
       # opaque `gremlins exited N` with no artifact and no measured-vs-threshold
       # numbers.
       #
-      # We consume gremlins via the tsouza/gremlins fork, which carries three
+      # We consume gremlins via the tsouza/gremlins fork, which carries five
       # fixes cerberus needs:
       #
       #   --on-shutdown-status=not-run   in-flight mutants killed by the
@@ -192,6 +192,37 @@ jobs:
       #     the compiler did: `Not viable: 0` on every leg was the tell that the
       #     status had never once been reported. The fork no longer generates
       #     them (#2930).
+      #
+      #   a candidate mutant is type-checked before it is emitted
+      #     The line above is one instance of a wider gap: a token table says
+      #     what a rewrite MEANS for a token read on its own, and whether the
+      #     result is a program depends on the operand types, the constant
+      #     values around it, and what statements the enclosing function
+      #     admits. `a + "-" + b` has no `-` on this tree either; `const week
+      #     = 7 * 24 * time.Hour` mutates into a legal constant expression
+      #     whose `d / week` further down is a division by zero; and
+      #     INVERT_LOOPCTRL turns the `continue` that keeps a `for {}` from
+      #     terminating into a `break` that leaves the function missing a
+      #     return. The fork type-checks the candidate's whole PACKAGE — the
+      #     package, because the error need not appear where the mutation is —
+      #     and does not emit it if that fails (#2933). A package it cannot
+      #     load and type-check as it stands is not used as an oracle at all:
+      #     its mutants are generated and left to the compiler, as before.
+      #
+      #   the mutated tests run with `-vet=off`
+      #     `go test` runs a subset of `go vet` before it builds anything and
+      #     reports a finding as `FAIL pkg [build failed]`, which from the
+      #     outside is source that does not compile. Its bools analyzer rejects
+      #     exactly what INVERT_LOGICAL produces from `name == a || name == b`
+      #     — a conjunction of equalities against distinct constants, "suspect
+      #     and". Those mutants are legal Go and a real change of behaviour, so
+      #     the fix is not to stop generating them (that would drop mutants the
+      #     suite ought to be asked about) but to stop letting a style analyzer
+      #     decide whether a mutant may be adjudicated. This is the one entry
+      #     in this list that can MOVE a leg's number rather than only correct
+      #     its meaning: 78 mutants that used to leave the ratio as NOT VIABLE
+      #     now get a real verdict, and one that nothing kills is a genuine gap
+      #     in the suite (#2933).
       # Routed through go-module-fetch.mjs because gremlins is NOT in go.mod, so
       # the warm step in ./.github/actions/setup-go — which runs
       # `go mod download` over the main module's own requirements — does not
@@ -204,7 +235,7 @@ jobs:
       - name: install gremlins
         run: >
           node .github/scripts/go-module-fetch.mjs --
-          go install github.com/tsouza/gremlins/cmd/gremlins@v0.6.0-cerberus-unary-operators-consume
+          go install github.com/tsouza/gremlins/cmd/gremlins@v0.6.0-cerberus-viable-mutants-consume
       # `workers: 0` keeps gremlins' default (runtime.NumCPU()); a
       # non-zero value caps the parallel `go test` fan-out per the
       # comment above the matrix entry.
--- a/docs/toolchain.md
+++ b/docs/toolchain.md
@@ -49,7 +49,7 @@ because that rule has no auto-fixer of its own.
 `mutation.yml` installs the `tsouza/gremlins` fork rather than upstream `go-gremlins/gremlins@v0.6.0`:
 
 ```bash
-go install github.com/tsouza/gremlins/cmd/gremlins@v0.6.0-cerberus-unary-operators-consume
+go install github.com/tsouza/gremlins/cmd/gremlins@v0.6.0-cerberus-viable-mutants-consume
 ```
 
 The fixes it carries defend one thing between them: that the number a run reports is a number the
@@ -119,13 +119,44 @@ leg's honest score is identical either way, since `NOT VIABLE` leaves both sides
 set padded with entries no compiler accepts measures nothing. The `NOT VIABLE` classification stays
 for a genuine build failure from any other source.
 
+**A candidate mutant is type-checked before it is emitted.** Reading a prefix operator as a prefix
+operator is one instance of a wider gap: the mutation table describes what a rewrite *means* for a
+token read on its own, and whether the result is a program depends on the operand types, on the
+constant values around it, and on what statements the enclosing function admits — none of which is
+in the token. Three shapes of that survived on this tree. `hint.Name + "=" + hint.Value.String()`
+became `operator - not defined on ... (variable of type string)`, since Go defines `+` on strings
+and nothing else; `const week = 7 * 24 * time.Hour` became a legal constant expression whose `d /
+week` three lines down is a division by zero; and `INVERT_LOOPCTRL` turned the `continue` that keeps
+a `for {}` from terminating into a `break` (`missing return`), and a `break` inside a `switch` no
+loop encloses into a `continue` that is not in a loop.
+
+The fork type-checks the candidate's whole **package** before emitting it — the package, because the
+error a mutation causes need not appear where the mutation is, as the constant case shows. One
+type-check (~100ms on `internal/promql`) buys back a whole recompile-link-run cycle (~10s) whenever
+it rejects, and generation runs on its own goroutine behind the executor pool, so the lane pays
+nothing for it. A package that cannot be loaded and type-checked as it stands is not used as an
+oracle at all: its mutants are generated and left to the compiler exactly as before, and a log line
+names the package. Dropping a mutant nobody proved illegal would shrink the set a score is measured
+against, which is the one direction a mutation tool must not move in.
+
+**`go vet` does not decide whether a mutant may be adjudicated.** `go test` runs a subset of `go
+vet` before it builds anything and reports a finding as `FAIL pkg [build failed]` — from the outside
+indistinguishable from source that does not compile. Its `bools` analyzer rejects exactly what
+`INVERT_LOGICAL` produces from `name == a || name == b`: a conjunction of equalities against
+distinct constants, which it calls "suspect and". Those mutants are legal Go and a real change of
+behaviour — the predicate becomes unsatisfiable, and any test exercising either operand kills it —
+so the fork keeps generating them and runs the mutated tests with `-vet=off`. This is the one fix in
+this list that can MOVE a leg's number rather than only correct its meaning: mutants that used to
+leave the ratio as `NOT VIABLE` now get a real verdict, and one that nothing kills is a genuine gap
+in the suite rather than an artefact.
+
 `.gremlins.yaml`'s `exclude-files` paths are interpreted relative to the run's scope, not the repo
 root, and the matcher is RE2 with no lookahead. A path in the wrong form silently excludes nothing.
 
 The fork ships two branches on purpose. `cerberus-sigterm-fix` at tag
-`v0.6.0-cerberus-unary-operators` is the branch the upstream pull request is built from, and keeps
+`v0.6.0-cerberus-viable-mutants` is the branch the upstream pull request is built from, and keeps
 the upstream module path `github.com/go-gremlins/gremlins` so the diff stays reviewable.
-`cerberus-sigterm-fix-consume` at tag `v0.6.0-cerberus-unary-operators-consume` is the branch
+`cerberus-sigterm-fix-consume` at tag `v0.6.0-cerberus-viable-mutants-consume` is the branch
 `mutation.yml` installs; it adds one commit renaming the `go.mod` module path to
 `github.com/tsouza/gremlins` and rewriting the internal imports, because `go install` otherwise
 rejects the module with `module declares its path as: github.com/go-gremlins/gremlins`. The fixes
--- a/test/regression/mutation_timeout_max_test.go
+++ b/test/regression/mutation_timeout_max_test.go
@@ -67,7 +67,18 @@ const (
 //     the exit-status collapse above then booked each one KILLED. `Not viable:
 //     0` on every leg was the tell: the status had never once been reported,
 //     and three legs measured 52, 57 and 88 lost kills apiece once it was.
-const gremlinsForkTag = "github.com/tsouza/gremlins/cmd/gremlins@v0.6.0-cerberus-unary-operators-consume"
+//   - the `viable-mutants` family (#2933): type-checks a candidate mutant's
+//     whole package before emitting it, closing the 156 remaining mutants no
+//     compiler accepts — string concatenation mutated to subtraction, a
+//     constant mutated to zero and then divided by, a loop-control statement
+//     moved somewhere it is not legal. It also runs the mutated tests with
+//     `-vet=off`, which is the only entry in this list that MOVES a leg's
+//     number rather than correcting its meaning: 78 INVERT_LOGICAL mutants
+//     that `go vet` used to reject as "suspect and" — legal Go, and a real
+//     change of behaviour — now reach a test binary and are adjudicated.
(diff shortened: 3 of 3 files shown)


tsouza:
## CI measurement — first 10 legs

Control: main's own full-matrix run [33635018460](https://github.com/tsouza/cerberus/actions/runs/33635018460)
(`5af3f78d5`, the `unary-operators` tag). It still reports **exactly 234 `NOT VIABLE` across 7067
mutants** — the same 234 #2933 counted, reproduced independently on today's main, which is what makes
this a control rather than a re-measurement. 15 of its 30 legs are already below the 95% floor;
those reds are pre-existing and tracked in #2917.

After: this PR's run [33643658437](https://github.com/tsouza/cerberus/actions/runs/33643658437),
same tree apart from two CI-only commits, `viable-mutants` tag.

| leg | NOT VIABLE | total | killed | lived | efficacy |
| --- | --- | --- | --- | --- | --- |
| `phase3-optimizer` | 9 → **0** | 119 → 110 | 107 → 108 | 1 → 1 | 98.17 → 98.18 |
| `phase4-logql-aggregation` | 1 → **0** | 103 → 103 | 88 → 89 | 7 → 7 | 92.63 → 92.71 |
| `phase4-logql-other-a` | 1 → **0** | 51 → 51 | 45 → 46 | 4 → 4 | 91.84 → 92.00 |
| `phase4-logql-parser` | 1 → **0** | 80 → 80 | 68 → 69 | 0 → 0 | 95.77 → 95.83 |
| `phase4-promql-d` | 3 → **0** | 107 → 105 | 86 → 87 | 8 → 8 | 91.49 → 91.58 |
| `phase4-promql-e` | 5 → **0** | 125 → 121 | 101 → 103 | 5 → 5 | 95.28 → 95.37 |
| `phase4-promql-i` | 8 → **0** | 157 → 153 | 134 → 138 | 12 → 12 | 91.78 → 92.00 |
| `phase4-promql-other` | 5 → **0** | 116 → 116 | 92 → 95 | 12 → **14** | 88.46 → **87.16** |
| `phase4-traceql-lower` | 8 → **0** | 175 → 175 | 137 → 145 | 8 → 8 | 94.48 → 94.77 |
| `phase6-spansscan` | 5 → **0** | 112 → 107 | 102 → 102 | 4 → 4 | 96.23 → 96.23 |

`NOT VIABLE` is **0 on every leg measured so far**. The two halves separate cleanly and visibly:
where the total drops, the type-check removed candidates (`optimizer` −9, `spansscan` −5,
`promql-e` −4, `promql-i` −4, `promql-d` −2 — `optimizer` and `spansscan` match the local dry-run
counts exactly); where the total is unchanged, the residue was vet-only and those mutants are now
adjudicated instead of withheld.

No leg went from green to red. Nine of the ten moved up or stood still.

### The one leg that moved down, and why

`phase4-promql-other`: 88.46% → 87.16%. Its 5 `NOT VIABLE` were all vet-blocked `INVERT_LOGICAL`
(total unchanged at 116), and adjudicating them killed three and found **two real survivors**:

```
NOT VIABLE -> KILLED  INVERT_LOGICAL histogram_native_scalar_binop.go:77:35
NOT VIABLE -> KILLED  INVERT_LOGICAL histogram_native_set_op.go:208:36
NOT VIABLE -> KILLED  INVERT_LOGICAL histogram_native_ts_of_first_last_over_time.go:85:58
NOT VIABLE -> LIVED   INVERT_LOGICAL histogram_native_mixed_or_aggregate_topk.go:73:35
NOT VIABLE -> LIVED   INVERT_LOGICAL histogram_native_scalar_binop.go:240:26
```

Both survivors are honest gaps, not artefacts:

- `topKOverMixedExpHistogramSetOp`'s guard
  `!ok || (agg.Op != parser.TOPK && agg.Op != parser.BOTTOMK)` inverted to `||` makes the second
  clause a tautology, so the recognizer always returns false and the topk-over-mixed-set-op lowering
  is never entered. Nothing in the suite notices.
- `isExpHistogramValuedShape`'s `(b.Op == parser.MUL || b.Op == parser.DIV) && …` inverted to `&&`
  makes it always false, so `<exp-hist> * scalar` and `<exp-hist> / scalar` stop being recognised
  as exp-histogram-valued. The arm above still covers `scalar * <exp-hist>`, which is why nothing
  fails.

This is the fix working as designed rather than a regression: `killed / (killed + lived + timedOut)`
is being asked about two mutants it was previously allowed to skip, and it cannot answer. The leg was
already at 88.46% — far under its floor and on #2917's list — so nothing flips here; what changes is
that two more of its gaps are now named. They are routed to #2917 with the rest of that class rather
than absorbed, and no floor moves.

The `distinct_sample_rows.go:147 -> 171` line shift in the raw diff is #2934, which landed between
the control's SHA and this branch's base. It is not part of this change.

Remaining legs are still queued behind main's concurrent full-matrix run; this comment will be
completed when they land.

PR #2998, opened 2026-09-03
test(qlcommon): reach the two capture-probe declines a real regex can produce

### LINKED ISSUE #2995: test(qlcommon): the coverage floor cannot be met without reaching five capture-probe declines
## The shortfall is exactly the size of the guards #2977 added

`internal/qlcommon` measures **98.17% (535/545 statements)** against a committed floor of
**99%**. It is not in the seven packages tsouza/cerberus#2991 quotes from run
[33692780097](https://github.com/tsouza/cerberus/actions/runs/33692780097) because it fell
after that run: the cause merged on 2026-09-03 in #2977, *"make a scanner that stops advancing
an error, not a memory runaway"*.

#2977 planted a progress invariant at the top of four scanner loops. Each is a guard of the
same shape:

```go
start := j
switch src[j] {
...
}
// Progress invariant — see scanSourceGroups. A class scan that
// stops advancing runs away on the same untrusted pattern.
if j <= start {
    return 0, false
}
```

Those guard bodies are **five statements**, and all five are unreachable by construction —
every arm of every one of those switches either advances the cursor by at least one byte or
returns:

- `regex_source_scan.go`'s `skipCharClass` — arms are `j += 2`, `j += 2 + end + 2` (with
  `end >= 0`), `j++`, or `return`.
- `regex_source_scan.go`'s `scanSourceGroups` — same shape.
- `label_replace.go`'s `resolveSegments` — `emptyCapturesStep`'s sibling `replacementStep`
  documents and returns `>= 1` unconditionally.
- `label_replace.go`'s `EmptyCapturesReplacement` — two statements, guarded on
  `emptyCapturesStep`, whose doc comment states it "returns the number of input bytes it
  consumed (always >= 1)".

This is deliberate defence-in-depth against the class #2977 was fixing, and it should stay.
But it is dead weight in the ratio: 545 statements now need 540 covered to clear 99%, and the
package has 535.

## What restoring the floor actually requires

The other five uncovered statements are genuine rejection arms in `capture_probe.go`, all
reachable in principle:

| construct | uncovered arm |
| --- | --- |
| `capture_probe.go`'s `planCaptureProbes` | the `insertProbes` decline |
| `capture_probe.go`'s `planCaptureProbes` | the `readBackPlan` decline |
| `capture_probe.go`'s `planCaptureProbes` | the `stripAnchors` decline |
| `capture_probe.go`'s `negativeSiblingSpans` | the `!branchShape.unconditional` decline |
| `capture_probe.go`'s `probeSpanFor` | the `syntax.Parse` failure on a sub-slice span |

Covering **all five** lands the package at 540/545 = **99.08%** — above the floor by 0.08
percentage points, with no headroom for the next defensive statement anyone adds. Covering
four of the five does not clear it at all.

Each needs an input that drives the capture-probe planner into a specific decline: a regex
whose candidate span slices into a fragment that does not parse on its own, or one whose
carrier capture is conditional within its alternation branch. Reaching them by construction
rather than by accident is the work; a test that merely lands on the line would be coverage
theatre and is explicitly not what this asks for.

## Not in scope for #2991's restoration PR

That PR restores the six packages whose drops were plain missing tests for merged production
code. This one is a different shape — the code is right, the arms are real but need genuine
regex-planner reasoning to reach honestly, and the margin afterwards is 0.08pp. It is filed
rather than rushed.

Note the floor is not the thing to move (CLAUDE.md invariant 7). If the conclusion after
attempting the five arms is that the ratio itself is the problem, that is a finding to bring
back with the arithmetic, not a ledger edit.

tsouza:
`internal/qlcommon` was the last package below its committed coverage floor —
**98.17% against 99%** — and the last thing keeping `main`'s `coverage` lane red.

## The issue's arithmetic was impossible

#2995 assumed five coverable capture-probe declines. There are **two**. The floor needs 540
statements; the package had 537, and **539 even under the most generous forcing**. The ceiling sat
below the floor, so no amount of test-writing could have reached it.

**Six of the ten uncovered statements are unreachable by construction**, not five. Five are #2977's
progress guards. The sixth is `planCaptureProbes`'s `insertProbes` decline: the spans it assembles
are ancestor group bodies, alternation branches and sibling branches of one source — a **laminar
family** — and `insertProbes` declines only on two spans that overlap *without* nesting. Same
category, arrived at independently.

Search effort behind "unreachable": token-level enumeration, character-level random sweeps over
**~50M compiling patterns** (each tried with every single carrier and the full carrier set), then a
22.5-minute coverage-guided `go test -fuzz` campaign over an 882-entry corpus. Three arms never
fired.

## Two real declines, reached through the public path

The reaching input class is a regex whose own `)` closes the `^(?s:` wrapper early. Prometheus
anchors the same way, so `a)|b` is a legitimate user pattern; anchored it is `^(?s:a)|b$`, an
alternation whose second arm sits *beside* the wrapper.

- `a)|(?P<d>b?)(?P<d>c` — the probe swallows the closing wrapper → **stripAnchors** decline.
- `a)|(?P<d>b?)(?P<d>c)\Q` — the unterminated quoted run swallows the probe's own `)` → the rewrite
  does not compile → **readBackPlan** decline.

Both reach it through `ReplacementToCH` with `carriersNeedingProbes()` supplying the carriers, not a
synthetic `need`. The stripAnchors arm is mutation-killed. **The readBackPlan arm is dominated** —
removing its early return leaves the plan at its zero value, whose empty regex the wrapper check
refuses instead — so the domination is recorded in the test comment rather than claimed as a kill.

## The guards: reshaped, not deleted

The standing decision is *delete unreachable code, do not accept the red*. But these five guards were
deliberate defences whose accepted justification was that a scanner which stops advancing is an
unbounded allocation driven by untrusted input — they earn their place *because* they never fire.

The third answer is #2977's own precedent. `lsyntax.NormalizeDottedLabels` is a **single-advance
walker**: `next := i + 1` before the dispatch, arms overwrite it or not, one `i = next` at the
bottom. The four scanners now use that form.

This does not weaken the safety argument — it makes the same property **unfalsifiable instead of
asserted**. A guard catches a non-advancing arm after the fact; this shape makes it unrepresentable.
The `|`, `)` and class-member arms now write no offset at all, which *is* the "forgot to advance"
arm, and `TestArmsThatComputeNoOffsetStillConsumeAByte` asserts their exact offsets against the
production scanners including a 4096-byte run. What the shape does not decide — the three arms that
*do* write `next`, each taking its offset from a helper's own contract — is pinned separately by
`TestScannerHelpersAlwaysConsumeAByte` (1299 calls) and `TestReplacementStepsAlwaysConsumeAByte`
(7352 steps).

`errTemplateStalled` goes with the guard that was its only reader, and four `NOT KILLABLE` verdicts
adjudicating those guards are retired with them (equivalence verdicts 68 → 64, exactly the four).

## Evidence

- **Byte-identical behaviour**: differential against verbatim replicas of the pre-reshape loop bodies
  over **8,000,000 random patterns** — 8M scans, 2,400,497 class skips, 8M template renders — zero
  divergence, declines included. No golden regenerated, no fixture churned.
- 15 downstream packages green under `-race`, including `test/spec` and `test/regression`.
- **Coverage 98.17% → 99.4%**, above the floor with margin rather than the 0.08 points the original
  arithmetic assumed.

## Perf, reported honestly

The local A/B measurement was run on a host shared with other work and **came back inconclusive**, so
no figure is quoted here rather than rounding an ambiguous result into a confident one.
`benchstat diff (baseline vs ref)` on clean CI hardware is the authority for this repository, and the
new `scan_bench_test.go` gives it the four reshaped loops to measure — the same treatment #2977 gave
the lexer loop when it added the invariants this removes.

If that gate flags a regression, the hypothesis to check first is inlining: #2977 found an earlier,
*clearer* draft cost 1.8–4.0% purely because an extracted call would not inline, and this change has
the same shape and the same hazard.

Closes #2995

### DIFF
--- a/internal/qlcommon/capture_probe_boundary_test.go
+++ b/internal/qlcommon/capture_probe_boundary_test.go
@@ -770,3 +770,173 @@ func TestCarriersNeedingProbesIgnoresUnsharedNames(t *testing.T) {
 // index of a '(', and `bar` is the byte index of a '|'. One byte cannot be
 // both, so offset == bar is unreachable and the two comparisons agree on
 // every offset a caller can supply.
+
+// anchorBreakoutWitnesses are the regexes that reach the two arms
+// [planCaptureProbes] applies to its OWN rewrite after the scanner has
+// picked a span — the read-back against the compiler, and the removal of
+// the anchoring wrapper. Both arms are about the same event, which no
+// other test in this package produces: a probe boundary landing at an END
+// of the anchored pattern rather than strictly inside it.
+//
+// The event needs a regex whose own `)` closes the `^(?s:` wrapper early.
+// Reference Prometheus anchors the same way, so `a)|b` is a pattern a
+// user may legitimately write and cerberus must accept: after anchoring
+// it is `^(?s:a)|b$`, a top-level alternation whose second arm sits
+// BESIDE the wrapper instead of inside it. A carrier in that second arm
+// has the whole arm — the wrapper's closing `)$` included — as its
+// nearest probeable span, so the probe cerberus would insert brackets
+// text the wrapper was supposed to bracket.
+//
+// The two witnesses differ by a trailing `\Q`, which quotes the rest of
+// the pattern. Without it the probed pattern still compiles and it is the
+// wrapper check that refuses the rewrite; with it the probe's own closing
+// parenthesis falls inside the quoted run, so the pattern the rewrite
+// produces does not compile at all and the read-back refuses it first.
+var anchorBreakoutWitnesses = []struct {
+	name  string
+	regex string
+	// probed is the pattern the rewrite would hand back if the arm under
+	// test were not there. It is written out rather than derived so that
+	// a change in which span gets probed is visible here as a diff.
+	probed string
+}{
+	{
+		name:   "probe_swallows_the_closing_wrapper",
+		regex:  `a)|(?P<d>b?)(?P<d>c`,
+		probed: `^(?s:a)|(?P<` + probeNamePrefix + `0>(?P<d>b?)(?P<d>c)$)`,
+	},
+	{
+		name:   "quoted_run_swallows_the_probe",
+		regex:  `a)|(?P<d>b?)(?P<d>c)\Q`,
+		probed: `^(?s:a)|(?P<` + probeNamePrefix + `0>(?P<d>b?)(?P<d>c)\Q)$)`,
+	},
+}
+
+// TestPlanCaptureProbesRefusesARewriteItCannotHandBack pins that a rewrite
+// which does not survive its own verification is refused rather than
+// trimmed, patched, or handed back anyway.
+//
+// The assertion is the CONTRACT, not the decline: whenever the planner
+// reports success, the pattern it returns must be the caller's own
+// spelling with capture groups added and nothing else — it re-anchors to
+// a pattern that compiles, carries exactly the original's capture groups
+// in the original's order once the probes are struck out, and accepts
+// exactly the strings the user's own pattern accepts. Asserting only
+// "it declines" would be satisfied by a planner that declined for the
+// wrong reason.
+//
+// The two arms are not independent, and this test does not pretend they
+// are. The wrapper check is the one the contract catches: with its early
+// return removed the planner hands back the anchored text it was given,
+// and this test fails. The read-back's early return is DOMINATED by it —
+// removing that one leaves the plan at its zero value, whose empty regex
+// the wrapper check refuses instead, so no input tells the two apart and
+// the whole package passes with the mutation applied. What the read-back
+// verification itself does is pinned by
+// [TestReadBackPlanRejectsInvalidRewriteMetadata]; the second witness
+// here is what drives a real rewrite into it.
+func TestPlanCaptureProbesRefusesARewriteItCannotHandBack(t *testing.T) {
+	t.Parallel()
+
+	inputs := stringsUpTo("abc", 3)
+
+	for _, w := range anchorBreakoutWitnesses {
+		t.Run(w.name, func(t *testing.T) {
+			t.Parallel()
+
+			anchored := anchorRegex(w.regex)
+			original, err := regexp.Compile(anchored)
+			if err != nil {
+				t.Fatalf("anchoring %q gives %q, which must compile for this witness to be "+
+					"about the rewrite at all: %v", w.regex, anchored, err)
+			}
+
+			// The span the scanner picks is what makes this witness the
+			// shape it claims to be: it reaches the END of the anchored
+			// pattern, so the probe's own parenthesis lands past the
+			// wrapper's. Pinning it here means a future change that made
+			// the planner choose a different span would surface as this
+			// test losing its subject rather than as a silent pass.
+			groups, ok := scanSourceGroups(anchored)
+			if !ok {
+				t.Fatalf("the scanner declined %q outright, so this witness never reaches the "+
+					"rewrite it is about", anchored)
+			}
+			span, ok := probeSpanFor(anchored, groups, groupIndexOf(groups, 1))
+			if !ok || span.end != len(anchored) {
+				t.Fatalf("probeSpanFor(%q) chose %v (ok=%v); this witness is only about the "+
+					"rewrite arms when the span reaches the end of the anchored pattern, at %d",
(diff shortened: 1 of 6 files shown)


tsouza:
## The perf question this PR left open, now actually measured

The PR body said the local A/B was inconclusive and named no figure, deferring to CI's
`benchstat diff (baseline vs ref)` job. That job never ran the real comparison here — it is a
hard no-op on every `pull_request` event by design (`RUN_HEAVY: ${{ github.event_name != 'pull_request' }}`),
schedule/dispatch only. And even a dispatch against this branch alone would not have produced a
comparison: `scan_bench_test.go` was added in the same commit as the reshape, so no commit in
history has the old scanner bodies *and* these benchmark functions together — baseline would show
them as new with nothing to diff against.

So: confirmed no commit since this PR merged touched `internal/qlcommon` production code
(`git log baa2455d6..origin/main -- internal/qlcommon/` is empty), confirmed the reshape changed no
function signature (`git diff ac89ad058 origin/main -- internal/qlcommon/*.go | grep '^[+-]func '`
is empty), built a throwaway branch = the pre-reshape parent commit (`ac89ad058`) plus this PR's
`scan_bench_test.go` unmodified (it compiles against the old bodies without changes, confirming the
signatures really are identical), and dispatched `perf-benchmark.yml` with that branch as
`baseline_ref` against current `main` as `ref` — CI's own runner, `-count=6 -benchtime=1s -cpu=1`,
the statistical minimum `benchstat` needs for a confidence interval.

Run: https://github.com/tsouza/cerberus/actions/runs/33766247147 (success, all steps green).

```
pkg: github.com/tsouza/cerberus/internal/qlcommon
                                 │ baseline.bench │              ref.bench              │
                                 │     sec/op     │    sec/op     vs base               │
ScanSourceGroups/plain                1.014µ ± 0%    1.009µ ± 0%   -0.54% (p=0.002 n=6)
ScanSourceGroups/groups               2.646µ ± 2%    2.719µ ± 1%   +2.76% (p=0.002 n=6)
ScanSourceGroups/classes              620.9n ± 1%    646.6n ± 2%   +4.14% (p=0.002 n=6)
ScanSourceGroups/escapes              662.1n ± 0%    618.0n ± 0%   -6.65% (p=0.002 n=6)
SkipCharClass                         58.26n ± 2%    49.13n ± 0%  -15.67% (p=0.002 n=6)
ResolveSegments/literal               9.742µ ± 2%    9.857µ ± 1%   +1.18% (p=0.039 n=6)
ResolveSegments/refs                  9.758µ ± 0%   10.008µ ± 1%   +2.56% (p=0.002 n=6)
ResolveSegments/dollars               9.209µ ± 0%    9.301µ ± 1%   +1.00% (p=0.002 n=6)
EmptyCapturesReplacement/literal      1.199µ ± 2%    1.197µ ± 0%        ~ (p=0.180 n=6)
EmptyCapturesReplacement/refs         4.248µ ± 5%    4.401µ ± 1%        ~ (p=0.394 n=6)
EmptyCapturesReplacement/dollars      1.952µ ± 0%    2.093µ ± 0%   +7.25% (p=0.002 n=6)
geomean                               1.777µ         1.773µ        -0.24%
```

`B/op` and `allocs/op` are identical on every benchmark (`~`, p=1.000, "all samples are equal") —
the reshape changes no allocation shape, as expected from a control-flow-only change.

**Net**: geomean -0.24%, a wash. The individual deltas are real (all statistically significant at
`n=6`) but small and mixed, both directions. `SkipCharClass` is the one worth explaining rather than
shrugging at: it improved 15.67%, and the diff shows why — the pre-reshape loop carried an explicit
per-iteration progress check (`if j <= start { return 0, false }`) that ran on every byte; the
reshape's `next := j + 1` default plus one unconditional `j = next` removes that branch entirely.
The single-digit regressions on `ScanSourceGroups/groups|classes` and `ResolveSegments/refs` and the
+7.25% on `EmptyCapturesReplacement/dollars` are plausibly the mirror cost — an arm that now writes
`next` explicitly where the old code let a bare `j++` fall through as the default. None of this
approaches a scale worth reverting a structural safety property for, and it's within the same
neighbourhood as #2977's own measured cost for the identical technique (+0.31%/+1.98%/+0.27%).

Throwaway branch `perf/2998-scanner-reshape-baseline` deleted after this run.

Issue #1527, opened 2026-08-01
e2e seed: rolling re-seeder only deletes the b0… showcase traces — the a0… base fixture (and metrics/logs) stack one copy per tick


tsouza:
PR #762 recorded this in its body instead of an issue:

> The rolling re-seeder still stacks duplicate rows for the legacy `a0…` base fixture (unchanged); recursive closures now tolerate duplicates, but trace-detail views of the base fixture show N copies after long stack uptimes.

The follow-up was never filed. Still true on `origin/main`, and broader than the PR body says.

## Evidence

**The only stale-row DELETE in the whole seeder is scoped to the showcase traces.** `test/e2e/seed/cmd/seed/showcase_traceql.go:143-149`:

```go
const deleteStaleShowcaseTracesSQL = `DELETE FROM otel_traces
WHERE TraceId LIKE 'b00000000000000000000000000000%'
  AND Timestamp < (
    SELECT max(Timestamp) - INTERVAL 20 SECOND
    FROM otel_traces
    WHERE TraceId LIKE 'b00000000000000000000000000000%'
  )`
```

A repo-wide grep for `DELETE FROM` / `TRUNCATE` / `OPTIMIZE TABLE` under `test/e2e/seed/` returns **that one line and nothing else**.

**The base fixture is re-inserted on every tick.** `test/e2e/seed/cmd/seed/main.go:487-496` defines `insertTracesSQL`, re-inserting `a0000000000000000000000000000001..3` with `now64(9) - INTERVAL n SECOND` timestamps; `main.go:658` runs it (`return conn.Exec(ctx, insertTracesSQL)`); `seedAll` (`main.go:258-276`) calls `insertMetrics`, `insertLogs`, `insertTraces`, `insertShowcaseTraces` together; the ticker at `main.go:152` re-runs `seedAll` for the life of the stack.

**Wider than the PR body claimed:** because the single DELETE is `b0…`-only, the metrics and logs fixtures accumulate on the same schedule as the `a0…` traces. Every fixture except the showcase traces grows one copy per tick.

## Consequence

Trace-detail views of the base fixture show N copies after long stack uptimes (the PR body's own words), and any assertion over the metrics/logs fixtures that is sensitive to row counts rather than distinct series is uptime-dependent. That makes long-running local stacks and any lane that reuses a stack across ticks quietly divergent from a fresh one.

## What to do

[...]

Issue #3128, opened 2026-09-06
fix(solver): DataShardFanoutGate concurrency ceiling exceeded under real N=2/N=4 load


tsouza:
Part of epic #3074. Found while driving PR #3123 (issue #3122)'s real N=2/N=4 CI verification — a
distinct, unrelated correctness gap in the SAME `DataShardFanoutGate` admission-control mechanism
issue #3081 shipped, exposed only now because #3122's own fix let the datashard e2e lane's
`e2e-datashard-verify.mjs` script reach a genuinely-informative point 1/point 2 for the first time
(every earlier attempt failed upstream on #3105/#3118/#3119, then on #3122's own point 4).

## Problem

`e2e-datashard-verify.mjs`'s point 2 assertion — the real, concurrent, cluster-wide per-shard
statement count (`system.query_log` rows with `is_initial_query=0`) must never exceed
`DataShardFanoutCap` at any instant — **fails on both `datashard (N=2)` and `(N=4)`**, on dispatch run
34024076917:

```text
real per-shard statements observed: 3880; peak concurrent=14
##[error]peak concurrent per-shard ClickHouse statement count 14 exceeded DataShardFanoutCap=8
```

```text
real per-shard statements observed: 1788; peak concurrent=31
##[error]peak concurrent per-shard ClickHouse statement count 31 exceeded DataShardFanoutCap=8
```

**This is not new** — it is confirmed present, byte-identical in shape, in the very first captured
log this epic has (dispatch run 34020019580, captured before PR #3123 changed a single line):

```text
real per-shard statements observed: 2444; peak concurrent=12
##[error]peak concurrent per-shard ClickHouse statement count 12 exceeded DataShardFanoutCap=8
```

`DataShardFanoutGate` (`internal/solver/executor.go`'s `admitAndGate`, cerberus issue #3081) is
supposed to enforce `Σ(kEff_i × DataShardCount) <= DataShardFanoutCap` across every
concurrently-admitted request, process-wide, via a real `semaphore.Weighted` acquire — so the
observed peak (12/14/31) exceeding the configured cap (8) by 1.5x–4x means either the semaphore
isn't bounding what its own doc claims, or the TEST's measurement of "concurrent" doesn't correspond
1:1 to what the semaphore actually bounds (e.g. a `system.query_log` child row's wall-clock window —
`query_start_time_microseconds` + `query_duration_ms` — may include time the statement spent queued
[...]

Issue #3081, opened 2026-09-05
feat(solver): bound aggregate ClickHouse data-shard fan-out in admission control


tsouza:
Part of epic #3074.

## Problem

`internal/solver/executor.go`'s `admitAndGate` sizes `kEff = min(k, pEff, GateCap/2)`
(`executor.go:132-144`) and charges a single, **process-global** connection semaphore, `Gate`
(`executor.go:66-76`, documented as shared across every head), purely from cerberus-to-ClickHouse
connection count — one `sc.client.QueryCursor` call (`executor.go:468`) per admitted slot.

If the ClickHouse target becomes a `Distributed` table over N real data shards, everything past that
one admitted connection happens inside ClickHouse, invisible to this arithmetic. Critically, because
`Gate` is shared across every concurrently-routed request, a fix that only shrinks one request's own
`kEff` (e.g. dividing `pEff` by N) does not shrink the process-wide total: at saturation, `Σ kEff`
across all concurrently-admitted requests still approaches `GateCap`, so real ClickHouse-side
concurrent statement count approaches `GateCap × N`, uncapped. A naive per-request divide is also
independently broken at this repo's own shipped default (`defaultParallel = 3`,
`internal/solver/config.go:222`): `floor(3 / N)` is `0` (clamped to `1`) for every `N ≥ 4`, so it
provides zero additional attenuation as cluster size grows past 3 — exactly the regime this epic
exists to support.

## Design

Add `internal/chopt.ClusterTopology.DataShardCount int` (default `1`), the single source of truth
for the ClickHouse data-shard count, read once at startup (Helm-bundled: `CERBERUS_CH_DATA_SHARDS`
env var; externally-managed cluster: a one-time `system.clusters` probe). `internal/solver.Config`
reads it once at construction into `solver.Config.DataShardCount` — same name, used identically in
both packages, each site carrying a comment disambiguating it from `kEff`/`pEff` and from
`perShardMemoryBytes`/`runShard` (this package's own pre-existing bare-`Shard` identifiers).

Two separate fixes, addressing two separate risk vectors — neither is a divisor applied to the
other's inputs:

**Concurrency — a new, dedicated global semaphore, not a per-request divide.**

- `internal/solver.Executor.DataShardFanoutGate *semaphore.Weighted` — `nil` (structurally inert)
  whenever `DataShardCount ≤ 1`, so a single-shard deployment never allocates or exercises this
  path at all.
[...]

Issue #3122, opened 2026-09-06
fix(solver): per-shard max_memory_usage apportionment not applied under DataShardCount>1


tsouza:
Part of epic #3074. Found while driving issues #3105/#3118/#3119's real CI verification (dispatch
run 34020019580, the first run to ever get PAST all of cluster bring-up, seeding, TraceQL, and
PromQL query execution to reach the lane's own `e2e-datashard-verify` correctness check).

## Problem

`.github/scripts/e2e-datashard-verify.mjs`'s point-4 assertion — every solver-dispatched
statement's observed `max_memory_usage` (read from `system.query_log`) should stay within
`cap/(kEff*DataShardCount)` (`internal/solver/executor.go`'s documented per-shard memory
apportionment, issue #3081) — fails on both `datashard (N=2)` and `(N=4)`:

```
observed max_memory_usage=1073741824 (trace=..., kEff=1) is not within (0, 536870912]
predicted by perShardMemoryBytes = cap/(kEff*DataShardCount) with cap=1073741824, kEff=1, DataShardCount=2
```
(N=2: 117 assertions failed; N=4: 19 assertions failed, spanning both kEff=1 AND kEff=2 traces.)

Every single failing observation reports `max_memory_usage` as EXACTLY the raw, unscaled `cap`
(1073741824 = `CERBERUS_CH_QUERY_MAX_MEMORY_BYTES`) — never a value consistent with ANY
apportionment, regardless of kEff or DataShardCount. This is the first time this specific
assertion has ever run against a real multi-shard cluster; every earlier attempt failed upstream
(issues #3105, #3118, #3119, all now fixed).

## What's known

- `internal/solver/executor.go`'s apportionment code (`perShardMemoryBytes = cap /
  (int64(kEff) * int64(dataShardCount))`, stamped via `chclient.WithQuerySetting(pctx,
  "max_memory_usage", perShardMemoryBytes)`) reads correctly and DOES reference
  `x.Cfg.DataShardCount` unconditionally (not gated on kEff > 1) — so on paper this should apply
  to kEff=1 traces too.
- The doc comment immediately above that code distinguishes "Route A" (a single statement that
  never enters the solver's K-shard fan-out, running under the client's raw configured
  `max_memory_usage`, unapportioned) from the K-shard fan-out path (apportioned). Whether kEff=1
  in `system.query_log` actually correlates 1:1 with "took Route A, apportionment code never ran"
[...]

PR #3132, opened 2026-09-06
fix(chclient): move data-shard fanout gate from solver Executor to chclient dispatch

tsouza:
Moves `DataShardFanoutGate` (issue #3081) from `internal/solver`'s `Executor`
(which only gated "route B", the solver's K-shard split path) into
`internal/chclient`, acquired at the one seam every dispatch this package
makes — `queryOpen` and `queryCursorColumnar` — so "route A" (the ordinary,
non-split single-statement path, the vast majority of real traffic) is
finally covered too. Route A dispatches against a `Distributed`-engine table
also fan out to `DataShardCount`-many real per-shard ClickHouse statements,
and previously never acquired the gate at all.

## Status: NOT closing #3128 — real limitation found and documented, not yet fixed

Two independent real e2e dispatches surfaced a genuine gap in this
mechanism. This PR is being left open, with the finding documented in code
(`internal/chclient/fanout_gate.go`'s doc comment) and here, rather than
force a fix that doesn't hold up.

### What the diagnostic proved

A temporary held-weight counter + WARN (`chclient: data-shard fanout gate
held weight exceeds cap`, since removed) was added to
`acquireDataShardFanout` to distinguish two hypotheses: a real accounting
bug (double-acquire / missed release) vs. a timing/measurement gap.

Across three real e2e runs with the diagnostic live, **the WARN never fired
once** — cerberus's own gate bookkeeping never believed it held more than
`DataShardFanoutCap`. This rules out an accounting bug in the semaphore
logic itself.

### The real root cause: a cancellation-driven timing gap, not an accounting bug

Real cluster evidence (dispatch run
[34032272914](https://github.com/tsouza/cerberus/actions/runs/34032272914),
second attempt, commit `9528638a0`'s parent `d9b9068ee`):

| Leg | Peak concurrent (real, `system.query_log`) | `DataShardFanoutCap` | Overshoot |
|---|---|---|---|
| N=2 | 10 | 8 | +2 (= 1 × `DataShardCount`) |
| N=4 | 24 | 8 | +16 (= 4 × `DataShardCount`) |

(Earlier attempts on the same run measured N=2=34/N=4=11 and N=2=blocked/
N=4=23 — see run history; the pattern below held across every attempt where
data was collected: overshoot is consistently an integer multiple of
`DataShardCount`, never a diffuse/random excess.)

The mechanism, verified against clickhouse-go v2 v2.48.0's actual vendored
source (`conn_process.go`'s `process()`/`firstBlock()`/`cancel()`):

- `acquireDataShardFanout`'s release fires the instant
  `queryOpen`/`queryCursorColumnar`'s underlying `c.conn.Query` / `pool.Do`
  call **returns** — this was already correctly wired and verified against
  the *normal* completion path (a full, undisturbed drain genuinely blocks
  until the server finishes; confirmed via `clickhouse_rows.go`'s `Close()`,
  which drains the `stream`/`errors` channels to completion).
- But when that call returns because its **ctx was cancelled** —
  `internal/solver/executor.go`'s `errgroup.WithContext` cancels every
  sibling shard's ctx the instant ANY ONE of a routed query's `kEff`
  concurrently-admitted shards errors, and an HTTP client
  disconnect/timeout cancels `r.Context()` the same way, both of which
  chclient's dispatch inherits — clickhouse-go v2's `connect.cancel()`
  fires a single `ClientCancel` packet and **closes the local connection
  immediately**, without waiting for ClickHouse to confirm the (possibly
  still-running, possibly Distributed-fanned-out-to-every-data-shard)
  statement has actually stopped server-side.
- This package therefore releases that dispatch's `DataShardCount` weight
  the instant the *local* socket closes, while the *real* per-shard
  ClickHouse statement(s) it dispatched may keep running server-side for an
  uncontrolled further interval — during which the gate believes that
  capacity is free and admits a new dispatch on top of it.
- The observed overshoot fits this exactly: each premature-cancellation
  release frees (and over-admits by) one dispatch's whole
  `DataShardCount`-sized weight, so the excess is a small integer multiple
  of `DataShardCount`, not a diffuse accounting drift — matching both real
  measurements above bit-for-bit.

Also checked and ruled out along the way:
- Multiple independent gate instances, cap-sizing-basis drift, and
  `Client.Exec` bypassing the gate for the real query-serving path — all
  previously ruled out by static review, reconfirmed here.
- `queryCursorColumnar`'s own path: not reached at all on this e2e lane —
  `columnar_result_decode` is opt-in-only and never enabled in
  `test/e2e/k3s/cerberus-values-datashard.yaml`, so the row path
  (`queryOpen`/`gatedRows`) is the only path actually exercised.
- The verify script's own point-2 measurement itself (window, sweep-line
  logic) — structurally sound; a real (if second-order) contamination
  source is the always-running rolling re-seeder's own INSERTs against the
  same `Distributed` public table names, uncounted by the script's
  `is_initial_query=0` sweep since it doesn't filter by `query_kind` —
  worth a follow-up but not the dominant effect given the clean
  `DataShardCount`-multiple fit above.
- An unrelated, pre-existing e2e flake (`TestReSeedRowCountStability/
  base-traces`, citing already-closed issue #1527) blocked the N=2 leg on
  the diagnostic run's first attempt; confirmed unrelated to this PR's
  changes (this branch's `stale.go` is byte-identical to main's post-#3130
  version) and did not recur on the rerun that produced the N=2=10 result
  above.

### What was fixed vs. what remains

- Removed the temporary diagnostic (`dataShardFanoutHeld` counter + WARN) —
  it served its purpose (ruled out an accounting bug) and is not being kept
  as permanent instrumentation.
- Documented the confirmed limitation directly in
  `internal/chclient/fanout_gate.go`'s doc comment, including the concrete
  fix direction: on the cancellation-unwind path specifically, issue `KILL
  QUERY WHERE query_id = ? SYNC` (every dispatch already mints a query_id)
  on a fresh, uncancelled connection **before** releasing that dispatch's
  gate weight, so the release genuinely corresponds to ClickHouse having
  stopped rather than to the local socket having closed. This has **not**
  been implemented — it needs its own connection-pool-access design,
  bounded-timeout handling for the KILL itself, and a fresh
  cluster-validation cycle, which is more than this investigation's
  remaining scope covers responsibly.
- Route A's admission gap this PR closes (it was completely ungated before)
  is still a genuine improvement over the pre-move state — but the gate is
  not yet a hard ceiling on ClickHouse's own real concurrent per-shard
  statement count under cancellation, so #3128's acceptance criteria
  (`peak concurrent per-shard ClickHouse statement count <=
  DataShardFanoutCap` on both N=2 and N=4, with zero overshoot) are **not**
  genuinely met yet. Leaving #3128 open, not closing it from this PR.

## Local verification on the final commit

- `go build ./...`, `go vet ./...` — clean.
- `go test -race ./internal/chclient/... ./internal/solver/... ./internal/config/...` — pass.
- `gofumpt -l .` / `goimports -local github.com/tsouza/cerberus -l .` — clean on the changed files.
- Full e2e round-trip: run
  [34032272914](https://github.com/tsouza/cerberus/actions/runs/34032272914)
  (both attempts) — `datashard (N=2)` and `datashard (N=4)` both still fail
  point 2 of `e2e-datashard-verify.mjs` for the reasons documented above.

https://claude.ai/code/session_016H8AVBwHzAYcMi4kuYub9F


tsouza:
Do NOT merge this yet — real CI evidence shows the fix makes the actual overshoot WORSE, not
better, on one leg. This needs further investigation before it's mergeable.

## Real CI evidence (run 34030324992, this branch's second commit, 503c64756)

- `datashard (N=2)`: **peak concurrent per-shard statement count 34**, exceeding
  `DataShardFanoutCap=8` by over 4x. The pre-fix baseline for N=2 (issue #3128's own body, run
  34024076917) was 14 — this got substantially worse.
- `datashard (N=4)`: peak concurrent 11 (down from the pre-fix baseline of 31) — improved, but
  still over cap.

## What I checked and ruled out (static code review)

- `CERBERUS_SOLVER_DATA_SHARD_FANOUT_CAP=8` is explicitly set in
  `test/e2e/k3s/cerberus-values-datashard.yaml` and correctly threads through
  `dataShardFanoutCapOverrideFromEnv` → `chConfigInputs.dataShardFanoutCapOverride` →
  `chclient.Config.DataShardFanoutCapOverride` → `NewDataShardFanoutGate`'s resolved cap — this is
  NOT a cap-sizing-basis drift between the old (`solver.NewDataShardFanoutGate(cfg, gateCap)`) and
  new (`chclient.NewDataShardFanoutGate(cfg)`) default-derivation paths, since the explicit override
  wins in both.
- `ForHead` is a genuine shallow struct copy (`view := *c`), so `dataShardFanoutGate
  *semaphore.Weighted` (a pointer field) is correctly SHARED across every head's `*Client` view —
  not re-derived per head. Only one base `*chclient.Client` is constructed in `cmd/cerberus/main.go`
  (every `promClient`/`lokiClient`/`tempoClient`/probe client is a `.ForHead(...)` view of it), so
  this isn't a "multiple independent semaphores" bug either.
- `chclient.Client.Exec` (client.go:1465) calls `c.conn.Exec` directly, bypassing
  `acquireDataShardFanout` entirely — but grepping every caller confirms it's used only for
  DDL/schema/mutation paths (`internal/optcorpus`, `internal/schema/ddl`, the e2e seeder's own raw
  `driver.Conn`, `internal/chopttest`), never in the real PromQL/LogQL/TraceQL query-serving path
  (`internal/api`, `internal/engine`, `internal/solver` all route through `QueryCursor` →
  `queryOpen`/`queryCursorColumnar`, which DO acquire the gate). So this isn't the leak either,
  though it's worth double-checking there's no OTHER less-obvious dispatch path I missed.
- `.github/scripts/e2e-datashard-verify.mjs`'s own point-2 measurement (the `maxConcurrent`
  sweep-line over `is_initial_query=0` / `type='QueryFinish'` rows, windowed over the whole burst +
  flush slack) was NOT touched by PR #3127's kEff-measurement fix (that was point 1, a different
  query) and looks structurally sound on inspection — I don't have a specific reason to suspect the
  TEST'S measurement here, unlike the earlier #3122/#3127 episode.

## Leading hypothesis, unverified

`acquireDataShardFanout`'s release is tied to `gatedRows.Close()` — i.e., cerberus's OWN belief
that "the dispatch is done" is whenever the CALLER finishes draining/closes the cursor. For a
Route-A query that reads only a partial result (an early `LIMIT`, or a metadata/probe query that
stops after a few rows), `Close()` may return well before ClickHouse's OWN server-side execution of
the (potentially Distributed, multi-shard) query is actually complete — i.e., before the
`system.query_log` `QueryFinish` event for that statement's per-shard children actually fires. If
so, cerberus's gate under-counts real hold duration for exactly this class of query, admitting new
dispatches while old ones are still genuinely running server-side — which would show up exactly as
"real observed concurrency exceeds what cerberus's own bookkeeping ever believed it admitted."

This is a hypothesis, not a confirmed root cause — needs real investigation (ideally against a live
cluster, correlating actual gate acquire/release timestamps against `system.query_log`'s own
`query_start_time_microseconds`/`query_duration_ms` for the SAME statements) before committing to a
fix.

## Status

Not mergeable as-is. Continuing investigation.

Issue #3141, opened 2026-09-06
fix(chclient): audit remaining self-referencing TraceQL shapes for DataShardFanoutGate under-charge


tsouza:
## Background

Part of epic #3074, follow-up from issue #3128 round 4.

Round 4 of the `DataShardFanoutGate` investigation (issue #3128) proved, with
direct real-cluster evidence (an isolated re-run + `EXPLAIN PIPELINE` against
a real over-width dispatch, both captured in CI run 34051673070), that
`internal/chsql/search_trace_limit.go`'s `emitSearchTraceLimit` emits its own
row source TWICE (an outer drain query plus an inner top-N trace-ranking
subquery, both scanning the same `Distributed` table) — and that under
`distributed_product_mode=global` (pinned unconditionally, cerberus issue
#3118), this self-reference makes ClickHouse execute roughly twice the real
per-shard `Select` statements a single `chclient` dispatch's fixed
`DataShardCount`-wide gate weight assumes. The fix (in the PR that files this
issue) adds a per-request fan-out multiplier
(`chclient.WithDataShardFanoutMultiplier`) that `internal/engine.execContext`
stamps to 2 specifically for a `chplan.SearchTraceLimit`-shaped plan.

## The gap

`internal/chclient/distributed_query_settings.go`'s own doc (predating this
fix) already names **four other** query shapes that self-reference the same
`otel_traces` Distributed wrapper table within one physical SQL statement,
under the identical `distributed_product_mode=global` rewrite:

- TraceQL's structural-child/parent/sibling operators (`>` / `<` / `~`) —
  `internal/chsql/structural_join.go`'s `emitStructuralDirectJoin` /
  `emitStructuralSiblingJoin`.
- TraceQL's descendant/ancestor operators (`>>` / `<<`) —
  `emitStructuralRecursive`'s `WITH RECURSIVE` self-join, plus every
  structural operator's `rootedStructuralLeftSub` gate.
- `select(nestedSetLeft, nestedSetParent, nestedSetRight)` —
  `internal/chsql/nested_set_annotate.go`'s `emitNestedSetAnnotate`.
- `| compare(...)` — `internal/traceql/metrics_compare.go`'s
  `compareRootLookup`.

None of these four has been round-4-verified against a real
`DataShardCount > 1` cluster the way `SearchTraceLimit` now has — issue
#3128's own e2e burst never exercises them (it fires only a bare,
non-structural TraceQL attribute search, a PromQL range query, and a LogQL
[...]

PR #3143, opened 2026-09-06
fix(helm): apportion a cluster-wide DataShardFanoutCap across cerberus replicas

### LINKED ISSUE #3128: fix(solver): DataShardFanoutGate concurrency ceiling exceeded under real N=2/N=4 load

tsouza:
Resolves #3128 — the admission-ceiling finding, end to end (see the closing note at the bottom for what is verified and what is not).

## Root causes (three, found in this order, each with real-cluster evidence)

1. **Per-process gate measured cluster-wide.** `DataShardFanoutGate` is a per-process semaphore. The datashard e2e lane deployed cerberus through the base overlay, which enables the HPA (2–4 replicas), so 2–4 pods each held a cap of 8 while `e2e-datashard-verify.mjs` asserted the cluster-wide peak against 8.
2. **`SearchTraceLimit` re-fanned its ranking subquery out on every shard.** Its SQL reads the traces table through a derived table on both arms; `distributed_product_mode=global` only rewrites an `IN` whose subquery reads the Distributed table *directly*, so each shard executed the ranking subquery as its own distributed query: N²−1 children per dispatch (3 at N=2, 15 at N=4, up to 10 concurrent). Reproduced on a local two-node ClickHouse rig — plain `IN` = 3 children, explicit `GLOBAL IN` = 2.
3. **A per-shape multiplier could never keep up.** With the trace search fixed, the remaining over-width dispatches were PromQL `rate()` range queries: the SQL renders the metrics table three times (three `Union(ReadFromMergeTree, ReadFromRemote)` blocks in the captured `EXPLAIN PIPELINE`), so a dispatch charged N produced 3N statements.

## Fix

**Chart — a real cluster-wide budget.** New `clickhouse.bundled.dataShards.fanoutCap` (default `null`) is apportioned by the new `cerberus.effectiveReplicas` helper (HPA `maxReplicas`, else `replicaCount`, split-mode head sum) into the binary's per-process `CERBERUS_SOLVER_DATA_SHARD_FANOUT_CAP`. Refuses a share below `dataShards.count`, a *different* explicit per-process value, and `count: 1`; idempotent across the several `nonSecretEnv` passes. Schema, values docs, helm-docs README, `chart-render-assert.mjs` (every divisor source, the floor, all refusals, idempotency, default emits nothing).

**Config — refuse a cap that can never admit.** `FromEnv` rejects an effective cap (override or `CERBERUS_CH_MAX_OPEN_CONNS`) below `CERBERUS_CH_DATA_SHARDS`, naming the variable to raise.

**Emitter — `GLOBAL IN` for `SearchTraceLimit`.** New `chsql.GlobalInSubquery`; `emitSearchTraceLimit` uses it. The initiator ranks once and broadcasts the LIMIT-bounded id set. Goldens regenerated (traceql spec + migration explain) via `update-golden.yml`.

**Gate — charge the statement's physical scan count.** The emitter counts every physical table it renders (`Builder.physicalScans`: once per rendered table reference, once per *splice* of a pre-rendered sub-statement, once per `merge()` member, none for `system.*`); `chsql.EmitCounted` returns it; route A (`engine`) and route B (`solver.runShard`) stamp it as the dispatch's `WithDataShardFanoutMultiplier`. The per-shape constant and `planHasSearchTraceLimit` are gone. A statement wider than the whole cap is admitted alone with the cap as its weight rather than parked until its deadline (`semaphore.Weighted` never admits a weight above its size). Tests: `EmitCounted` counts cross-checked against the emitted text (1 / 2 / 3 / 0), per-splice counting, the width clamp.

**e2e — assert the contract that exists.** Overlay pins `replicaCount: 2` (HPA off) and `dataShards.fanoutCap: 96` (48 per process — `rate()` over the 3-member metrics `merge()` union is 6 scans, 24 wide at N=4, and the lane must admit two at once to observe a solver split). The verify script reads cap and replica count back from the live ConfigMap/Deployment, attributes every Select child to the cerberus pod that issued it via `query_log.client_hostname` (the rolling seeder's own direct Selects are reported, never counted; an *empty* host is an error), and asserts each pod ≤ per-process cap and the cluster ≤ replicas × cap.

## Verification

- Local two-node ClickHouse rig (Docker, `remote_servers`, no Keeper): per-shard children for every SQL shape variant, confirming plain `IN` = N²−1 and `GLOBAL IN` = 2(N−1) remote children.
- `go build ./...`, `go vet ./...`, `go test ./internal/... ./test/regression/...` green; `node .github/scripts/chart-render-assert.mjs` all PASS; `helm lint`; helm-docs and `just gen-config-docs` regenerated.
- One confirming `datashard_only=true` e2e dispatch on the final head (run linked in the PR comments).

## Why this closes #3128

Closes #3128

The admission ceiling is resolved: the ceiling is now defined per process and per cluster, charged by the statement's real width, and asserted per pod. The over-width *correctness* audit of the other self-referencing TraceQL shapes stays on #3141 (updated there with the derived-table `GLOBAL` finding).


tsouza:
Confirming `datashard_only=true` e2e dispatch on the final head (`480e6c406`): https://github.com/tsouza/cerberus/actions/runs/34059346299 — all assertions passed on both legs.

| leg | per-pod peak Select fan-out (cap 48) | cluster peak (cap 96) | solver-split groups (kEff > 1) |
|---|---|---|---|
| N=2 | 8 / 6 | 10 | 34 (max kEff 3) |
| N=4 | 12 / 13 | 14 | 43 (max kEff 2) |

The seeder's own direct Selects (31 / 78 children from the runner host) were attributed and excluded, never counted. The preceding run on the 16-per-process budget (34058864698) had already held the ceiling on every pod (4/4 and 9/9) and failed only the "solver split observed" check, because a 24-wide `rate()` statement at N=4 left no room for a second one — the budget is now sized so two fit.

Issue #3033, opened 2026-09-04
solver: fold route-B's K per-shard actuals into ONE per-request observation


tsouza:
## Problem

`internal/chclient/progress.go`'s `WithProgressFor` correctly gives each of a route-B sharded
dispatch's K shards its own fresh `progressRecorder` (fixed by the PR filing #2853: previously a
shard's fresh recorder silently dropped `actualsIntent`). But each shard's recorder independently
calls `tracker.RecordActual(shapeID, ...)` from its own `flush()` (on that shard's cursor Close) —
so a K-shard dispatch makes **K independent `RecordActual` calls**, each carrying only that shard's
own `(rows, bytes, peakMemory)` share, against the ONE un-sharded `RecordPredicted` prediction the
whole request made. This under-counts the true per-request cost by roughly a factor of K relative to
what the prediction describes, and `internal/actuals.Tracker.CalibrationFactor` (which feeds
`internal/engine`'s `calibrateEstimate`) inherits the corrupted EMA — everything downstream of it,
including the per-rung admission learner's actuals-driven seed
(`maybeSeedPerRungAdmissionFromActuals`), is calibrated wrong by the shard count.

## Fix

Fold route-B's K shard totals into ONE per-request actuals observation before it reaches the
tracker, in `internal/solver/executor.go` / the shard-composition seam (`shardCursor`) — not K
independent ones. Per the folding rule:
- `ReadRows` / `ReadBytes`: SUM across shards (each shard scanned a disjoint slice of the total).
- `MemoryTrackerPeakUsage`: MAX across shards, never sum — `progress.go`'s `onProfileEvents` already
  latches the max PACKET within one shard's own dispatch; summing K shards' independent peaks would
  overstate the actual peak (the K shards' cursors are not necessarily concurrent in a way that makes
  a summed "peak" meaningful, and ClickHouse's own per-query MemoryTrackerPeakUsage is already a
  single-process high-water mark, not additive across independent queries).

The natural shape: each shard's `progressRecorder.flush()` should feed its own
`(rows, bytes, peakMemory)` into a shared per-DISPATCH accumulator (not call `tracker.RecordActual`
directly), and the accumulator makes exactly ONE `RecordActual` call once all K shards are known
complete — likely at `shardCursor`'s own Close/composition point, since that already owns the
lifecycle of "all shards done." Route A (single dispatch, no sharding) must be unaffected — it should
still record its one observation exactly as today.

## Definition of done

[...]

Issue #2705, opened 2026-08-28
solver,engine: route-B shards emit the density guard calibrated for the whole cap, not their kEff share


tsouza:
## Summary

`routeBExecCtx` threads the deployment's resolved
`RangeBucketGridNativeMaxRows` / `MaxDensityUnits` into every route-B shard
(`internal/engine/engine.go:1574-1575`), un-apportioned — while the same
dispatch gives each shard only `cap / kEff` bytes
(`internal/solver/executor.go:252`, stamped as `max_memory_usage`).

So a shard executes with 1/kEff of the memory but the density ceiling
calibrated for the **whole** cap. With `defaultMaxK` now 32
(`internal/solver/config.go:166`), that ceiling can be up to 32x too permissive
relative to what the shard is actually allowed to use.

The units are a near-linear proxy for bytes — `rbgnDensityUnitsPerGiB` is
9,000,000 against a measured cliff of ~11.7M/GiB — so there is only ~1.3x of
margin absorbing the discrepancy, nowhere near 32x.

## Consequence

A shard can pass its own density guard and then be aborted by ClickHouse's
`max_memory_usage`. That is the "sailed past the guard and died on ClickHouse's
own limit" mode #2677 / #2681 closed for route A, still open on route B.

Bounded, which is why this is not a blocker:

- **No wrong results.** The shard aborts; the memo records the route-B failure;
  the request converges back to a 422.
- **Not a regression.** Before this threading existed the shards used chsql's
  compiled-in 400,000,000 fallback, which is *looser* still. Threading the
  resolved bound tightened them; it just did not go all the way to the shard's
  own allowance.

What is lost is the cheap pre-flight rejection and its diagnostic
cost/bound/anchors/width numbers (#2693), plus the ClickHouse work the fan-out
burns before aborting.

## Why it is not a one-line fix

`kEff` is computed at `executor.go:192-205`, **after** the emit loop at
`executor.go:145-150`. Apportioning the ceiling therefore needs either a
reorder of the executor so the per-shard allowance is known before emit, or a
MaxK-conservative value threaded at emit time. Both are real changes to the
routed dispatch path, and nothing currently pins the property —
`internal/engine/route_b_tsgrid_setting_test.go` only calls `routeBExecCtx`
with `0, 0`.

## Scope

Apportion the rows/density ceilings by the same `kEff` the memory cap uses, and
pin it: a test that a routed dispatch emits a shard whose density ceiling is
[...]

PR #3201, opened 2026-09-08
fix(engine,telemetry): give route B route A's bounds, settings and observability

### LINKED ISSUE #3184: engine: route B diverges from route A in bounds, settings and observability
## What

The ACPR (#3180) audited route B (the sharded-pushdown path) against route A and
found that the two diverge in bounds, settings, observability and admission.
Distributed / multi-data-shard mode is EXPERIMENTAL and off by default, so none
of this is a shipped-default defect — but every item below weakens a bound or
falsifies a claim the code makes about itself.

`ReanchorRange` dropping `Project.Replacements` was the wrong-answer member of
this family and is fixed in #3180.

## Bounds and admission

- **The memo-hit dispatch releases its `AdmitDispatch` token at cursor OPEN, not
  at drain end.** `route_memo_wiring.go:223` defers `release()` and
  `dispatchDone()` immediately, and `Executor.Execute` returns as soon as the K
  shard cursors are opened, so the shards' ClickHouse work is entirely outside
  the token's span. Proven with an overlay test: 4 of 4 tokens re-admitted with
  the cursor still open and undrained. The retry path holds to drain end, so the
  two sibling paths disagree, and `docs/solver.md`'s process-wide K-shard
  concurrency bound is hollow on what the file itself calls "the SINGLE MOST
  common route-B dispatch shape".
- **`scalarInteriorAnchorCompatible` has no `RangeBucketGridNative` arm.** The
  kind IS registered slice-invariant since #2677, so the generic `Children()`
  fall-through admits a full-span classic-histogram ladder inside a scalar or IN
  subquery as CHEAP, and route B replicates it K times — the heaviest shape in
  this repo's incident history. Its fan-out sibling `RangeBucketFanout` is
  explicitly refused two lines above for exactly this reason.
- **The route memo escalates RBGN row/density rejections to route B, but those
  verdicts are K-invariant.** `engine.go:1893` apportions both RBGN ceilings by
  K, so `groups x anchors/K > maxRows/K` is the identical verdict route A just
  produced. Every such escalation is a guaranteed re-failure that burns a scarce
  dispatch token. `route_outcome.go:117`'s rationale ("sharding genuinely lowers
  each shard's cost below the same bound") was true before #2705 apportioned
  them, and `engine.go:1795` says the opposite in the same package.

## Settings and observability

- **Route B never applies `SettingsRules.apply`.** `settings().apply` is called
  only in `execContext` (route A), so a routed query carries no `workload` —
  contradicting the env var's own doc, "stamp on every outgoing query" — no
  `log_comment` shape id, no result cache, no aggregation-in-order. Meanwhile
  `observeRoutedQuery` records `enabledOpts()` for the route-B corpus row, so
  calibration compares an optimized A against an un-optimized B and attributes
  the cost to a posture that never rode the shards.
- **The route-B eager path drops the outcome stamping route A performs.**
  `observeOutcomeForErr` is called at exactly two sites, both route A, so a
  routed drain that hits a sample budget, byte budget, OOM or open breaker is
  never recorded — the corpus keeps it as an "ok"-by-join. Every calibration
  that keys on `exit_status` by route is biased.
- **A cancelled query only gets a `KILL QUERY` on the experimental sharded
  path.**

## Actuals

- **The actuals tracker records every completed query twice** — once from the
  progress-packet path and once from the `system.query_log` poller, which has no
  query-id set-difference against what the packet path already recorded. With
  `MinObservations` = 2, one query satisfies the corroboration floor. For route B
  the EMA is additionally dragged toward per-shard rows by K poll samples per
  request, reopening #3033 through the other source. These feed the K clamp and
  per-rung admission, so it is a wrong routing input, not only a wrong metric.

## Related, same package

- `internal/downsampletier`'s backfill runs `INSERT … SELECT` under
  `read_overflow_mode=break`, so a read-cap trip silently persists a PARTIAL
  `timeSeriesLastTwoSamplesState` per bucket. `Verify` counts buckets, so a
  wrong-but-present bucket is invisible — the third state the package's
  "absent, never wrong" safety argument says cannot exist.
- `internal/routerrules`' ClickHouse reader uses the same `break` mode, so a
  watermark can be fitted on a silently truncated population.

### LINKED ISSUE #3197: telemetry: a client cancellation is bad_request on Tempo and backend_unavailable elsewhere
## What

A client cancellation is recorded under a different `cerberus_error_reason` on
Tempo than on Prometheus and Loki, for the same event.

- Tempo: `internal/api/tempo/errclass.go` maps a cancellation to 499, which is
  a 4xx, so the reason resolves to `bad_request`.
- Prom and Loki: the same cancellation surfaces as 503, so the reason resolves
  to `backend_unavailable`.

Neither is right. The caller went away; the request was not malformed and the
backend was not unavailable.

## Why it is filed rather than fixed

This is the third mis-attribution caused by deriving the reason from the HTTP
status. #3196 fixed the first two — `timeout` and `resource_exhausted` were
structurally unreachable and every timeout read as `backend_unavailable` — by
adding a request-scoped `telemetry.SetReason` and a shared
`httperr.TelemetryReason` classifier, without moving any status.

Fixing this one the same way needs a decision that PR should not make on its
own: `cerberus_error_reason` is a closed enum published in
`docs/observability.md`, and a cancellation has no honest member of it today.
Either a `canceled` value is added — a change to the public metric contract,
which the contract test deliberately pins — or cancellations are excluded from
the error population entirely, since a caller hanging up is arguably not a
query failure at all.

## Done when

Cancellations record the same thing on all three heads, and
`docs/observability.md` says which value that is and why.

Found while fixing #3185.

tsouza:
## What

Closes two issues from the #3180 audit. Every claim in both was re-verified against the code before anything was changed; one claim is **refuted** below and deliberately not "fixed".

Closes #3184

Closes #3197

## #3184 — route B's guarantees now match route A's

Distributed / multi-data-shard mode is EXPERIMENTAL and off by default, so none of this is a shipped-default defect. No sharding capability is added; route B is only brought up to the guarantees route A already had.

**The dispatch token was released at cursor OPEN.** `Executor.Execute` returns as soon as the K shard cursors are opened, and the memo-hit deferred `release()` across that return — so the shards' entire ClickHouse execution ran outside the token's span and `maxConcurrentRoutedDispatches` bounded nothing on what the file itself calls the single most common route-B dispatch shape. The A→B retry sibling held to drain end, so the two disagreed. Both now release through one `dispatchTokenCursor`, on `Close` — a stronger seam than the retry's own, which released from the *optional* `ObserveDrainOutcome` hook and leaked a token for any caller that skipped it.

**Route B never applied `SettingsRules.apply`.** A routed query carried no `workload` (contradicting that env var's documented "stamp on every outgoing query"), no `log_comment` shape id, no result cache, no aggregation-in-order — while `observeRoutedQuery` recorded `enabledOpts()` for its corpus row regardless, so calibration compared an optimized A against an un-optimized B and attributed the difference to a posture that never rode the shards. The two seams kept parallel hand-maintained lists, which is *how* they drifted; they now share one `applySharedQuerySettings`, parameterised only by the memory cap the statement actually runs under.

**The route-B eager path dropped outcome stamping.** `observeOutcomeForErr` was called at exactly two sites, both route A, so a routed eager drain that hit a sample budget, byte budget, OOM or open breaker was never recorded and the corpus kept it as an "ok"-by-join.

**`scalarInteriorAnchorCompatible` had no `RangeBucketGridNative` arm.** #2677 registered the kind slice-invariant, which silenced the sweep that had been refusing it inside a scalar/IN subquery — so the generic `Children()` fall-through began admitting a full-span classic-histogram ladder as CHEAP, and route B replicates an Expr-embedded interior K times unsliced. Held to the same grid+cadence equality `RangeWindowGridNative` gets, rather than refused outright.

**RBGN escalations were guaranteed re-failures.** #2705 apportions both ceilings by K (it must, or a shard is guarded against a whole-query ceiling while running under 1/K of the memory), which makes the verdict K-invariant: a shard is judged by `maxRows/K` against a cost of `groups × anchors/K` — the identical inequality route A just failed. Both messages move to the exclusion set, whose own doc already stated the test they fail. The three fan-out guards stay: their ceilings are threaded un-apportioned, so cost falls while the ceiling holds.

**The actuals tracker recorded every query twice.** The packet path and the `system.query_log` poller feed the same Tracker with no query-id set difference. The poller's `type = 'QueryFinish'` filter also means it cannot see the failed-query case its doc says it exists for, so the overlap was near-total. `MinObservations` is 2, so one query satisfied a corroboration floor whose purpose is that one observation is never enough; and on route B the poller saw K per-shard rows against the fold's one summed observation, dragging the EMA toward `total/K` (measured: 1000 → 557.2 at K=4) — re-opening #3033 through the other source. These feed the K clamp and per-rung admission, so it was a wrong routing input.

**`read_overflow_mode: "break"` silently truncated four readers.** Every one sat next to `timeout_overflow_mode: "throw"`. `break` returns the partial result as if complete, so `internal/downsampletier`'s `INSERT … SELECT` persisted a partial aggregate state per bucket — the "present but wrong" third state its own "absent, never wrong" safety argument says cannot exist — and `internal/routerrules` / `internal/optcorpus` fitted watermarks on truncated populations. `internal/deltaprefix`'s own doc claimed the caps make a runaway backfill "fail loudly", which is what `throw` actually does. All four flipped.

### Refuted

**"A cancelled query only gets a `KILL QUERY` on the experimental sharded path"** is factually true but is not a route-A/route-B divergence and is not a defect. The `KILL QUERY` lives in `acquireDataShardFanout`'s release closure, gated on `dataShardFanoutGate == nil` — i.e. the *data-shard* axis (`DataShardCount <= 1`), which is orthogonal to route B's time-slicing. It exists to prove the coordinating Distributed statement is dead before returning gate weight; with no gate there is no bound to protect, and clickhouse-go's `ClientCancel` already drives ClickHouse's own pipeline-cancel path. Adding the round-trip to every single-node cancellation would be pure cost. Not changed.

## #3197 — one reason for a cancellation on all three heads

A caller hanging up recorded `bad_request` on Tempo (499, a 4xx) and `backend_unavailable` on Prometheus and Loki (503). Same event, two verdicts, neither true.

The issue proposes fixing this through #3196's `telemetry.SetReason` / `httperr.TelemetryReason`. **#3196 is still open and conflicting, so neither exists on `main`** — I verified this directly rather than assuming the issue described current code. Rather than duplicate an unmerged branch's mechanism, each transport supplies the fact from a signal it already has: the request context on HTTP, `codes.Canceled` on the Tempo gRPC stream. That is also more robust than a sentinel classifier here, because prom's own cancellation constructor does `errors.New(...)` and destroys the sentinel.

Neither status moves: Tempo's 499 keeps a client hang-up out of the 5xx band, and prom/loki's 503 is upstream wire parity the compat harnesses assert. A per-request *deadline* is deliberately not folded in — `ApplyQueryTimeout` derives a child context and expires with `DeadlineExceeded`, so a timeout still reads as a timeout — and a recovered panic stays `internal`.

Cancellations stay `result="error"` (the query was not answered) but are now separable, which matters because Grafana cancels every in-flight request on a panel re-render, query edit or tab switch.

The enum's membership was re-typed by hand in four places, one of which silently dropped unknown values — so adding a member could compile and pass while going unrecorded. `telemetry.ErrorReasons()` is now the one authoritative list; every consumer derives from it, leaving a single deliberate literal pin in the public-contract test, which also asserts the two agree. `docs/observability.md` and the three dashboard copies gain the member, and the docs state plainly that `cerberus_status_class` is **not** unified — a cancellation stays 4xx on Tempo and 5xx elsewhere, so alerts keyed on the 5xx rate should exclude `cerberus_error_reason="canceled"`.

## Contradictions corrected

Five comments that disagreed with their own code, all found while verifying the claims above:

- `route_outcome.go` claimed detection is "never a string match" while its guard matching is by containment.
- The `chsql` RBGN bound doc still asserted the kind is unregistered and named a test since renamed to its opposite.
- A planner test row justified refusing `RangeBucketFanout` by calling it unroutable on the main spine, where `planner.go` says it is routable.
- `prom` and `loki` both called the cancellation reply "emphatically NOT a 5xx" while returning 503 — the contradiction that let the wrong telemetry reason go unnoticed. Tempo's equivalent comment was the honest one.

## Verification

Every fix was neutralized and watched go red before being restored. Notably, the neutralizations reproduce the reported symptoms exactly: a 5th memo-hit admitted with 4 cursors open and undrained; `bad_request` on 499 and `backend_unavailable` on 503 for the same cancellation; and 1 query → 2 observations, with route B's EMA dragged 1000 → 557.2 at K=4.

`go test -race ./...` clean, `just vet-tagged` clean, `forbid-sql-raw` and `forbid-skip` clean.

## Risk

Default-config behaviour is unchanged on the route-B work: the solver is off by default and every `SettingsRules` rule is individually flag-gated. `TestRouteBExecCtx_SettingsMatchRouteAAtK1` pins the two seams byte-equal at K=1 and fails for any future setting added to one and not the other. The `read_overflow_mode` and cancellation-reason changes are user-visible: a read that trips a cap now fails loudly instead of returning partial data, and a cancelled request moves out of `bad_request` / `backend_unavailable` into `canceled`.


tsouza:
Correction to the squashed commit message, for anyone reading `a8865236` in `git log`.

Auto-merge fired as soon as the required checks went green, so the squash body is
the default concatenation of the branch's commit messages rather than the summary
prepared for the merge. That concatenation includes a paragraph from
`557fcd63a` describing an **earlier, superseded** approach to #3197:

> each transport supplies the fact from its own signal instead: the request
> context on HTTP, `codes.Canceled` on the Tempo gRPC stream. No per-head error
> plumbing is involved…

That is not what merged. It described the design written before #3196 landed,
when `telemetry.SetReason` and `httperr.TelemetryReason` did not exist on `main`.
#3196 merged mid-review and the branch was resolved onto it. What is actually in
the tree:

- `httperr.TelemetryReason` gained a `context.Canceled` arm — the single shared
  classifier all three heads already call, not a per-transport signal.
- The Prometheus and Loki cancellation constructors carry
  `Reason: telemetry.ReasonCanceled` on the error, because each restates the
  failure in upstream's wording via `errors.New` and destroys the sentinel. That
  is precisely what `httperr.Error.Reason` was added for in #3196.
- Tempo needed no per-head change; its `errclass.go` already passes the raw error.
- The request-context check in `QueryMiddleware`, the `Outcome.AsCanceled` helper
  and the gRPC `codes.Canceled` special case were all **deleted** during the
  resolution rather than layered on top, since they were a second mechanism doing
  #3196's job.

One thing #3196 could not have caught, added here: its `knownReason` accept-list
is a fifth hand-written copy of the reason enum and it fails **closed** — an
unlisted value is silently dropped, not reported. `canceled` could have been
threaded through every head and still never reached the metric. It now derives
from `telemetry.ErrorReasons()`, which is the single authoritative list.

No code change is implied by this comment; the merged tree is correct. Only the
commit-message body is stale, and it cannot be rewritten now that it is on `main`.

tsouza:
One factual correction to the comment above, so the record is right about how
this landed: **auto-merge did not fire.** I merged #3201 by hand at 21:09 UTC
with `gh pr merge 3201 --squash --delete-branch`, after verifying all 17
required contexts green against the branch ruleset. `mergeStateStatus` read
`UNSTABLE` at the time, and I checked why before merging rather than through
it — a single non-required `gremlins phase2-other (./internal/chsql)` shard was
still in progress.

The consequence is the same, and the correction above stands: a manual
`--squash` also takes the default concatenation of the branch's commit
messages, so the superseded paragraph from `557fcd63a` is in the body either
way. That was my doing, not a race — the merge was mine to sequence, and a
prepared squash summary would have been used had I asked for one.

Everything else in the comment is unaffected: what merged is what it describes.

Issue #3215, opened 2026-09-08
test(mutation): the traceql-lower shard reports survivors its own tree kills


tsouza:
## What

The `phase4-traceql-lower` mutation shard reports a mutant as `LIVED` that the
tree under test provably kills. The efficacy it publishes therefore does not
describe the commit it ran on.

## Evidence

Two runs of the shard, on two different heads of `fix/unscoped-attr-numeric-coercion`:

| run | head | attempted | killed | lived | efficacy |
|---|---|---|---|---|---|
| 34282042253 | `9338eda1d` | 171 | 162 | 9 | 94.7368% |
| 34284947976 | `31bdbe20f` | 171 | 162 | 9 | 94.7368% |

The survivor sets are **byte-identical** — same nine `(file, line, column, type)`
tuples, diffed from both runs' `gremlins.json` artifacts:

```
lower.go 1176:38 INVERT_LOGICAL   lower.go 2178:26 INVERT_LOGICAL
lower.go 1938:34 INVERT_LOGICAL   lower.go 2203:30 INVERT_LOGICAL
lower.go 1939:14 INVERT_LOGICAL   lower.go 2347:44 INVERT_LOGICAL
lower.go 2061:27 INVERT_LOGICAL   lower.go 2446:9  INVERT_LOGICAL
lower.go 2827:94 CONDITIONALS_NEGATION
```

`31bdbe20f` is a commit whose ONLY content is a new test
(`internal/traceql/unscoped_attr_numeric_test.go`,
`TestIsAttributeReadRequiresBothCoalesceArms`) written specifically to kill
`lower.go:2203:30` — the `&&` between `v.Fn == chplan.FnIf` and
`len(v.Args) == 3` in `isAttributeRead`.

Applying that exact mutation locally, on that exact commit:

```
$ sed -i 's/return v.Fn == chplan.FnIf && len(v.Args) == 3 &&/return v.Fn == chplan.FnIf || len(v.Args) == 3 \&\&/' internal/traceql/lower.go
$ go test -count=1 ./internal/traceql/
--- FAIL: TestIsAttributeReadRequiresBothCoalesceArms
    --- FAIL: .../only_the_then-arm_is_an_attribute_read
    --- FAIL: .../neither_arm_is_an_attribute_read
    --- FAIL: .../only_the_else-arm_is_an_attribute_read
FAIL	github.com/tsouza/cerberus/internal/traceql	0.129s
```

[...]

Issue #2741, opened 2026-08-30
chsql: gremlins phase2-other efficacy below threshold — 13 surviving mutants, mixed equivalence status


tsouza:
## What

`gremlins phase2-other (./internal/chsql @ 95%)` failed on the v1.19.0 release-gate run (PR #2736) at 94.80% efficacy (237/250 mutants killed), 0.2 points under its 95% threshold. Found during the release ritual's release-gate run.

## The 13 surviving mutants

```
emit_size_bound.go:251:23              CONDITIONALS_BOUNDARY
prewhere.go:131:4                      INVERT_LOOPCTRL
prewhere.go:148:20                     CONDITIONALS_BOUNDARY
prewhere.go:283:15                     INVERT_LOGICAL
prewhere.go:287:18                     INVERT_LOGICAL
range_bucket_grid_native.go:197:12     CONDITIONALS_BOUNDARY
range_bucket_grid_native.go:200:13     CONDITIONALS_BOUNDARY
range_bucket_grid_native.go:203:22     INVERT_LOGICAL
range_bucket_grid_native.go:212:29     INVERT_LOGICAL
range_bucket_grid_native_bound.go:706:84  CONDITIONALS_BOUNDARY
set_op.go:327:35                       CONDITIONALS_BOUNDARY
set_op.go:490:5                        INVERT_LOOPCTRL
structural_join.go:525:53              CONDITIONALS_BOUNDARY
```

## What's already known (verified this session, before filing)

- **`structural_join.go:525:53`** — already documented as a genuine equivalent mutant in `structural_join_anchor_mutation_test.go`'s own "NOT KILLABLE" footer note (both the original and mutated code render byte-identical SQL: `QueryBuilder`'s renderer writes a bare `*` for an empty SELECT list, same as the explicit `verbatim("*")` fallback). Not a real gap.
- **`emit_size_bound.go:251:23`** — verified equivalent by hand this session: the mutation (`d > deepest` -> `d >= deepest`) only changes behavior when `d == deepest`, where the reassignment `deepest = d` is a no-op (same value, plain int, no other side effects). Also not a real gap. This file is new this release cycle (PR #2734) but the mutant is genuinely unkillable, not an undertested addition.
- **`set_op.go:327:35`** and **`set_op.go:490:5`** — already documented as "NOT KILLABLE" in `set_op_mutation_test.go`'s own footer notes.
[...]

Issue #2730, opened 2026-08-29
test(promql): phase4-promql-h mutants pass only via incidental, parallelism-sensitive coverage


tsouza:
`gremlins phase4-promql-h` (`./internal/promql @ 95%`) has repeatedly failed
CI on PR #2727 with an identical set of ~30 LIVED mutants concentrated in
`histogram_native_binop.go`, `histogram_native_mixed_or.go`, and the whole
`histogram_native_mixed_or_subquery_*` cluster (`aggregate_range_fn.go`,
`further_setop_range_fn.go`, `last_first.go`, `resets_changes.go`) — files
this PR never modifies.

## Root cause (confirmed locally)

These mutants are **not equivalent mutants and not random CI noise** — they
are genuine, pre-existing test-coverage gaps that happen to pass
*incidentally* under generous CPU parallelism and fail once parallelism
drops to match GitHub's actual `ubuntu-latest` free-tier runner spec
(2 vCPU).

Confirmed directly: `gremlins unleash` with the exact pinned CI fork/version,
exact flags (`--on-shutdown-status=not-run --timeout-max 15s
--timeout-coefficient 15`), and the leg's own exact file scope:

- **8 or 4 workers/cores (this machine's actual and an emulated mid-point):
  100% efficacy, 0 lived** — every time, across half a dozen runs.
- **2 workers with `GOMAXPROCS=2` and `taskset -c 0,1` (matching
  `ubuntu-latest`'s real 2-vCPU spec): reproduces `LIVED INVERT_LOGICAL at
  histogram_native_mixed_or.go:123:31`** — byte-identical to what CI itself
  reports.

Line 123 of `histogram_native_mixed_or.go` (`mixedExpHistogramSetOp`):

```go
if s.ExpHistogramTable == "" || ctx.metadataFullRange {
    return nil, false
}
```

This is a plain, timing-independent boolean guard — the `||` vs `&&`
mutant can only be told apart by a test where *exactly one* of
`s.ExpHistogramTable == ""` and `ctx.metadataFullRange` is true. Grepping
the package's own `gremlins_kill_*` test files (the established
convention for hand-written mutant-killing tests, see
`gremlins_kill_histogram_mixed_or_test.go`'s own header) turns up no
dedicated test for this exact line — the current 95%+ efficacy on `main`
must come from *incidental* coverage in some other test that happens to
exercise this exact branch combination as a side effect, and that
incidental coverage is evidently sensitive to test-scheduling / parallel
[...]

Issue #2338, opened 2026-08-17
gremlins phase2-other (chsql) mutation efficacy below 95% floor — 18 pre-existing LIVED mutants


tsouza:
## What happened

PR #2334's `gremlins phase2-other (./internal/chsql @ 95%)` CI job failed on a rerun triggered by commit `0c6b7fb34b601094d7bfcceb8be03d82fab48cd2`:

```
Mutation testing completed in 19 minutes 42 seconds
Killed: 335, Lived: 18, Not covered: 140
Timed out: 8, Not viable: 0, Skipped: 0
Test efficacy: 94.90%
Mutator coverage: 71.60%
gremlins efficacy 94.90084985835693% < threshold 95%
```

Run: https://github.com/tsouza/cerberus/actions/runs/32054218099/job/95460571171

## Why this isn't PR #2334's bug

PR #2334 only touches `internal/chsql/emit.go` (+2 lines, a dispatch case) and adds the new file
`internal/chsql/histogram_vector_join.go`. That new file is explicitly excluded from phase2-other's
mutation universe (it's claimed by `phase2-builder` per the mutation-phases partition fixed in
commit `e97e2326d`). None of the 18 LIVED mutants fall in either changed file — all 18 are in
pre-existing, untouched files:

- `histogram_quantile.go:217:16`, `:318:15` — CONDITIONALS_NEGATION
- `range_window.go:411:32`, `:1646:37` (×2), `:1812:43` (×2) — CONDITIONALS_NEGATION / INVERT_NEGATIVES / ARITHMETIC_BASE
- `range_window_fused.go:408:47`, `:408:60`, `:234:19` — ARITHMETIC_BASE / CONDITIONALS_NEGATION
- `range_window_variants.go:92:24`, `:114:18`, `:129:39`, `:151:26`, `:293:50`, `:293:42`, `:333:48`, `:351:23` — INVERT_LOGICAL / CONDITIONALS_BOUNDARY / ARITHMETIC_BASE / CONDITIONALS_NEGATION

`gremlins phase2-other` is not in the branch's required status checks
(`gh api repos/tsouza/cerberus/branches/main/protection --jq '.required_status_checks.contexts[]'`),
so it did not block PR #2334's merge, but the underlying mutation-coverage gap is real and pre-existing
in `internal/chsql`.

## Ask

Add or strengthen tests in `internal/chsql` so these 18 mutants are killed (pushing phase2-other's
test efficacy back over the 95% floor):

[...]

PR #3263, opened 2026-09-10
test(chsql): adjudicate every phase2-other survivor and settle the catch-all

### LINKED ISSUE #3221: epic(chsql): phase2-other has fallen below its efficacy floor three times; settle the catch-all's shape
## Not a blocker

This is a **pre-existing mutation-efficacy floor**, not a correctness defect. No
query answers wrongly because of it. The rolled-up `mutation` context is not one
of `main`'s 17 required status checks and is not in `release.yml`'s
`RELEASE_REQUIRED_CHECKS` either, so nothing here blocks a merge or a release.
It is filed as an epic because the same leg has now failed the same way three
times, and each previous fix closed the instance rather than the class.

## The class, not the number

| issue | closed | what it reported |
|---|---|---|
| #2338 | 2026-08-17 | `phase2-other` below the 95% floor — 18 pre-existing LIVED mutants |
| #2741 | 2026-08-30 | `phase2-other` below the threshold — 13 survivors, mixed equivalence status |
| this | — | `phase2-other` at 93.01% — 27 survivors |

Both prior issues were closed COMPLETED, and the leg is below the floor again
nine days later with **more** survivors than either of them reported. An epic
that says "93.01%, 27 survivors" will be closed a fourth time and reopened a
fifth. What has never been answered is why this one leg keeps coming back when
its three siblings over the same package do not.

## The structural question this epic exists to settle

`phase2-other` is `./internal/chsql`'s **catch-all** leg: `phase2-range`,
`phase2-builder` and `phase2-compare` each name the files they own, and
`phase2-other` takes everything else by denylisting those three sets
(`.github/scripts/mutation-phases.mjs`). Every new emitter therefore lands in it
by default, and it currently carries 386 attempted mutants — the largest leg
over the package.

All 27 survivors sit in **seven** of its files, and 21 of the 27 sit in just
three of them. So the question is not "is 95% the right number", it is:

**Do these seven files need tests of their own, or does the catch-all's
composition make a single package-wide floor the wrong shape for it?**

Concretely, the things worth deciding:

- Are the clustered files under-tested in a way the suite should fix directly
  (remedy 2 in `docs/test-strategy.md` § "Surviving-mutant policy"), or are some
  of these genuinely equivalent mutants that need documenting (remedy 1)? Nobody
  has adjudicated all 27 — #2741's title already conceded "mixed equivalence
  status" without resolving it.
- Should the heaviest clusters — `range_window_downsample_tier.go`,
  `prewhere.go`, `histogram_quantile_rankwalk_native.go` — become **curated legs
  of their own** with `include_files` allowlists, the way `phase4-traceql-lower`
  and `phase4-logql-lower` already are? A leg whose survivors are concentrated in
  three files is measuring three files and charging the whole package for it.
- Does the catch-all keep absorbing every new emitter indefinitely, or does the
  partition need a rule for when a file graduates out of it?

**Lowering the floor is not an available answer.** `MUTATION_MIN_EFFICACY` (95)
is a hard floor no phase entry may go below, checked by `tableViolations`, and
dropping a leg's `efficacy` to absorb survivors is the tolerance-file move this
repository does not make (hard invariant 7). Re-partitioning is a legitimate
answer; re-partitioning *to hide* a survivor is not — the mutants have to end up
somewhere and be adjudicated there.

## Evidence — the 27 survivors, grouped by file

From run
[34313501819](https://github.com/tsouza/cerberus/actions/runs/34313501819)'s
`gremlins-phase2-other` artifact, on `main` at `ab1d995f4`:

```text
test_efficacy   93.00518134715026
mutants_total   386   killed 359   lived 27   not viable 0   not covered 177
```

### `range_window_downsample_tier.go` — 9

```text
 94:8    CONDITIONALS_NEGATION
 95:32   INVERT_NEGATIVES
 95:32   ARITHMETIC_BASE
205:9    CONDITIONALS_NEGATION
209:9    CONDITIONALS_NEGATION
213:9    CONDITIONALS_NEGATION
226:48   ARITHMETIC_BASE
226:56   ARITHMETIC_BASE
274:23   CONDITIONALS_NEGATION
```

### `prewhere.go` — 7

```text
 66:28   CONDITIONALS_NEGATION
140:4    INVERT_LOOPCTRL
157:20   CONDITIONALS_BOUNDARY
196:5    INVERT_LOOPCTRL
216:4    INVERT_LOOPCTRL
292:15   INVERT_LOGICAL
296:18   INVERT_LOGICAL
```

### `histogram_quantile_rankwalk_native.go` — 5

```text
169:32   INVERT_LOGICAL
173:9    CONDITIONALS_NEGATION
304:8    CONDITIONALS_BOUNDARY
304:8    CONDITIONALS_NEGATION
381:15   CONDITIONALS_NEGATION
```

### `set_op.go` — 2

```text
327:35   CONDITIONALS_BOUNDARY
490:5    INVERT_LOOPCTRL
```

### `structural_join.go` — 2

```text
525:53   CONDITIONALS_BOUNDARY
738:69   CONDITIONALS_BOUNDARY
```

### `emit_size_bound.go` — 1

```text
251:23   CONDITIONALS_BOUNDARY
```

### `search_trace_limit.go` — 1

```text
 58:27   INVERT_LOGICAL
```

`prewhere.go` already carries a documented-equivalent note for a
`best < 0 || r < best` CONDITIONALS_BOUNDARY pair (see
`docs/test-strategy.md`), so some of the seven above may be that same construct
and some may not — which is exactly the adjudication nobody has completed.

## Read the mutant before adjudicating any of these

A survivor's `(file, line, column, mutator)` tuple does **not** identify the
program gremlins ran. The mutation is applied to the AST and the file is
re-printed with `go/printer`, so an operator swap that changes precedence comes
back re-parenthesised: `INVERT_LOGICAL` on the first `&&` of `a && b && c` is
`(a || b) && c`, never `a || b && c`. Retyping the operator at the cited
line and column produces a different program, and adjudicating that one proves
nothing about the survivor. #3215 is a full worked example of the mistake.

Since #3220, `.github/scripts/mutation-run.mjs` passes
`--output-diff-statuses l`, so each LIVED line in the leg's log is followed by a
unified diff of the source that actually ran. **Start from that diff**, not from
this issue's line numbers — the line numbers here are as of `ab1d995f4` and will
drift.

The `INVERT_LOOPCTRL` entries deserve their own warning: that mutator turns a
`continue` into a `break` (and back), which changes control flow rather than a
predicate, and the diff is the only readable statement of what it did.

## Done when

- Every one of the leg's survivors carries an adjudication — a distinguishing
  test, or a documented equivalence note that
  `.github/scripts/forbid-contradicted-mutants.mjs` accepts — rather than a
  count.
- The catch-all's composition has been decided rather than defaulted: either the
  clustered files have their own curated legs, or there is a written reason they
  should not.
- `phase2-other` clears its 95% floor on `main` without any leg's `efficacy`
  having been lowered.

tsouza:
## What this settles

`phase2-other` fell below its 95% efficacy floor three times — #2338 (18
survivors), #2741 (13), #3221 (27) — while its three siblings over the same
package never did. Each earlier issue closed the instance. This closes the
class.

**No leg's `efficacy` is lowered and no file changes legs**, so every mutant is
still executed by the same leg that executed it before. The partition change
here is documentation: a rejected option, a written reason, and a rule.

## All 27 survivors, adjudicated

Every verdict was established from the LIVED diff in the leg's own CI log (run
34416648191, job 102683004881), never from the `line:col` — a mutant is an AST
edit re-printed by `go/printer` (#3215). Each kill was then proven by hand:
apply the mutant, run the narrowed test, confirm it fails; revert, confirm it
passes.

Seventeen were real gaps and are now killed. Ten are equivalent, were already
documented in their files' `NOT KILLABLE` footers, and were re-derived
independently here rather than taken on trust.

| # | file | mutant | verdict |
|---|---|---|---|
| 1 | `emit_size_bound.go` | `d > deepest` → `>=` | equivalent — running max; at equality the mutant reassigns the same `int` |
| 2 | `histogram_quantile_rankwalk_native.go` | `\|\|` → `&&` on the two array-column guards | gap → killed |
| 3 | ″ | `err != nil` → `== nil` after `subqueryFrag` | gap → killed |
| 4 | ″ | `i < len(GroupByAliases)` → `>=` | gap → killed |
| 5 | ″ | `i < len(GroupByAliases)` → `<=` | gap → killed |
| 6 | ″ | `PhiExpr == nil` → `!=` | gap → killed |
| 7 | `prewhere.go` | `MaterializedColumn != ""` → `==` | gap → killed |
| 8 | ″ | `break` → `continue`, `touchesWide` latch | equivalent — set once, never reset |
| 9 | ″ | `r < best` → `<=` | equivalent — `r == best` unreachable across distinct columns |
| 10 | ″ | `break` → `continue`, `hitsSkip` latch | equivalent — same shape as 8 |
| 11 | ″ | `break` → `continue`, insertion-sort exit | equivalent — the prefix below the stop point is already ordered |
| 12 | ″ | `!columnOK \|\| !literalOK` → `&&` | equivalent — all four inputs reach the same final pair |
| 13 | ″ | `columnOK && literalOK` → `(columnOK \|\| literalOK)` | gap → killed (see the retraction below) |
| 14 | `range_window_downsample_tier.go` | `fn == "last_over_time"` → `!=` | gap → killed |
| 15-16 | ″ | `downsampleTierValFrag(-1)` → `(+1)`, two mutators | gap → killed |
| 17-19 | ″ | three `err != nil` → `== nil` guards | gap → killed |
| 20-21 | ″ | `numAnchors` `+ 1` → `- 1` and `/` → `*` | gap → killed |
| 22 | ″ | `TimestampColumn != RangeWindowAnchorAlias` → `==` | gap → killed |
| 23 | `search_trace_limit.go` | `\|\|` → `&&` on the two column-name guards | gap → killed |
| 24 | `set_op.go` | `i < j` → `i <= j`, in-place reversal | equivalent — adds only the `i == j` self-swap |
| 25 | ″ | `break` → `continue`, `shared[i]` latch | equivalent — only ever written false |
| 26 | `structural_join.go` | `len(cols) > 0` → `>= 0` | equivalent — at zero an empty `Select()` renders `SELECT *` |
| 27 | ″ | `len(seedWhere) > 0` → `>= 0` | equivalent — `Where()` with no args appends nothing |

## The gaps had one cause, and it is not the partition

An emitter lands in the catch-all by default, and an emitter's real tests are
routinely **build-tagged** (`//go:build chdb`, `//go:build integration`)
because emitted SQL is proven against an engine. The mutation lane compiles the
**untagged** build, so for it those tests do not exist. What was left untagged
for the clustered files was only `chplan_ir_discriminates_test.go`, which
asserts that two plans emit *different* SQL and never what either one emits —
executing the emitter's lines while pinning no token in them.

Covered-and-unpinned is exactly the state that reports LIVED, and no coverage
floor can see it. That is why 21 of the 27 sat in three files:
`range_window_downsample_tier.go` and `histogram_quantile_rankwalk_native.go`
(discrimination test only) and `search_trace_limit.go` (nothing untagged at
all).

Three new **untagged** `_mutation_test.go` files fix that. They also reach
mutants that were never *executed* before, not merely unkilled: nothing
untagged called `downsampleTierValueExprFrag`'s `idelta` / `irate` arms, so its
13 `NOT COVERED` mutants now run, and are killed by the byte-exact expected
strings.

## The partition decision

**Rejected:** carving `range_window_downsample_tier.go`, `prewhere.go` and
`histogram_quantile_rankwalk_native.go` into curated `include_files` legs the
way `phase4-traceql-lower` is. They execute 12, 69 and 8 mutants; two of the
three would be a leg measuring a single small file, and a one-file leg over an
undefended emitter reports its own catastrophic efficacy instead of dragging a
386-mutant leg to 93%. Splitting relocates the number and kills no mutant.

**Written down instead**, in the phase table where the next person will hit it:

- why the catch-all keeps absorbing new emitters (it must — see that file's own
  `include_files` vs `exclude_files` note) and why that is not the defect;
- a **graduation rule** with exactly one criterion: a file leaves this leg for
  a curated sibling on measured **wall clock**, when `gremlins unleash
  --dry-run` puts the leg materially past the ~315-executed-mutant band the
  four-way split targets. Never because it hosts survivors — the survivors move
  with it and the receiving leg inherits the failure, which is the same evasion
  as lowering `efficacy`;
- the corrected documented-equivalent tally: 10 of 386 executed (≈2.6%), and
  after this change those 10 are the leg's only survivors.

`docs/test-strategy.md` § "Surviving-mutant policy" gains the diagnostic that
comes before choosing a remedy: covered-and-unpinned is remedy 2 every time and
never remedy 1, and the tell is a source file with no untagged
`<file>_mutation_test.go` beside it.

## One retraction

#2741 recorded `prewhere.go`'s `columnOK && literalOK` INVERT_LOGICAL as a
"#2730-class CI-timing flake on an existing, correct test". It is not, and the
note claiming so is removed.

The real mutant is `(columnOK || literalOK) && …` — `go/printer`
re-parenthesises to preserve the tree — not the `columnOK || (literalOK && …)`
a reader gets by retyping the operator in the source. The test was written
against the retyped form. I checked out the pre-existing test, applied the real
mutant, and it **passed**: the kill claim was false and the LIVED report was
correct all along. The test now feeds `OtherCol = ServiceName`, where the
surviving `column` *is* the registered discriminator, and kills it.

## Verification

- Each of the 17 kills: mutant applied by hand, narrowed test **fails**;
  reverted, **passes**. Sources restored byte-identical (`git status` clean
  between runs).
- `go build ./...`, `go test ./internal/chsql/` — green.
- `verify-code-citations.mjs`, `forbid-contradicted-mutants.mjs`,
  `mutation-matrix.mjs` (table sound, 30 phases), `markdownlint-cli2` — green.

Closes #3221

https://claude.ai/code/session_0196feCkr8wd6meQ3jr9W8rF

### DIFF
--- a/.github/scripts/mutation-phases.mjs
+++ b/.github/scripts/mutation-phases.mjs
@@ -168,46 +168,93 @@ export const PHASES = [
     scope: './internal/chsql',
     efficacy: EFFICACY,
     workers: DEFAULT_WORKERS,
-    // catch-all leg: prewhere + ddl + structural_join + set_op +
-    // range_window_grid_native + scan_resource_bound + emit_size_bound +
-    // query_exemplars + histogram_quantile + tableshape +
-    // range_bucket_grid_native + range_bucket_grid_native_bound, plus every
-    // file no other leg claims (the true zero-mutant/uncovered files:
-    // absent_over_time, chaos_sleep, chaos_sleep_stub, doc,
-    // histogram_float_vector_join, histogram_vector_join, info_join,
-    // mixed_vector_join, search_trace_limit — range_bucket_grid_native and
-    // range_bucket_grid_native_bound were wrongly listed here before cerberus
-    // issue #2741: both carry real, covered mutants, they just had no
-    // dedicated `_mutation_test.go` file defending their input-validation
-    // guards until #2741 added one). A file newly added to internal/chsql
-    // needs no edit here — it is picked up automatically, which is why this
-    // leg keeps no positive file list to fall out of sync.
+    // The scope's CATCH-ALL leg: every internal/chsql file the three curated
+    // legs above do not name, including every file added since this table was
+    // last edited. On the CI run cerberus issue #3221 worked from, the files
+    // carrying executed mutants here were ddl, prewhere, structural_join,
+    // set_op, range_window_grid_native, scan_resource_bound, query_exemplars,
+    // histogram_quantile, range_bucket_grid_native,
+    // range_window_downsample_tier, range_window_grid_native_instant,
+    // range_bucket_grid_native_bound, emit_size_bound,
+    // histogram_quantile_rankwalk_native, tableshape, search_trace_limit,
+    // distinct_attr_keys and attr_strategy_fullmap; the rest of the leg's
+    // files carry only mutants the untagged build never reaches. A file newly
+    // added to internal/chsql needs no edit here — it is picked up
+    // automatically, which is why this leg keeps no positive file list to fall
+    // out of sync, and see the include_files-vs-exclude_files note at the top
+    // of this file for why exactly one leg per scope may work that way.
+    //
+    // WHY THIS LEG KEPT FAILING, AND WHY THE ANSWER IS NOT THE PARTITION
+    // (cerberus issue #3221). It fell below the 95% floor three times —
+    // #2338 (18 survivors), #2741 (13), #3221 (27) — while its three siblings
+    // over the same package never did. The first two closed the instance.
+    // #3221 adjudicated all 27 and found ONE cause behind every real gap, and
+    // it is neither this leg's size nor its composition:
+    //
+    //   An emitter lands here by default, and an emitter's real tests are
+    //   routinely BUILD-TAGGED (`//go:build chdb`, `//go:build integration`)
+    //   because emitted SQL is proven against an engine. The mutation lane
+    //   compiles the UNTAGGED build, so for it those tests do not exist. What
+    //   is left untagged is often only a DISCRIMINATION test —
+    //   chplan_ir_discriminates_test.go asserts that two plans emit DIFFERENT
+    //   SQL, never what either one emits — which makes the emitter's lines
+    //   COVERED while pinning no token in them. Covered-and-unpinned is
+    //   exactly the state that reports LIVED, and a coverage floor cannot see
+    //   it.
+    //
+    // That is why the survivors clustered: 21 of the 27 sat in the three files
+    // whose only untagged defence was that discrimination test
+    // (range_window_downsample_tier.go,
+    // histogram_quantile_rankwalk_native.go) or nothing at all
+    // (search_trace_limit.go). Carving those three into curated legs of their
+    // own — the phase4-traceql-lower shape #3221 asked about — was considered
+    // and REJECTED: they execute 12, 8 and 7 mutants, so each would be a leg
+    // measuring a single file, and a one-file leg over an undefended emitter
+    // reports its own catastrophic efficacy instead of dragging a 386-mutant
+    // leg to 93%. Splitting relocates the number and kills no mutant. #3221
+    // fixed the files instead — three new UNTAGGED `_mutation_test.go` files
+    // that pin the SQL those emitters render.
+    //
+    // GRADUATION RULE. A file leaves this leg for a curated sibling on ONE
+    // criterion, WALL CLOCK, measured: when `gremlins unleash --dry-run` puts
+    // this leg materially past the ~315-executed-mutant band the four-way
+    // split targets, re-measure every file and re-balance greedily, exactly
+    // as that split was derived. Never move a file because it hosts
+    // survivors. The survivors move with it and the receiving leg inherits
+    // the failure, so the only thing that changes is which number reports it
+    // — the same evasion as lowering `efficacy`, and the sibling of what
+    // "Rebalance by re-measuring" above already forbids.
     //
     // Documented-equivalent tally (docs/test-strategy.md's "Surviving-mutant
     // policy" #1 — proven, permanent, and NOT absorbed by lowering `efficacy`
-    // below `MUTATION_MIN_EFFICACY`; see that section for why). Re-measured
-    // via `gremlins unleash --dry-run` scoped to this leg's own
-    // `exclude_files`: 356 executed mutants today. 10 are proven equivalent
-    // (≈2.8%, comfortably under the ~5-point margin `efficacy` below leaves):
-    // prewhere.go:131,:148,:187,:207 (INVERT_LOOPCTRL — a "found it, stop"
-    // boolean latch or a sorted-subarray early exit; scanning further can
-    // never change the result) and :283 (INVERT_LOGICAL — a swap guard whose
-    // two orientations always resolve the same (columnOK, literalOK) pair —
-    // see prewhere_mutation_test.go's own footer), set_op.go:327,:490 and
-    // structural_join.go:525,:738 (set_op_mutation_test.go /
-    // structural_join_anchor_mutation_test.go's own footers), and
-    // emit_size_bound.go:251 (CONDITIONALS_BOUNDARY on a running-max update
-    // that reassigns the SAME value at the boundary — see
-    // emit_size_bound_mutation_test.go's own footer). cerberus issue #2741's
-    // own 13-survivor CI failure was NOT caused by this list growing past
-    // the margin — six of the thirteen were real, previously-undefended
-    // gaps in range_bucket_grid_native.go / range_bucket_grid_native_bound.go
-    // (now fixed) and one more was a #2730-class CI-timing flake on an
-    // existing, correct test (prewhere.go:287, reproduced and confirmed by
-    // manual mutation-and-revert — see prewhere_mutation_test.go's own
-    // TestIsNarrowIntegerDiscriminatorFinalReturnLogical). Re-count this
-    // tally the next time a mutant here gets a new "NOT KILLABLE" note, and
-    // re-partition the leg wider if it ever approaches the margin.
+    // below `MUTATION_MIN_EFFICACY`; see that section for why). Measured from
+    // the leg's own CI run on `main`: 386 executed mutants, of which exactly
+    // 10 are proven equivalent (≈2.6%, comfortably under the ~5-point margin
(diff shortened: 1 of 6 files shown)


tsouza:
## The leg's own verdict on this PR

Because this diff touches `.github/scripts/mutation-phases.mjs`, `mutation-select`
selected the FULL matrix rather than a diff-scoped subset — so `phase2-other`
genuinely executed here, and this is direct evidence rather than the
empty-matrix green the epic's own comment warned must not be read as one.

[Job 102710013816](https://github.com/tsouza/cerberus/actions/runs/34425357237/job/102710013816):

```text
Killed: 390, Lived: 10, Not covered: 164
Test efficacy: 97.50%
efficacy 97.51% >= threshold 95%
```

The ten survivors are exactly the ten documented equivalents, and nothing else:

```text
LIVED CONDITIONALS_BOUNDARY at emit_size_bound.go:251:23
LIVED INVERT_LOOPCTRL       at prewhere.go:140:4
LIVED CONDITIONALS_BOUNDARY at prewhere.go:157:20
LIVED INVERT_LOOPCTRL       at prewhere.go:196:5
LIVED INVERT_LOOPCTRL       at prewhere.go:216:4
LIVED INVERT_LOGICAL        at prewhere.go:292:15
LIVED CONDITIONALS_BOUNDARY at set_op.go:327:35
LIVED INVERT_LOOPCTRL       at set_op.go:490:5
LIVED CONDITIONALS_BOUNDARY at structural_join.go:527:53
LIVED CONDITIONALS_BOUNDARY at structural_join.go:740:69
```

Two numbers worth reading together. `Not covered` fell 177 → 164 while `Killed`
rose 358 → 390: the new untagged tests do not merely kill the seventeen
survivors, they also execute thirteen mutants in `downsampleTierValueExprFrag`'s
`idelta` / `irate` arms that no untagged test had ever reached, and kill those
too. (Attempted is 401 against the epic's 386 for that reason, plus two mutants
`main` gained since the run the epic quoted.)

Quoting the formula honestly: 97.51% credits the one `RUN TIMED OUT` as a
detection, which is correct per #2944 — the test binary's own `-timeout`
watchdog fired, which no compile can consume. Without that credit it is 97.26%
over the same 401 attempted. Both clear 95%, and no leg's `efficacy` was
touched.

The epic's three "Done when" conditions are therefore met on measured evidence
rather than on argument: every survivor carries an adjudication, the catch-all's
composition is decided in writing rather than defaulted, and the leg clears its
floor.

PR #3283, opened 2026-09-10
refactor(chplan): derive physical row schemas for every plan operator

tsouza:
Implements #3277 stage 3: compositional output schemas for all 38 plan operators. Nodes describe actual output columns, order, and declared roles; lowering stamps configured names and synthetic outputs; clones and optimizer rewrites preserve declarations. RowShapeOf remains unchanged, with explicit structural reconciliation of its existing approximations. No SQL expressions, settings, or wire output changes.

Driver checks bind wildcard scans on a clone to the independently parsed CREATE TABLE seed catalog. Focused tests cover sparse catalogs, immutability, explicit columns, database qualification, expression-subquery traversal, union catalog agreement, and wrong aliases. The verified leading-comment parser defect is corrected by prerequisite #3286, without weakening expected columns or altering generated fixtures.

## Physical acceptance

All three full TestLower corpora passed with package exit 0. Success markers are counted only for passing tests and report actual optimized root kinds, schema openness, and compared named columns.

| Head | Passing TestLower leaves | Driver-checked seeded leaves | Time |
| --- | ---: | ---: | ---: |
| PromQL | 777 | 767 | 1075.338s |
| LogQL | 191 | 191 | 92.767s |
| TraceQL | 234 | 226 | 14.171s |

The ten unmarked PromQL and eight unmarked TraceQL fixtures have neither seed nor expected_rows sections. There are no skipped seeded assertions. Every corpus driver schema was closed; compared named-column counts were 1, 2, 4, 5, 13, or 14 for PromQL, 1, 2, or 4 for LogQL, and 1–4 for TraceQL. Separate CrossJoin driver tests cover closed collisions and conservative open-schema cases, including cases with zero provable names; those are not overstated as exact-column proof.

Root counts:

- PromQL: Project 427, AbsentOverTime 13, RangeWindow 124, NaryVectorSetOp 20, VectorSetOp 24, VectorJoin 34, TopK 25, Filter 19, HistogramProjection 59, UnionAll 4, InfoJoin 8, Aggregate 1, RangeWindowStaleResample 3, RangeLWR 1, OrderBy 5.
- LogQL: AbsentOverTime 2, Project 34, TopK 5, VectorJoin 8, NaryVectorSetOp 6, VectorSetOp 3, RangeWindow 33, Filter 96, OrderBy 2, UnionAll 2.
- TraceQL: Project 194, MetricsAggregate 23, MetricsSecondStage 4, MetricsCompare 2, MetricsHistogramOverTime 3.

[...]

Issue #3277, opened 2026-09-10
epic(chplan): compositional row schemas and mixed-wrapper migration


tsouza:
## Epic tracker

Replace hand-maintained sample-shape inference with compositional physical row schemas and explicit live-sample-kind proof across `chplan`, PromQL lowering, SQL emission, optimization, and Prometheus API responses.

### Dependency chain

- #3345 — correct nested mixed `count_values` serialization; builds on completed #3326.
- #3347 — preserve mixed payload and live kinds through nested ranked/limit selectors.
- #3346 — consolidate selector policies after #3347.
- #3343 — separate physical mixed payload from live-float proof after #3346.
- #3342 — route subqueries by live sample kind after #3343.
- #3348 — reconcile generic mixed-wrapper consumers after #3342/#3343/#3346.
- #3349 — replace histogram shape guards after #3345/#3348.
- #3350 — route the API, ClickHouse emitter, optimizer, and PromQL set operations by validated schemas after #3349.
- #3351 — resolve temporal layouts by schema roles after #3350.
- #3352 — retire `Filter`/`TopK` legacy shape flags after #3348–#3351.
- #3353 — retire or demote the legacy `RowShapeOf` switch after #3352.
- #3344 — disposition residual bespoke families and run the final source-derived completion audit.

#3345 and #3347 can proceed independently; the displayed native order chooses #3345 first. Every delivery PR closes its child (`Closes #child`), never this parent. Keep #3277 open until #3344 verifies that every blocking child is closed.

### Completed delivery milestones

| Child | Closing PR | Delivered scope |
| --- | --- | --- |
| #3338 | #3280 | Source-derived `chplan` boundary ratchet |
| #3335 | #3282 | Optimized SQL, argument, and settings golden coverage |
| #3331 | #3283 | Compositional physical `RowType()` on every plan operator |
| #3354 | #3286 | Seed-catalog leading-comment prerequisite |
| #3328 | #3296 | Direct mixed-wrapper reference fixtures |
| #3329 | #3298 | Shared role-driven sample forwarders |
| #3332 | #3304 | Fail-closed mixed-operand policy table |
| #3334 | #3307 | Shared math, round, and clamp kernels |
| #3330 | #3310 | Shared label-rewrite kernels |
| #3341 | #3314 | Complete public-payload validation |
| #3336 | #3315 | Shared scalar-arithmetic projection |
| #3340 | #3316 | Shared ordinary sort policy ordering |
[...]

PR #3288, opened 2026-09-10
fix(promql): preserve IEEE-754 scalar division semantics

### LINKED ISSUE #3287: fix(promql): preserve IEEE-754 scalar division semantics
## Verified defect

`internal/promql/scalar.go` special-cases division by zero in `foldBinaryScalar`. Because `rhs == 0` also matches negative zero, the result loses the denominator's sign. Because the branch tests only the numerator's zero/sign, `NaN / 0` incorrectly becomes positive infinity.

Targeted `TryFoldScalar` regressions reproduced six failures: `1 / -0`, `-1 / -0`, `Inf / -0`, `-Inf / -0`, `NaN / 0`, and `NaN / -0`.

Prometheus v0.314.0's `promql/engine.go:scalarBinop` implements division directly as `lhs / rhs`, preserving IEEE-754 signs and NaN propagation.

Discovered while investigating a surviving mutation in #3283 for #3277. The `<` to `<=` mutation inside the existing zero branch is equivalent because an earlier branch already handles a zero numerator; source-spelling assertions would not repair the actual arithmetic defect.

## Acceptance

- Match Prometheus scalar division using native floating-point division.
- Exercise the public constant-folding entry point with positive/negative zero, finite values, both infinities, and NaN; assert zero signs as well as numeric classification.
- Preserve existing positive-zero division tests and correct comments that incorrectly claimed to distinguish the equivalent boundary mutation.

tsouza:
## Summary

- Fix constant-folded division by negative zero and NaN propagation using native IEEE-754 division, matching Prometheus `scalarBinop`.
- Add 16 public `TryFoldScalar` regression cases covering finite values, both zero signs, infinities, and NaN.
- Correct test comments that claimed to kill a provably equivalent conditional-boundary mutation in the removed special case.

Closes #3287. Found during #3277 / #3283 mutation verification; this PR is limited to the independent arithmetic defect.

## Test plan

- Reproduced six failures before the fix: `1 / -0`, `-1 / -0`, `Inf / -0`, `-Inf / -0`, `NaN / 0`, `NaN / -0`.
- Passed `go test -timeout 15m -race -count=1 -run '^(TestTryFoldScalarDivisionIEEE754|TestFoldBinaryScalar_DivByZeroNegativeBranches)$' ./internal/promql` after the fix.
- The direct invocation narrows to these regressions; `just test-unit` has no package/test arguments and runs the whole tree. Race and timeout match that recipe; no whole-tree preflight was run.
- CI supplies whole-tree checks. No generated artifacts were edited.

## Notes for reviewers

Prometheus v0.314.0 implements scalar division as `lhs / rhs` in `promql/engine.go`. The old `rhs == 0` special case discarded the denominator's sign and converted a NaN numerator to positive infinity. Tests assert explicit expected values, NaN classification, and signed zero through the public parser/folder, rather than duplicating the implementation as their oracle.

### DIFF
--- a/internal/promql/gremlins_kill_test.go
+++ b/internal/promql/gremlins_kill_test.go
@@ -315,18 +315,13 @@ func TestLowerClamp_MixedBoundsTakeComputedPath(t *testing.T) {
 	}
 }
 
-// TestFoldBinaryScalar_DivByZeroNegativeBranches pins the `<` boundary
-// at scalar.go:foldBinaryScalar:`if lhs < 0` in its DIV case. That
-// branch returns -Inf for any strictly-negative LHS divided by zero. The
-// sibling `if lhs == 0` branch just above it already handles the 0/0 case
-// (NaN), and the fall-through returns +Inf. A boundary mutant would
-// either misclassify the lhs=0 case (already caught earlier in the
-// switch) or shift the negative/positive split — pinning two opposite
-// signs catches both.
+// TestFoldBinaryScalar_DivByZeroNegativeBranches checks signed infinities
+// and NaN for division by positive zero through the public folding surface.
+// TestTryFoldScalarDivisionIEEE754 also covers negative zero and NaN inputs.
 //
 // Driven via TryFoldScalar on `(-1) / 0` (parses as
 // BinaryExpr{UnaryExpr{NumberLiteral{1}}, DIV, NumberLiteral{0}}) so
-// the kill ties to the public surface.
+// the assertions exercise the public surface.
 func TestFoldBinaryScalar_DivByZeroNegativeBranches(t *testing.T) {
 	t.Parallel()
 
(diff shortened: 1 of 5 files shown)


tsouza:
Independent review of `8bdee1549c9c6bac7366da43e24ea1071bf81def` completed with no findings.

The reviewer compared the change against the installed pinned Prometheus `scalarBinop`, whose DIV branch is `lhs / rhs`. All 16 explicit expected results are consistent with IEEE-754 division. The test requires successful folding, checks NaN separately, and checks `math.Signbit` alongside equality so positive and negative zero cannot pass as equivalent. The six pre-fix failures correspond exactly to the previously incorrect signed-zero and NaN branches.

The comment changes remove obsolete claims about the deleted conditional boundary; they do not weaken tests or suppress mutants. Current-head CI must still pass before merge.

tsouza:
The failing `phase4-promql-other` job on `8bdee1549` reported 145/155 detected mutants (93.55%), with the same ten survivors diagnosed on #3282/#3283. The shared test-only repair is applied here as `7ffbf7180` (original `776f87deffb5068524de980983a0d239854508f4`).

The owning agent's canonical local shard run confirmed these five exact previous survivors KILLED: `exp_histogram_merge_summap.go:277` arithmetic; `histogram_native_mixed_or_subquery_outer_aggregate_fold.go:101` boundary, `:119` negation, and both boundary/negation mutations at `:157`. The tests inspect actual lowered plan alignment/default inner grids and the returned partition slice capacity. No thresholds, suppression lists, or production behavior changed in this test repair.

On this scalar branch, the focused race-detected tests passed in 1.259s:

```sh
go test -timeout 15m -race -count=1 -run '^(TestMixedOuterAggregateFoldRejectsZeroRange|TestMixedOuterAggregateFoldDefaultStepAndAlignment|TestHistogramMergeScalePartitionCapacity)$' ./internal/promql
```

The repair was pushed immediately after this confirmation. Current-head CI remains the merge gate.

Issue #3292, opened 2026-09-10
fix(test): decode arbitrary map aliases in the chDB adapter


tsouza:
## Verified failure

The chDB test adapter cannot decode an otherwise identical metric query when its Map column uses an arbitrary output alias. This prevents full-handler execution coverage of arbitrary custom `schema.Metrics.AttributesColumn` names.

Verified on main `dee6253d72c489be3ee8a3401ac2efe1ff41297d`, while preparing #3289 acceptance coverage. A temporary test in package `internal/chclienttest` seeds the existing `sessionRecoverySeedDDL` and executes these three queries through the same `Client.Query`:

```sql
SELECT MetricName, Attributes AS Attributes, TimeUnix, Value FROM session_recovery_probe;
SELECT MetricName, Attributes AS labels, TimeUnix, Value FROM session_recovery_probe;
SELECT MetricName, Attributes AS labels_map, TimeUnix, Value FROM session_recovery_probe;
```

The first two pass and return one sample with label `k=v` and value `1`. The third fails with:

```text
chclienttest: rows.Err: could not cast to type: MAP
```

The isolated test ran with the canonical `test-chdb` recipe flags, narrowed to `./internal/chclienttest/`, plus `-race` and a temporary overlay. The failure takes less than one second and does not involve query lowering, a missing table, or missing optional seed columns. A full-handler variant with custom `AttributesColumn=labels_map` also fails during row decoding with `converting NULL to string is unsupported`.

## Cause and boundary

`internal/testsql/query_rewrite.go` determines Map serialization using `IsMapColumn`, a fixed list of known names. It contains `Attributes` and `labels`, but not arbitrary configured aliases. chDB's Parquet driver cannot directly decode the resulting raw Map cell. `internal/chclienttest/client.go` expects the serialized attribute value for its positional sample decoder.

This is a test-adapter limitation, not evidence that production custom-schema query execution is broken. Adding one more alias to the list would only hide this particular reproduction.

## Acceptance

- Decode both canonical and arbitrary custom Map output aliases without extending a name-specific allow-list for each new alias.
- Assert the actual returned sample labels and value, not just the absence of an error.
- Preserve existing metric, log, histogram, mixed-row, and metadata serialization behavior.
- Add a full-handler custom-attributes regression after the adapter supports arbitrary names.

[...]

PR #3291, opened 2026-09-10
fix(promql): classify guard failures from exception messages only

tsouza:
## Summary

Addresses the error-classification portion of #3289; the missing-name forwarding repair remains tracked there and this PR does not close the issue.

Classify guard failures from verified exception messages rather than guard literals embedded in failing SQL. Typed non-guard ClickHouse exceptions are authoritative. Support native and columnar typed guard messages, including retained engine envelopes; preserve chDB's anchored code-395/code-768 textual envelopes and its verified streaming `error in chunk: ` prefix. Static and dynamic duplicate-series messages must start the decoded exception message, not occur in its SQL trailer.

## Verification

- Focused pre-fix regression: eight negative cases returned incorrect HTTP 422 in both instant/open and cursor paths.
- Post-fix: 19 cases across both paths pass, pinning HTTP 502 for unrelated/unknown-identifier failures and HTTP 422 plus clean messages for genuine guards. Includes standalone and wrapped typed guards, typed non-guard precedence, columnar envelopes, textual/chunk errors, and dynamic duplicate-series messages.
- Existing typed grid-native budget regressions pass.
- Executed chDB positive regressions pass: `TestQueryRange_InstantNameDrop_DuplicateLabelset_ChDB` (395), `TestQuery_InstantNameDrop_DuplicateSeriesTags_ChDB` (768), and `TestQueryRange_Subquery_RateInner_DuplicateSeriesTags_ChDB` (768).

Targeted commands preserve race detection and the recipe's 15-minute timeout; no dedicated focused recipe exists:

```sh
go test -race -timeout 15m -count=1 -run '^TestGuardClassification_ExceptionMessageBoundary$' ./internal/api/prom/
GODEBUG=asyncpreemptoff=1 go test -race -tags chdb -timeout 15m -count=1 -run '^(TestQueryRange_InstantNameDrop_DuplicateLabelset_ChDB|TestQuery_InstantNameDrop_DuplicateSeriesTags_ChDB|TestQueryRange_Subquery_RateInner_DuplicateSeriesTags_ChDB)$' ./internal/api/prom/
```

The separately selected `TestInfo_ConflictingLabel_ChDB` fails before its guard on a missing histogram table. Reproduced unchanged on untouched main and tracked in #3290; no fixture or lowering changes are included here.

PR #3294, opened 2026-09-10
test(promql): pin aggregate-storage scan role contracts

tsouza:
## Summary

Closes #3293.

Pin the exact role declarations and actual `scanFromTables(...).RowType()` output for nine storage/configuration combinations: raw tables with delta-prefix support disabled/enabled, the configured prefix table, downsample tables independently of prefix configuration, and classic/native histogram controls. All sample aliases and the optional zero-threshold column are explicitly custom configured.

The sole production-source edit names the existing two-column identity-prefix length `aggregateStorageIdentityColumns`; classification and emitted SQL are unchanged.

## Verification

The post-merge #3283 phase4-promql-other job failed at 158/167 killed (94.61%). Four survivors changed the aggregate-storage predicate in `metricScanRoles`; the earlier phase-other proof did not include the newly added `schema.go`. Exact failure: https://github.com/tsouza/cerberus/actions/runs/34503420667/job/102961824208.

The focused unchanged-production test passed with race detection:

```sh
go test -timeout 5m -race -count=1 -run '^TestMetricScanRolesStorageContracts$' ./internal/promql
```

There is no parameterized recipe for one test; this narrows the race-detected unit recipe to the exact regression.

Four exact source-edit Go overlays each failed the new test with assertion failures, not compile errors:

- Outer OR to AND: configured-prefix and both downsample cases fail.
- Inner AND to OR: enabled-prefix raw/classic/native cases fail.
- Prefix nonempty test inverted: configured-prefix case fails.
- Prefix table equality inverted: configured-prefix and enabled-prefix raw/classic/native cases fail.

These are the exact AST changes printed by CI, with parentheses preserved. The unchanged 95% threshold and full phase are verified by this PR's CI; no full-phase local result is claimed. The focused proof adds four kills to the failing run's observed set, projecting 162/167 (97.01%) if other outcomes stay unchanged.

PR #3295, opened 2026-09-10
fix(promql): preserve label rewrites over name-dropped windows

### LINKED ISSUE #3289: bug(promql): label transforms over projected range grids reference missing MetricName
## Verified failure

While auditing the zero-behavior-change boundary of #3277 stage 4, two existing query-range compositions failed through the full HTTP handler and chDB execution path:

```promql
label_replace(rate(http_requests_total[5m] @ 1767229200), "dst", "$1", "job", "(.*)")
label_replace(abs(rate(http_requests_total[5m])), "dst", "$1", "job", "(.*)")
```

Both underlying ClickHouse errors are `Code: 47. Unknown expression identifier MetricName`. The pinned query returns HTTP 502. The nested query currently returns HTTP 422 with the unrelated `range window sample fanout exceeds the series-times-anchors resource bound` text because that string is also present in its failing SQL; the logged original error is still the missing column, not an actual resource-limit breach.

## Reproduction

Verified at commit `a20772c55` using a temporary Go test overlay in package `internal/api/prom` with the existing `newChDBServer` and `runRangeModeQueryRange` helpers; no repository or generated-fixture edits were needed.

- Schema: existing `rangeOffsetSumDDL` from `handler_chdb_range_offset_test.go`.
- Seed: 31 minute-spaced monotonically increasing samples, values 0 through 30, from `2026-01-01 00:40:00` through `01:10:00` UTC; metric `http_requests_total`, `job=api`, service `svc`.
- Request: start `2026-01-01 01:00:00` UTC, end `01:05:00`, step `1m`.
- Control `label_replace(rate(http_requests_total[5m]), "dst", "$1", "job", "(.*)")` succeeds with one series and six expected timestamps.
- Both failing queries above reproduce consistently with `go test -race -tags chdb -timeout 10m -count=1` restricted to the temporary test.

## Cause

`wrapRangeWindowAtBroadcast` and a grid `projectValueOverInner` publish `(Attributes, anchor_ts, TimeUnix, Value)` in a `Project`. `RowShapeOf(Project)` recognizes the mixed discriminator but otherwise reports `SampleRowShape`, even though these projects do not publish `MetricName`.

`projectAttributesOverInner` consequently selects the canonical quartet and references a nonexistent input name. Actual emitted shape:

```sql
SELECT MetricName, mapFilter(...) AS Attributes, TimeUnix, Value
FROM (
  SELECT Attributes, anchor_ts, anchor_ts AS TimeUnix, Value
  FROM (...)
)
```

The nested math case has the same missing-name seam above the first value-rewrite project. The subsequent label collision guard cannot repair that invalid reference.

## Repair and acceptance boundary

Pin both compositions with executed handler regressions and repair the forwarding contract. Check sibling label_join and nested label-transform paths when selecting the repair scope. Preserve the collision guard, explicit column order, custom-schema names, and histogram/mixed payload behavior. This is a real behavioral repair and must not be hidden inside a refactor described as SQL-identical. The observed misleading nested-query error classification also needs explicit regression coverage and correction.

This issue does **not** claim an offset timestamp bug. The same executed reproduction asserts all six exact unshifted timestamps for both `abs(rate(http_requests_total[5m] offset 1m))` and `abs(abs(rate(http_requests_total[5m] offset 1m)))`; both pass. The canonical TimeUnix emitted by `gridAnchorFrag` is already unshifted, so the nested canonical path and the single-wrapper HTTP anchor-relabel path are equivalent for those cases.

tsouza:
## Summary

Fix label_replace/label_join over name-dropping value wrappers and pinned range broadcasts. A legacy Sample-classified input can publish attributes, a real timestamp, and value without publishing its metric-name column. When its closed physical schema proves that float-only contract, materialize the existing empty-name convention in the canonical quartet instead of referencing a missing column.

Keep the established canonical temporal layout: simply dropping the name and forwarding anchor/time changes the collision aggregate to a derived output, which synthesizes the wrong timestamp for pinned broadcasts. The canonical layout preserves the actual request grid and the duplicate-labelset guard.

The repair does not change open/unknown schemas, name-preserving inputs, timestampless outputs, histogram helpers/payloads, or discriminator-bearing outputs. Excluding any histogram helper here is a conservative boundary for this specific float repair, not a definition of a complete histogram payload. General role-driven forwarding remains part of #3277.

Closes #3289. Its independent error-classification repair already merged in #3291.

## Verification

- New full-handler chDB regressions: all 24 cases pass with race detection. Two schemas cover label_replace/label_join, pinned rate, single/nested abs, nested label chains, offset rate, and name-preserving last_over_time controls. Assertions compare complete label maps (including absence of __name__), all six exact timestamps, and numeric sample values.
- Before the repair, those cases produced 14 missing-name failures and 10 passing controls. The final implementation preserves the controls and repairs every failure.
- New structural regressions: all 34 cases pass with race detection, covering arbitrary attribute aliases and closed/open, named/unnamed, missing or opaque required roles, timestamp, histogram-payload/helper, and discriminator boundaries.
- Existing label-related TXTAR comparisons pass unchanged, including optimized SQL/args/settings and IR. No generated artifact changed.
- Existing native-histogram/mixed label and collision-guard unit regressions are included in the focused final check.

The custom-schema HTTP lane uses the supported non-default Map alias `labels`; arbitrary `labels_map` execution is not claimed. The independently reproduced chDB adapter limitation is tracked in #3292. Arbitrary aliases are covered structurally by the new unit tests.

### DIFF
--- a/internal/api/prom/handler_chdb_label_forwarding_test.go
+++ b/internal/api/prom/handler_chdb_label_forwarding_test.go
@@ -1,0 +1,128 @@
+//go:build chdb
+
+package prom_test
+
+import (
+	"fmt"
+	"maps"
+	"math"
+	"net/http"
+	"net/http/httptest"
+	"strconv"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/tsouza/cerberus/internal/api/prom"
+	"github.com/tsouza/cerberus/internal/chclienttest"
+	"github.com/tsouza/cerberus/internal/schema"
+)
+
+// TestQueryRange_LabelRewritePhysicalColumns_ChDB exercises name-dropping
+// windows behind broadcasts and value rewrites through the complete HTTP path.
+// The label rewrite must forward the physical timestamps without referring to
+// a metric-name column that its input no longer publishes.
+func TestQueryRange_LabelRewritePhysicalColumns_ChDB(t *testing.T) {
+	const (
+		seedLeadMinutes = 20
+		seedSampleCount = 31
+		querySteps      = 5
+		valueTolerance  = 1e-12
+	)
+	start := time.Date(2026, 1, 1, 1, 0, 0, 0, time.UTC)
+	step := time.Minute
+	end := start.Add(querySteps * step)
+	rows := make([]string, 0, seedSampleCount)
+	for i := range seedSampleCount {
+		ts := start.Add(time.Duration(i-seedLeadMinutes) * step).Format("2006-01-02 15:04:05.000000000")
+		rows = append(rows, fmt.Sprintf("('http_requests_total', map('job','api'), 'svc', toDateTime64('%s',9), %d.0)", ts, i))
+	}
+	seedDDL := strings.Replace(rangeOffsetSumDDL, "Value Float64", fmt.Sprintf("Value Float64, AggregationTemporality Int32 DEFAULT %d", schema.AggregationTemporalityCumulative), 1)
+	seed := seedDDL + "\nINSERT INTO otel_metrics_sum (MetricName,Attributes,ServiceName,TimeUnix,Value) VALUES " + strings.Join(rows, ",") + ";"
+	for _, schemaName := range []string{"default", "custom"} {
+		t.Run(schemaName, func(t *testing.T) {
+			metrics := schema.DefaultOTelMetrics()
+			physicalSeed := seed
+			if schemaName == "custom" {
+				metrics.MetricNameColumn = "metric_id"
+				metrics.AttributesColumn = "labels"
+				metrics.TimestampColumn = "sample_time"
+				metrics.ValueColumn = "sample_value"
+				// Preserve ResourceAttributes as a separate physical field while
+				// renaming the canonical sample attributes column.
+				physicalSeed = strings.NewReplacer("ResourceAttributes", "ResourceAttributes", "MetricName", metrics.MetricNameColumn, "Attributes", metrics.AttributesColumn, "TimeUnix", metrics.TimestampColumn, "Value", metrics.ValueColumn).Replace(seed)
+			}
+			client := chclienttest.NewChDB(t)
+			client.Seed(t, physicalSeed)
+			handler := prom.New(client, metrics, nil)
+			mux := http.NewServeMux()
+			handler.Mount(mux)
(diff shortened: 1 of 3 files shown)


tsouza:
Reviewed the narrow forwarding repair at `add833a77555d8b16dcf913b90ce0ff396cc3a00`.

The new path only synthesizes a dropped metric name for a closed physical float sample with the required named attributes, timestamp, and value roles. Existing names, open or opaque outputs, histogram helpers, and discriminators retain the prior path. The canonical layout preserves the timestamp behavior required by pinned broadcasts and label-collision handling.

The structural tests pin the exact projection order, input identity, rewritten attributes expression, and name policy across 34 default/custom-schema cases. The 24 executed HTTP cases assert complete labels, all six timestamps, and numeric values for direct, pinned, nested, offset, and name-preserving controls. Existing label snapshots remain unchanged; no generated artifacts are edited.

This is the explicit behavior repair for #3289, not the general role-driven forwarder refactor in #3277. That refactor must use this corrected behavior as its baseline. Final-head CI remains required before merge.

tsouza:
The optional `phase4-promql-other` failure on this PR's original head is resolved on main by merged #3294 (`f7cedd839e937857daf01582da2b3f641d0b4e46`).

The failed run reported 158/167 kills. Its four observable `schema.go` predicate survivors are the exact four mutations killed by #3294's new storage-contract tests; the remaining five match the previously inspected equivalent survivors. There are no new `label_fns.go` survivors. The repaired full phase has now completed successfully: https://github.com/tsouza/cerberus/actions/runs/34508340093/job/102976246525.

All 17 ruleset-required contexts pass on the reviewed head `add833a77555d8b16dcf913b90ce0ff396cc3a00`. A conflict-free prospective merge into current main `286866279ad63bda5b8aaf69099cf6530619ba37` produces tree `2123d1a8ffbdcda77694779d9ba3c5f015ddbac8`: it changes only this PR's three reviewed files, preserves the repaired schema and new storage-contract test byte-for-byte, and preserves all three forwarding files byte-for-byte from the reviewed head.

The old-head optional failure is therefore accounted for by an already-merged, verified repair, not ignored or bypassed. Normal squash merge retains both fixes; no admin override or required-check exemption is used.

Issue #3284, opened 2026-09-10
refactor(chplan): retire redundant input-column fields using child row schemas


tsouza:
## Purpose

Track the separate input-column field migration explicitly required by the corrective scope comment on #3277. The compositional schema implementation is being introduced in #3283; this follow-up depends on that schema landing. Field retirement is not part of #3277's implementation.

An output schema alone does not replace input-column resolution. Classify each field by its actual emitter use before removing it, and resolve against the correct child or lookup table.

## Verified distinctions

- `RangeLWR`'s column quartet and `HistogramProjection`'s payload column fields identify inputs.
- `MetricsSecondStage.ValueAlias` identifies an input despite its alias terminology.
- `MetricsAggregate.ValueAlias` and `MetricsHistogramOverTime.BucketAlias` / `ValueAlias` are output naming contracts, not redundant input fields.
- `RangeWindowGridNativeVectorAgg.AnchorAlias` / `PartialCountAlias` name outputs.
- `RangeBucketFanout.TimestampCol` identifies an input; `AnchorAlias` names an output.
- `RangeBucketGridNative.BucketCountsCol` / `ExplicitBoundsCol` serve both input and output naming responsibilities; separate those responsibilities before migration.
- `AbsentOverTime.TimestampColumn` identifies an input; its metric-name, attributes, and value fields name synthesized outputs.
- `MetricsCompare` has child-input names, separate-table lookup names, and output aliases. A child schema cannot resolve a separate table's columns.

## Acceptance

1. Inventory and classify the named column fields, including `*Column`, `*Col`, `*Alias`, and partition-key spellings, by actual use rather than spelling.
2. Migrate redundant input names incrementally through the appropriate child schema, preserving configured/custom names and multi-input ownership.
3. Preserve explicit output aliases and separate-table lookup names unless a separately specified replacement preserves their contracts.
4. Update constructors, clone/equality behavior, and durable documentation together.
5. Prove unchanged emitted SQL and query results with focused tests, schema-versus-driver checks, and appropriate reference-parity lanes. Include intermediate-node tests where root-only fixtures cannot exercise a kind.
6. Regenerate affected artifacts with canonical recipes, never manual golden edits.

Keep migrations independently reviewable. This work does not replace PREWHERE's predicate-reference traversal or ProjectionPushdown's required-input analysis.

tsouza-squid:
All tracked children are closed or merged:

| Child | Delivery | |
|---|---|---|
[...]

Issue #3302, opened 2026-09-10
bug(promql): timestamp over nested mixed vectors drops histogram series


tsouza:
## Problem

On merged-main commit `b4ea3f1dc64139b5384487c2877d8df800c49177`, `timestamp` over a mixed vector wrapped in `sort_by_label` incorrectly discards histogram samples:

```promql
timestamp(sort_by_label(fo_wrapped_hist_side_exp_hist or fo_wrapped_float_side_gauge, "series"))
```

The same failure occurs with the union operands reversed and with `sort_by_label_desc`.

## Verified evidence

A six-case diagnostic lowers each query, runs the optimizer, executes the resulting SQL against chDB, and compares the decoded result with the live Prometheus oracle. The seed contains two histogram series (`h1`, `h2`) and three distinct float series (`f1`, `f2`, `f3`).

- Both direct `timestamp(histogram or float)` orientation controls return all five series and pass reference parity.
- All four nested cases return only the three float series. Prometheus returns all five, including `h1` and `h2`.
- The nested plans contain one strict float-discriminator filter; the direct controls contain none. `timestamp` reads sample time, not the float payload, so discarding histogram samples is incorrect.
- Surviving sample values correctly equal the evaluation time, `1767225601`. The diagnostic also observes different physical SQL timestamp columns between direct and nested paths; this is not a claim of an additional HTTP timestamp bug without handler-level verification.

## Repair boundary

Preserve both sample kinds when `timestamp` consumes an already-admitted mixed plan, while producing the normal float-valued timestamp result. Preserve union shadowing, metric-name dropping, sample-time versus evaluation-time rules, range anchors, and existing query-admission boundaries. Pin both union orientations and sort directions, direct mixed controls, ordinary float and histogram-only inputs, and instant/range behavior before changing the timestamp-family lowering.

This behavior defect was found while auditing #3277's wrapper migrations. Fix it explicitly in a separate change before that family's behavior-preserving refactor; do not regenerate expected results to accept lost series.

tsouza:
Reproduced on `fb0d84143f72d5e86291bf0833ed91ce5e27d7a9` while preparing the timestamp family for #3277.

Both direct mixed-union controls return five series. All four nested `sort_by_label` / `sort_by_label_desc` cases (both union orientations) return only three floats instead of five samples. The nested route reaches `guardedValueProjection`, which filters out histogram rows before timestamp conversion. Optimized chDB execution and live Prometheus comparison confirm the mismatch.

[...]

PR #3304, opened 2026-09-10
refactor(promql): gate mixed operand admission by family

### LINKED ISSUE #3332: refactor(promql): gate mixed operand admission by family
Parent epic: #3277

## Outcome

Introduce the fail-closed per-family, per-site mixed-operand policy table without widening any admitted query family.

## Completion evidence

Delivered by #3304 with policy inventory and mutation coverage.

tsouza:
## Summary

Part of #3277: make the existing mixed-sample admission rules explicit before migrating wrapper families to shared kernels.

- Add a fail-closed table of 50 family/site registrations. Every registration initially uses the existing bespoke behavior; this change does not enable another query shape or alter payload handling.
- Check the actual wrapper family at the existing root, recursive-operand, and already-lowered-plan seams. Missing registrations and unsupported modes return an error before invoking the guarded builder.
- Preserve current argument evaluation order, including literal-inverted clamp and computed-bound clamp. The original mixed operand is authorized before a subsequent filter can hide its legacy shape; this table-only change does not add payload narrowing.
- Regenerate rejection-catalogue evidence and reconcile its documented curated internal-invariant rationales with the new dispatch paths.

## Verification

- The 50-row inventory and focused authority tests cover missing rows, unsupported modes, callback non-execution, actual family propagation, and existing recursive admission boundaries.
- Focused race-detected policy and wrapper controls pass.
- All 799 existing PromQL lowering fixtures pass without changing their expected output.
- All 1,322 tracked `test/spec/` TXTAR files are unchanged from the role-forwarder prerequisite.
- Canonical parity-ledger regeneration and non-update catalogue/citation checks pass.

This is a behavior-preserving policy extraction. Wrapper-kernel migrations and legacy shape retirement remain part of #3277; this PR does not close that issue. Input-field retirement is separately tracked in #3284.

Closes #3332


tsouza:
CI follow-up on the policy extraction:

- Extracted `timestampFunctionName` for the PromQL function spelling, separately from the `timestamp-eval` policy-family identifier. The production change is constant substitution only; parser dispatch and payload behavior are unchanged.
- Added independent generic date/timestamp family checks: the opposite family can be denied without affecting an authorized call, while denying the actual family rejects before projection. Ordinary float controls remain admitted with both mixed-family registrations removed. A forged unknown function must return the explicit adapter error instead of constructing an invalid projection.
- Reproduced both surviving date-adapter mutants locally: each passed the previous 79 focused race subtests. The new tests kill the exact unknown-function guard mutant by detecting its invalid projection and kill the exact family-dispatch mutant in four positive/negative authorization cases. Both temporary mutations were restored before verification.
- The final focused policy/date/projection/comparison suite passes 14 top-level tests and 137 race-detected subtests. Canonical parity-ledger regeneration passes; its two source-evidence updates retain the same internal classifications with re-reviewed rationales. CI remains the lint authority; no local lint result is claimed.

This does not change admission boundaries or implement the separately tracked nested-timestamp and name-matching defects (#3302, #3305).

Issue #3469, opened 2026-09-15
bug(e2e): bwc-minio hot-cold TestReSeedRowCountStability fails with unbounded row growth


tsouza-squid:
## What

The nightly `e2e` workflow's `bwc-minio (hot-cold)` job has failed on every
run from 2026-09-12 through 2026-09-15 (it was green on 2026-09-11), with
`TestReSeedRowCountStability` failing across every data-shape subtest:

```
--- FAIL: TestReSeedRowCountStability (9.67s)
    --- FAIL: TestReSeedRowCountStability/metrics-gauge-shaped (0.02s)
        reseed_stability_test.go:268: metrics-gauge-shaped: row count 2096
        exceeds the bounded-duplication ceiling 1920 (320 rows/tick x 6
        ticks of slack) — growth looks unbounded
    --- FAIL: TestReSeedRowCountStability/metrics-sum (0.03s)
    --- FAIL: TestReSeedRowCountStability/metrics-histogram (0.02s)
    --- FAIL: TestReSeedRowCountStability/metrics-exponential-histogram (0.01s)
    --- FAIL: TestReSeedRowCountStability/logs (0.02s)
```

Run 104312549049 (part of nightly run 34948197074, 2026-09-15) also shows the
`cerberus` pod failing its liveness/readiness probes during the run
(`connect: connection refused`, then HTTP 503), which may be a symptom of the
same underlying issue or a separate one — not disentangled in this pass.

Every data shape fails uniformly, which points at the shared mechanism
(`test/e2e/seed/cmd/seed/stale.go`'s `ALTER TABLE ... DELETE`-based stale-row
cleanup, driven by both this test and the concurrently-running rolling
re-seeder) rather than any one metric family.

## What was ruled out

`git log` between the last-green nightly (2026-09-11T08:02Z, run 34577191134)
and the first-red one (2026-09-12T07:54Z, run 34681913822) shows 8 merged
PRs (#3355, #3356, #3358, #3359, #3363, #3370, #3372, #3373) — none touch
`test/e2e/seed/`, ClickHouse storage-policy/TTL config, or mutation
behaviour. `test/e2e/seed/cmd/seed/reseed_stability_test.go` and `stale.go`
themselves have no commits since #3145, well before this window.

## What wasn't determined

This job is one of the three e2e lanes that need a live k3d + MinIO cluster
to reproduce and profile (`CLAUDE.md` invariant 5) — this diagnosis pass
[...]

Issue #3468, opened 2026-09-15
perf(promql): exp-histogram query-duration panel still trips ClickHouse's memory cap at modest ranges


tsouza-squid:
## What

The self-observability dashboard's P95-latency-by-language panel, when it
queries the native/exponential histogram `cerberus_queries_duration_exp_hist`
via `histogram_quantile(0.95, sum by (cerberus_ql)
(rate(cerberus_queries_duration_exp_hist[5m])))`, can still trip either
ClickHouse's own per-query memory cap or cerberus's own
`HistogramMergeBudgetMessage` resource-bound guard — at modest, not just wide,
`(range, step)` combinations. This was found while diagnosing the nightly
`e2e` failures tracked by #3402 (2026-09-11 through 2026-09-15).

## Evidence

All from the nightly `dashboard` lane, k3d stack, after the panel migrated to
the native histogram in #3171/#3174 (merged 2026-09-08):

- `iterate-filter-drill.spec.ts`, a **5-minute** baseline window, step 15s:
  ```
  [Cerberus :: P95 latency by language :: A] baseline query → 422
  body: {"status":"error","errorType":"execution","error":"query processing
  would use too much memory in query execution (ClickHouse memory limit
  exceeded; per-query cap 1073741824 bytes)"}
  ```
  This is the raw ClickHouse OOM message, not cerberus's own pre-rejection
  guard — meaning nothing caught this before ClickHouse itself hit the wall,
  on a window an order of magnitude smaller than anything the existing
  resource-bound calibration work (#3178/#3234/#3239/#3247/#3252/#3266/#3268)
  measured against.

- `iterate-time-ranges.spec.ts`, `range=now-1h step=5m`:
  ```
  body: {"status":"error","errorType":"execution","error":"native histogram
  merge exceeds the series-per-group or merged-bucket-width resource bound"}
  ```
  This is `chplan.HistogramMergeBudgetMessage` firing instead — a different
  guard than the pinned `MEMORY_LIMIT_MESSAGE` dual contract that spec file's
  own header doc documents (that contract predates this panel's native-
  histogram migration and only names the raw ClickHouse message).

- `compose_grafana_smoke.spec.ts`, visiting the dashboard at its own default
  `now-1h` time range (no deliberately wide probe):
  ```
  [dash:cerberus-self:Cerberus] http: POST /api/ds/query → 400
  body: {"results":{"A":{"error":"execution: native histogram merge exceeds
[...]

PR #3470, opened 2026-09-15
fix(e2e,telemetry): point the self-observability P95 panel back at the classic histogram

tsouza:
## What

Points the self-observability dashboards' P95-latency-by-language panel back
at the classic-bucket histogram `cerberus_queries_duration_seconds_bucket`,
reverting the query shape (not the metric collection) #3171/#3174 shipped on
2026-09-08.

## Why

Diagnosing the nightly `e2e` failures tracked by #3402 (five consecutive red
nights, 2026-09-11 through 2026-09-15, identical signature every night):

- `dashboard-shard-info` (crawl shards): `crawl/lints.spec.ts`'s
  classic-histogram-degeneracy lint requires at least one provisioned
  `histogram_quantile` panel consuming a classic `_bucket` family
  (`stacks.ts`'s `minQuantileConsumedFamilies: 1` for the k3d stack,
  documented as "carries one histogram_quantile panel (p95 over
  cerberus_queries_duration_seconds)"). Since #3171/#3174 the panel queries
  the native histogram `cerberus_queries_duration_exp_hist` instead — no
  `_bucket` suffix — so the lint finds zero, unconditionally, every night.
- `dashboard-shard` (smoke-a/b): `iterate-time-ranges.spec.ts` misclassifies
  the native-histogram query as the N6 fabricated-value bug (its
  `extractHistogramName` heuristic treats any non-`_bucket` histogram_quantile
  target as guaranteed-empty — true for the real N6 case, wrong for a
  genuine native histogram), and separately its pinned 422 dual-contract
  only recognizes the raw ClickHouse memory-limit message, not cerberus's
  own `HistogramMergeBudgetMessage` guard, which now also fires for this
  panel at `range=now-1h step=5m`.
- `iterate-filter-drill.spec.ts` and `compose_grafana_smoke.spec.ts` have no
  422 carve-out at all, and both hit real rejections: a plain 5-minute
  probe returned ClickHouse's own `MEMORY_LIMIT_EXCEEDED` (per-query cap
  1073741824 bytes), and the dashboard's own default `now-1h` view returned
  cerberus's merge-budget rejection — an operator opening this dashboard
  normally would see a broken panel.

All four came from the same panel change. Rather than teach five e2e spec
files to recognize a native-histogram-consuming panel as legitimate (a real
gap, but one that requires the query path itself to be proven safe first —
see below), this PR reverts the one panel to the classic-bucket metric,
which `internal/telemetry/metrics.go`'s `QueryDurationLegacy` still emits
correctly, deliberately kept alive for one release for exactly this kind of
rollback. Nothing about the native histogram's *collection* changes.

## Scope

- `test/e2e/grafana/dashboards/cerberus.json` (k3d source)
- `test/e2e/grafana/compose/dashboards/cerberus.json` (compose source)
- `test/e2e/k3s/grafana-dashboards.yaml` (k3d ConfigMap copy, kept
  byte-identical to the source per `TestK3dDashboardConfigMapMatchesItsSourceFile`)
- `.github/scripts/quickstart-canary.mjs` (pinned home-dashboard expression)

No engine, telemetry-collection, or test-helper code changed.

## Filed alongside

- #3468 — the deeper problem: even after #3268's samples-per-series-per-window
  guard, `histogram_quantile(0.95, sum by (cerberus_ql)
  (rate(cerberus_queries_duration_exp_hist[5m])))` can still trip ClickHouse's
  own memory cap at ranges far smaller than prior calibration work measured.
  Re-migrate the panel once that's resolved.
- #3469 — `bwc-minio (hot-cold)`'s `TestReSeedRowCountStability` has failed
  every night since 2026-09-12, unrelated to the above (different code area,
  different start date, no correlating commit found). Not fixed here; needs
  live-cluster reproduction.

## Verification

- `go test ./test/regression/... -run 'Dashboard|QuickstartCanary'` — all
  pass, including the byte-for-byte k3d/ConfigMap parity check.
- `node --test .github/scripts/quickstart-canary.test.mjs` — 26/26 pass.
- Both dashboard JSON files parse.
- The `dashboard` and `bwc-minio` e2e lanes themselves need a live k3d
  cluster (CLAUDE.md invariant 5) — not reproduced locally; the next nightly
  run is the confirmation for the dashboard lane. #3402 self-closes on the
  next clean nightly, so it isn't referenced with `Closes` here — this PR
  addresses the dashboard-side failures only, not `bwc-minio`.


tsouza:
Closing this stopgap — the user wants the actual root cause (tracked in #3468) fixed instead of reverting the panel. The native-histogram P95 query needs to be made safe against ClickHouse's memory cap (or pre-rejected cleanly by a tightened guard), not swapped back to the classic-bucket metric.

Issue #3607, opened 2026-09-19
test: close vacuous release-audit assertions and discovery gaps


tsouza:
## Release audit findings

The 1.21.0 audit found tests that can pass after losing the behavior they claim to validate. Resolve these before release qualification.

- `test/regression/e2e_hand_copied_literals_test.go:74`: the memory-cap comparison iterates over `yamlEnvValues`, which returns an empty slice when the setting is missing or unrecognized. Require the setting in each stack and reject malformed values.
- `test/regression/docs_only_action_test.go:119`: removing the paths-filter consumer leaves the renderer check satisfied. Require the consumer and validate its connection to the renderer.
- `test/regression/chart_config_parity_test.go:47`: an empty discovered key set passes. Require meaningful discovery and test its failure behavior.
- `test/regression/go_test_timeout_budget_test.go:111`: zero discovered test commands pass. Validate discovery and account for recipes delegating test execution to scripts.
- Initializer-only assertions in `migration-artifact.test.mjs`, `release-source-pr-dashboard-gate.test.mjs`, `notify-perf-nightly-failure.test.mjs`, `notify-perf-nightly-selfcheck-failure.test.mjs`, `e2e-bwc-replicated-verify.test.mjs`, and `run-prometheus-compatibility.test.mjs` do not validate their consumers. Replace with behavior or cross-source checks where the stated contract warrants a test.

Verification: demonstrate that representative removal or bypass mutations fail the revised focused checks, while the unmodified implementation passes. These findings were identified by source inspection; mutation runs remain necessary.

PR #3608, opened 2026-09-19
test: close vacuous release audit guards

tsouza:
## Problem and change

Release-audit regression checks could pass after losing their inputs or consumers. Require discovered memory-cap values, chart environment keys, workflow test commands, and a paths-filter consumer before reporting success.

Replace initializer-only checks with executable notifier-option checks, replica-topology verdict cases, tester default/override argument cases, and comparisons against the Dockerfile copy/entrypoint paths and workflow crawl-prefix selection.

Addresses #3607. Further validation of delegated timeout discovery and malformed memory-cap values remains tracked there; this PR does not close it.

## Validation

The four focused Go regression tests passed with the race detector. The six affected Node test files passed. Empty chart-key and workflow-job discovery mutations failed at their new guards. The combined missing-filter/missing-memory-setting run also failed. Temporary mutations were restored, and commit/push hooks passed.


tsouza:
Focused race-enabled checks pass on commit 9d7253cd7e14550c5cd5685d7a4ac71b50c8642f.

Additional local mutation validation:

- Feeding the chart environment-key scanner an empty string makes `TestChartEnvKeysAreReachableFromAConfigFile` fail with `contains no concrete CERBERUS_* environment keys`.
- Replacing the discovered workflow jobs with an empty slice makes `TestGoTestTimeoutFitsItsJobBudget` fail with `no workflow job invokes a Justfile recipe`.
- A prior focused run with the paths-filter consumer renamed and the k3s memory-cap setting renamed also exited nonzero.

All temporary mutations were restored; the worktree is clean. Consumer-behavior coverage remains tracked in #3607 and is still required for the release audit.

PR #3610, opened 2026-09-19
fix(e2e): avoid query log collector distributed join

### LINKED ISSUE #3609: fix: bound datashard query-log collector memory during release qualification
Main E2E run https://github.com/tsouza/cerberus/actions/runs/35454966834 failed in datashard N=2 after 133 burst requests with zero 5xx and zero transport errors.

The verifier's subsequent distributed query-log collection failed with ClickHouse code 241: total memory limit exceeded, current RSS 734.57 MiB, maximum 734.78 MiB, while executing MergeTreeSelect. The failing query uses GLOBAL LEFT JOIN between child rows and terminal initiator rows from clusterAllReplicas(system.query_log).

Reduce the verifier's query-log collection memory demand while retaining every admission-control and exception-attribution assertion. The log establishes where memory failed, but does not isolate join allocation from scan-block memory or concurrent server activity. Diagnose that distinction before choosing the fix. Validate with focused local collector tests and the real E2E lane.

This blocks release 1.21.0 qualification. Do not suppress code 241 or weaken the tested workload to obtain a pass.

tsouza:
## Problem and change

Main E2E run https://github.com/tsouza/cerberus/actions/runs/35454966834 failed with ClickHouse code 241 while the verifier collected query-log evidence after a burst with zero 5xx responses. The failing collection joined two distributed query-log scans inside ClickHouse.

Read terminal initiator metadata once, then attribute child statements and memory exceptions in Node. This removes the distributed join while retaining parent-host fallback, exceptional terminal attribution, and the original query snippet used to exclude direct writers and non-SELECT operations.

Fixes #3609, subject to the real E2E validation below confirming the observed memory failure is resolved.

## Validation

The focused Node verifier suite passes, including separate-parent attribution for a SELECT with an exceptional initiator and rejection of an INSERT using the same parent. `git diff --check` passes.

The real datashard E2E lane must pass before this fix is considered verified. The separate release-performance timeout remains unresolved and still blocks release 1.21.0.


tsouza:
Real datashard qualification passed on implementation commit `089a52c50edd3bb8103d8f338ec0cced83b3e3c1`:

- [Datashard-only E2E run 35457033107](https://github.com/tsouza/cerberus/actions/runs/35457033107): N=2, N=4, and replica-affinity jobs all succeeded.
- The N=4 verifier explicitly reported all assertions passed, with peak concurrent SELECT-only statements 14, maximum concurrent dispatches 4, maximum scans observed 5, and maximum effective shard count 2.
- Observed peak query memory at effective shard counts 1 and 2 was 717780 and 507174 bytes, against configured ceilings of 268435456 and 134217728 bytes.

Follow-up commit `adbfc0e72ff918f1a5200def84ebed249f435528` changes comments and indentation only: stale GLOBAL JOIN documentation now describes the separate metadata lookup, and the incident comment does not claim that the failure identified the particular allocation exhausting server headroom.

Focused `node --test .github/scripts/e2e-datashard-verify.test.mjs` and `git diff --check` pass. Required checks must also pass on the current PR head before merge. The issue remains open until the change is merged.

